PR #3, opened 2025-10-13
Upgrade to pyo3 0.26, fix warnings, remove unsafe usage

alex:


alex:
This removes the usage of `unsafe` in favor of using pyo3's safe APIs. The motivation is that there's no need for `unsafe` here and removing it improves the ease of validating correctness/safety.

svlandeg:
Hi! Sorry for the late follow-up. Anyway, with `rustbpe` now being a separate repo/package, I'll go ahead and close this one.

PR #5, opened 2025-10-13
replace eval() with AST-based safe math parser

jaberjaber23:


lukestanley:
This would add ~77 additional code lines and 200 test lines.
Perhaps a more targetted approach could keep the codebase lean?

karpathy:
perhaps just a hint of danger....

:D jk jk. We do have the whole machinery in `execution.py`, which I'd possibly reach for before adding stuff like this.

jaberjaber23:
haha, I just added the test file so you can test it before merging the commit:D

PR #6, opened 2025-10-13
Add fused-QKV attention path, gated MLP options, and LR warmup toggles

vimeto:
### Description
Added configurability by:
- config-controlled fused-QKV
- extended MLP blocks to support SwiGLU/GEGLU
- introduce warmup toggles for AdamW/Muon
- cover the new behavior with unit tests

### Testing

- `python -m pytest tests/test_attention_fusion.py tests/test_mlp_variants.py tests/test_schedules.py`
- Ran CPU and GPU pre-training loops (CPU)

### DIFF
--- a/nanochat/schedules.py
+++ b/nanochat/schedules.py
@@ -1,0 +1,50 @@
+"""Learning rate schedule utilities."""
+
+
+def compute_lr_multiplier(
+    step: int,
+    total_steps: int,
+    *,
+    warmup_ratio: float = 0.0,
+    warmdown_ratio: float = 0.0,
+    final_lr_frac: float = 0.0,
+) -> float:
+    """Compute LR multiplier with linear warmup and warmdown phases.
+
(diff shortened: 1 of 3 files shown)


svlandeg (commented):
Hi @vimeto, thanks for the PR! Did you use LLMs to write any of this? (we currently have an "LLM disclosure policy" just to help us understand what is manually written/verified and what isn't)

I haven't reviewed this in detail yet, nor checked with Andrej, but my gut feeling is that there are several potentially interesting features here and it might make more sense to split them up in separate PRs, for easier reviewing / decision making for each separate feature. I realise that's some work, but it might help speed up the reviewing process.

svlandeg:
Closing this after a quick discussion with Andrej - the goal is to keep `nanochat` a minimal implementation and not a configurable framework. The idea is that you can fork it easily and adjust anything you like etc.

PR #13, opened 2025-10-13
Prevents OOM crashes in production web server from unbounded cache growth:D

jaberjaber23:
the resize_() call was growing unbounded. replaced it with explicit allocation + bounds checking. now memory stays capped at 32k tokens instead of growing forever until crash:p

### DIFF
--- a/nanochat/engine.py
+++ b/nanochat/engine.py
@@ -53,20 +53,27 @@ def use_calculator(expr):
     return eval_with_timeout(expr)
 
 # -----------------------------------------------------------------------------
+# KV cache with bounded memory growth:D
+# >> why: Original resize_() grows unbounded causing OOM in web server with long conversations
+# >> what: Add MAX_CACHE_SIZE limit + proper memory management with explicit deallocation
+# >> how: we can check the bounds before resize, use new tensor allocation instead of resize_ for proper GC
 class KVCache:
-    """
-    Works hand-in-hand with the GPT model to maintain the KV cache.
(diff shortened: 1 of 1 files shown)


nanochat/engine.py
```diff
@@ -53,20 +53,27 @@ def use_calculator(expr):
...
-    Works hand-in-hand with the GPT model to maintain the KV cache.
-    Note that the .pos advances automatically after the last layer of the Transformer inserts.
-    """
+    MAX_CACHE_SIZE = 32768
 
     def __init__(self, batch_size, num_heads, seq_len, head_dim, num_layers):
-        # Each of K/V is of shape (B, H, T, D) and we have one per layer of the Transformer.
         self.kv_shape = (num_layers, 2, batch_size, num_heads, seq_len, head_dim)
         self.kv_cache = None
-        self.pos = 0 # current position in time in the cache
```
PhantomInTheWire, line 66: why are you removing comments and docstrings?

nanochat/engine.py
```diff
@@ -53,20 +53,27 @@ def use_calculator(expr):
     return eval_with_timeout(expr)
 
 # -----------------------------------------------------------------------------
+# KV cache with bounded memory growth:D
+# >> why: Original resize_() grows unbounded causing OOM in web server with long conversations
+# >> what: Add MAX_CACHE_SIZE limit + proper memory management with explicit deallocation
+# >> how: we can check the bounds before resize, use new tensor allocation instead of resize_ for proper GC
```
PhantomInTheWire, line 59: this seems a bit excessive

PR #15, opened 2025-10-14
tests: use urllib and skip on network failure

guangyusong:


guangyusong:
Thanks for reviewing. Yes, this happens with self‑hosted runners.

svlandeg:
`rustbpe` has been moved to a separate package, so I'll go ahead and close this.

PR #23, opened 2025-10-14
feat: Add ROCm and device-agnostic support

LokiMetaSmith:
This change adds support for ROCm and makes the codebase device-agnostic, allowing it to run on different hardware backends including ROCm, CUDA, and CPU.

The key changes are:
- Modified `pyproject.toml` to use ROCm-compatible PyTorch wheels and added the `pytorch-triton-rocm` dependency.
- Refactored `nanochat/common.py` to dynamically detect the available hardware and set the device and distributed backend accordingly.
- Updated all training, evaluation, and inference scripts to be device-agnostic, removing hardcoded CUDA references.
- Adapted `speedrun.sh` for single-device execution by replacing `torchrun` with `python`.
- Updated `nanochat/report.py` to provide more generic GPU information.

### DIFF
--- a/speedrun.sh
+++ b/speedrun.sh
@@ -17,6 +17,7 @@ export OMP_NUM_THREADS=1
 # For example, for a gfx1151 GPU, we can use gfx1100 (11.0.0).
 export HSA_OVERRIDE_GFX_VERSION=11.0.0
 NANOCHAT_BASE_DIR="$HOME/.cache/nanochat"
+export NANOCHAT_BASE_DIR="$HOME/.cache/nanochat"
 mkdir -p $NANOCHAT_BASE_DIR
 
 # -----------------------------------------------------------------------------
--- a/README.md
+++ b/README.md
@@ -6,6 +6,10 @@
 
 This repo is a full-stack implementation of an LLM like ChatGPT in a single, clean, minimal, hackable, dependency-lite codebase. nanochat is designed to run on a single 8XH100 node via scripts like [speedrun.sh](speedrun.sh), that run the entire pipeline start to end. This includes tokenization, pretraining, finetuning, evaluation, inference, and web serving over a simple UI so that you can talk to your own LLM just like ChatGPT. nanochat will become the capstone project of the course LLM101n being developed by Eureka Labs.
 
+## Talk to it
+
+To get a sense of the endpoint of this repo, you can currently find [nanochat d32](https://github.com/karpathy/nanochat/discussions/8) hosted on [nanochat.karpathy.ai](https://nanochat.karpathy.ai/). "d32" means that this model has 32 layers in the Transformer neural network. This model has 1.9 billion parameters, it was trained on 38 billion tokens by simply running the single script [run1000.sh](run1000.sh), and the total cost of training was ~$800 (about 33 hours training time on 8XH100 GPU node). While today this is enough to outperform GPT-2 of 2019, it falls dramatically short of moden Large Language Models like GPT-5. When talking to these micro models, you'll see that they make a lot of mistakes, they are a little bit naive and silly and they hallucinate a ton, a bit like children. It's kind of amusing. But what makes nanochat unique is that it is fully yours - fully configurable, tweakable, hackable, and trained by you from start to end. To train and talk to your own, we turn to...
+
 ## Quick start
 
 The fastest way to feel the magic is to run the speedrun script [speedrun.sh](speedrun.sh), which trains and inferences the $100 tier of nanochat. On an 8XH100 node at $24/hr, this gives a total run time of about 4 hours. Boot up a new 8XH100 GPU box from your favorite provider (e.g. I use and like [Lambda](https://lambda.ai/service/gpu-cloud)), and kick off the training script:
@@ -128,6 +132,10 @@ The `speedrun.sh` script has been configured to run on a single GPU by default, 
 ### Memory Considerations
 
 If your GPU has less than 80GB of VRAM, you may need to reduce the `device_batch_size` in the training scripts to avoid running out of memory. This will increase training time but will allow the model to train successfully on lower-VRAM cards.
+
+## Running on CPU / MPS
+
+If you'd like to tinker with nanochat on your Macbook or a CPU machine, there is a work in progress [CPU|MPS PR](https://github.com/karpathy/nanochat/pull/88) up here. If you're on Macbook, use `--device_type=mps` when running `base_train.py`. See the PR and its diff for more. You're not going to get too far without GPU nodes, but at least you'll be able to run the code and maybe train a very tiny LLM with some patience.
 
 ## Questions
 
--- a/nanochat/ui.html
+++ b/nanochat/ui.html
@@ -108,6 +108,15 @@
             background: transparent;
             border: none;
             padding: 0.25rem 0;
+            cursor: pointer;
+            border-radius: 0.5rem;
+            padding: 0.5rem;
+            margin-left: -0.5rem;
+            transition: background-color 0.2s ease;
+        }
+
+        .message.assistant .message-content:hover {
+            background-color: #f9fafb;
         }
 
         .message.user .message-content {
@@ -115,6 +124,21 @@
             border-radius: 1.25rem;
             padding: 0.8rem 1rem;
             max-width: 65%;
+            cursor: pointer;
+            transition: background-color 0.2s ease;
+        }
+
+        .message.user .message-content:hover {
+            background-color: #e5e7eb;
+        }
+
+        .message.console .message-content {
+            font-family: 'Monaco', 'Menlo', 'Ubuntu Mono', 'Consolas', 'Courier New', monospace;
+            font-size: 0.875rem;
+            background-color: #fafafa;
+            padding: 0.75rem 1rem;
+            color: #374151;
+            max-width: 80%;
         }
 
         .input-container {
@@ -255,6 +279,8 @@
 
         let messages = [];
         let isGenerating = false;
+        let currentTemperature = 0.8;
+        let currentTopK = 50;
 
         chatInput.addEventListener('input', function() {
             this.style.height = 'auto';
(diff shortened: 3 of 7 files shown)


LokiMetaSmith:
So far seeeems to be working. 

running it on a strix halo 128gb version.

LokiMetaSmith:
nvm, more work needs to be done.
current issue: nanochat/.venv/lib/python3.10/site-packages/torch/_utils.py:101: UserWarning: expandable_segments not supported on this platform (Triggered internally at /pytorch/c10/hip/HIPAllocatorConfig.h:36.)

jon-hotaisle:
Confirmed. Seeing the same issue on a box of MI355x in a recent container: docker.io/rocm/pytorch:rocm7.0.2_ubuntu24.04_py3.12_pytorch_release_2.8.0

2025-10-15 at 01 58 40 png

speedrun.sh
```diff
@@ -12,6 +12,10 @@
 
 # Default intermediate artifacts directory is in ~/.cache/nanochat
 export OMP_NUM_THREADS=1
+# For newer AMD GPUs that are not yet officially supported by PyTorch ROCm builds,
+# we can override the detected GPU architecture to a compatible one.
+# For example, for a gfx1151 GPU, we can use gfx1100 (11.0.0).
+export HSA_OVERRIDE_GFX_VERSION=11.0.0
```
jammm, line 18: PyTorch should be supporting gfx1151 already via. TheRock. This shouldn't be enabled by default IMO.
This could also cause issues when running on gfx9 which @jon-hotaisle was running on.

pyproject.toml
```diff
@@ -22,15 +23,18 @@ dependencies = [
 requires = ["maturin>=1.7,<2.0"]
 build-backend = "maturin"
 
-# target torch to cuda 12.8
+# target torch to rocm 6.3
 [tool.uv.sources]
 torch = [
-    { index = "pytorch-cu128" },
+    { index = "pytorch-rocm63" },
```
jammm, line 29: ROCm 6.4 is the latest supported one, and there's also ROCm 7. I suggest switching to TheRock wheels for best compatibility https://github.com/ROCm/TheRock/blob/main/RELEASES.md#installing-pytorch-python-packages - these also install ROCm itself as pip wheels, so you don't need to worry about whether the system contains a specific version of ROCm. 

EDIT: Currently TheRock wheels are device-specific, so maybe it's best to stick with 6.4/nightly 7.0 wheels from the official pytorch index until TheRock releases with multi-device wheels.

jammm:
> expandable_segments

It seems like it's not supported properly for ROCm. But this shouldn't be a blocker for initial support I'd say.

jammm:
@LokiMetaSmith I've pushed a fix for the issues that @jon-hotaisle faced in https://github.com/jammm/nanochat/commit/e3e21e2a4139fe45b52572cfba5d34c26d54c7f3

* Use ROCm 6.4
* Use python 3.12 (maybe this isn't necessary but for some reason uv wasn't able to fully install the env)
* Remove `export HSA_OVERRIDE_GFX_VERSION=11.0.0` (this can be added again with arch detection to see if it's gfx11 based first, though I highly recommend users use TheRock wheels instead to avoid setting this env var altogether)
* Use `nccl` as the distributed backend (see https://github.com/karpathy/nanochat/pull/23#discussion_r2431800107)

nanochat/common.py
```diff
@@ -92,31 +92,41 @@ def get_dist_info():
 def compute_init():
     """Basic initialization that we keep doing over and over, so make common."""
 
-    # CUDA is currently required
-    assert torch.cuda.is_available(), "CUDA is needed for a distributed run atm"
+    # Detect hardware
+    if hasattr(torch.version, 'hip') and torch.version.hip and torch.cuda.is_available():
+        device_type = "cuda" # ROCm uses cuda naming in torch
+        backend = "rccl"
```
jammm, line 98: This will throw an error because backend is still `nccl` for ROCm. PyTorch automatically maps it to RCCL.
Just edited this to `nccl` on a mi300x node and it works fine.

amd-pratmish:
> export HSA_OVERRIDE_GFX_VERSION=11.0.0



> Confirmed. Seeing the same issue on a box of MI355x in a recent container: docker.io/rocm/pytorch:rocm7.0.2_ubuntu24.04_py3.12_pytorch_release_2.8.0
> 
> 2025-10-15 at 01 58 40 png

@jon-hotaisle : Were you able to run it on MI355s/MI325s?

LokiMetaSmith:
it looks like some of the memory issue is that linux kernel doesn't automatically let you assign the full 128gb, I ran this, and it cleared the running out of memory issue, but there are other problems still. 

This is strix halo specific.

https://www.reddit.com/r/LocalLLaMA/comments/1mib7l9/pytorch_on_rocm_v650rc_gfx1151_amd_strix_halo/

jon-hotaisle:
Would be good to either add the `source` to the docs or to the script...

2025-10-17 at 00 08 04 png

jon-hotaisle:
Also compare the work done here by @indianspeedster https://github.com/karpathy/nanochat/compare/master...indianspeedster:nanochat:master

It runs on MI300x, but not on MI355x.

jon-hotaisle:
2025-10-17 at 00 15 03 png

2025-10-17 at 00 15 24 png

jammm:
> it looks like some of the memory issue is that linux kernel doesn't automatically let you assign the full 128gb, I ran this, and it cleared the running out of memory issue, but there are other problems still.
> 
> This is strix halo specific.
> 
> https://www.reddit.com/r/LocalLLaMA/comments/1mib7l9/pytorch_on_rocm_v650rc_gfx1151_amd_strix_halo/

I helped build those wheels in the link. They’re quite old. Please use the ones from https://github.com/ROCm/TheRock/blob/main/RELEASES.md#torch-for-gfx1151 instead.

jammm:
`@jon-hotaisle` can you try my branch ? https://github.com/jammm/nanochat/tree/rocm-support

The one you used from https://github.com/karpathy/nanochat/compare/master...indianspeedster:nanochat:master has hard-coded export HSA_OVERRIDE_GFX_VERSION=9.4.2 which is why it’s not running on your mi355. My branch removes that line altogether from speedrun.sh 

Make sure to do a fresh clone and also delete the ~/.cargo folder. 

Also, can you confirm which rocm version you have installed ? If you’re on 7.0, there’s a good chance it won’t work because the changes here are using 6.3/6.4 (my branch uses 6.4)

it’ll need a minor tweak to pyproject.toml to use the nightly wheels that supports rocm 7.

jon-hotaisle:
@jammm Thanks, yea, your branch doesn't work for me. I'm on MI355x which pretty much requires ROCm7 for everything.

There are a few things to fix:

1. auto source the env.
2. Make sure to install unzip.
3. It seems it is installing the nvidia python stuff, so I just hit ctrl-c.

I'm trying to run in a recent container:

docker.io/rocm/pytorch:rocm7.0.2_ubuntu24.04_py3.12_pytorch_release_2.8.0

Since I'm in a container, I don't need to remove the ~/.cargo as it isn't kept between runs.

This is how I start the container, keeping the ~/.cache is definitely helpful.

```
podman run -it --privileged --network=host --ipc=host -v $HOME:/workdir -v \
/shareddata:/shareddata -v /shared/apps:/shared/apps -v $HOME/.cache:/root/.cache \
--workdir /workdir \
docker.io/rocm/pytorch:rocm7.0.2_ubuntu24.04_py3.12_pytorch_release_2.8.0 bash
```

2025-10-17 at 04 21 05 png

jammm:
@jon-hotaisle I see. Can you try changing this line https://github.com/jammm/nanochat/blob/rocm-support/pyproject.toml#L37

`url = "https://download.pytorch.org/whl/rocm6.4"`

to
`url = https://download.pytorch.org/whl/nightly/rocm7.0`

Delete the `.venv` folder in nanochat and try speedrun.sh again?

jon-hotaisle:
Ok, I screwed up, I wasn't actually on your branch when I cloned your repo. That was stupid. =)

Now it installs torch, but it should already have torch in my container.

But... best news... it looks like it is running now!

2025-10-17 at 05 16 12 png

jon-hotaisle:
Only one GPU is being used!? 

2025-10-17 at 05 37 00 png

jammm:
> Only one GPU is being used!?

That's because of the changes in speedrun.sh, which replaced the `torchrun --nproc-per-node=8` with just `python` to make it compatible with single consumer GPU setups. https://github.com/karpathy/nanochat/pull/23/files#diff-f01ca501612c5f260e12ac1171d6705e6887825cba06360c99429531072ef130

You could undo those changes to get it working with all 8 GPUs. I did that on the mi300x. Ideally there should be a flag to toggle this.

jon-hotaisle:
8x cooking now!

2025-10-17 at 06 12 20 png

jammm:
> 8x cooking now!

Awesome. Once that's done (takes a couple hours or so), you can `source .venv/bin/activate` then run `python -m scripts.chat_web` to run the web UI.

Qubitium:
@jon-hotaisle  Your vram usage is too low ~84GB out of 290GB usage. You need to tweak values as much as possible to so it reaches max vram without oom. Otherwise, you can't get close to max mfu. Easier is to switch batch from 32 (nanochat default) to 64 or 128. If 64 works but 128 oom but vram still not full, you can then try to increase per batch token count.

jon-hotaisle:
@Qubitium helpful feedback, thank you! first goal was just to get it running, now we can focus on tuning. should this be something that is added to the PR? Have it set sane defaults on different architectures too?

speedrun.sh
```diff
@@ -92,25 +97,25 @@ echo "Waiting for dataset download to complete..."
 wait $DATASET_DOWNLOAD_PID
 
 # pretrain the d20 model
-torchrun --standalone --nproc_per_node=8 -m scripts.base_train -- --depth=20 --run=$WANDB_RUN
+python -m scripts.base_train -- --depth=20 --run=$WANDB_RUN
```
svlandeg, line 100: I don't think we'll want this edit either way - the assumption being that this usually will be run on more than one GPU, and those that have only a single one can just edit this on their own fork.

svlandeg (commented):
Hi @LokiMetaSmith! It looks like a lot of the edits from this PR have already been addressed by https://github.com/karpathy/nanochat/pull/88. There's quite a bit of merge conflicts as well. I suggest to close this one and open new PR(s) with any remaining issues you'd encounter with the latest version of `master`.

PS: discussions about different setups & recommendations etc are nice, and it's good to see how this i being run on different platforms & setups. Please use the [forum](https://github.com/karpathy/nanochat/discussions) to continue such conversations!

PR #24, opened 2025-10-14
Add script for training using Modal

broqdev:


svlandeg (commented):
Hi @broqdev, thanks for the PR!

I don't think we want to include this in the core codebase of `nanochat` though, as the aim is to be a minimal implementation and not an extensive framework with support for various cloud providers. As such, I'll go ahead and close this.

PR #29, opened 2025-10-14
[FIX] export env variable in speedrun.sh

burtenshaw:
turbo nit, but `dataset.py` will use the incorrect location if this bash variable is not exported beyond the script.


karpathy:
just merged a pr for this ty

PR #30, opened 2025-10-14
Add unit tests for RustBPE implementation

msuiche:
Currently, tests are executed using Python commands:
```
python -m pytest tests/test_rustbpe.py -v -s
```
This enhancement adds unit tests directly within the Rust codebase, enabling the use of:
```
cargo test
```
which makes more sense since it's a rust library.


willhama:
@msuiche what are the additiona lfiles you ahve in your PR. The mds are not relevant for the intended purpose.

PR #32, opened 2025-10-14
use `pyarrow.fs` to download parquet files from the huggingface hub

kszucs:
`pyarrow.fs.copy_file("hf://datasets/karpathy/fineweb-edu-100b-shuffle/{filename.parquet}", dest)` can be used to download the parquet files from HF hub. 

The implementation uses the [HfFileSystem](https://huggingface.co/docs/huggingface_hub/en/package_reference/hf_file_system) `fsspec` implementation underneath, with builtin handling for retries and automatic backoff.

### DIFF
--- a/nanochat/dataset.py
+++ b/nanochat/dataset.py
@@ -11,6 +11,7 @@ import os
 import argparse
 import time
 import requests
+import pyarrow.fs as fs
 import pyarrow.parquet as pq
 from multiprocessing import Pool
 
@@ -20,7 +21,7 @@ from nanochat.common import get_base_dir
 # The specifics of the current pretraining dataset
 
 # The URL on the internet where the data is hosted and downloaded from on demand
-BASE_URL = "https://huggingface.co/datasets/karpathy/fineweb-edu-100b-shuffle/resolve/main"
+BASE_URI = "hf://datasets/karpathy/fineweb-edu-100b-shuffle"
 MAX_SHARD = 1822 # the last datashard is shard_01822.parquet
 index_to_filename = lambda index: f"shard_{index:05d}.parquet" # format of the filenames
 base_dir = get_base_dir()
@@ -68,45 +69,17 @@ def download_single_file(index):
         return True
 
     # Construct the remote URL for this file
-    url = f"{BASE_URL}/{filename}"
+    uri = f"{BASE_URI}/{filename}"
     print(f"Downloading {filename}...")
-
-    # Download with retries
-    max_attempts = 5
-    for attempt in range(1, max_attempts + 1):
-        try:
-            response = requests.get(url, stream=True, timeout=30)
-            response.raise_for_status()
-            # Write to temporary file first
-            temp_path = filepath + f".tmp"
-            with open(temp_path, 'wb') as f:
-                for chunk in response.iter_content(chunk_size=1024 * 1024):  # 1MB chunks
-                    if chunk:
-                        f.write(chunk)
-            # Move temp file to final location
-            os.rename(temp_path, filepath)
-            print(f"Successfully downloaded {filename}")
-            return True
-
-        except (requests.RequestException, IOError) as e:
-            print(f"Attempt {attempt}/{max_attempts} failed for {filename}: {e}")
-            # Clean up any partial files
-            for path in [filepath + f".tmp", filepath]:
-                if os.path.exists(path):
-                    try:
-                        os.remove(path)
-                    except:
-                        pass
-            # Try a few times with exponential backoff: 2^attempt seconds
-            if attempt < max_attempts:
-                wait_time = 2 ** attempt
-                print(f"Waiting {wait_time} seconds before retry...")
-                time.sleep(wait_time)
-            else:
-                print(f"Failed to download {filename} after {max_attempts} attempts")
-                return False
-
-    return False
+    try:
+        # pyarrow.fs uses huggingface_hub with builtin exponential backoff
+        fs.copy_files(uri, filepath)
+    except (requests.RequestException, IOError) as e:
+        print(f"Failed to download {filename}: {e}")
+        return False
+    else:
+        print(f"Successfully downloaded {filename}")
+        return True
 
 
 if __name__ == "__main__":
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,6 +10,7 @@ dependencies = [
     "files-to-prompt>=0.6",
     "numpy==1.26.4",
     "psutil>=7.1.0",
+    "pyarrow>=21.0.0",
     "regex>=2025.9.1",
     "tiktoken>=0.11.0",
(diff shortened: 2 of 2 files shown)


nanochat/dataset.py
```diff
@@ -68,45 +69,17 @@ def download_single_file(index):
...
-    return False
+    try:
+        # pyarrow.fs uses huggingface_hub with builtin exponential backoff
+        fs.copy_files(uri, filepath)
+    except (requests.RequestException, IOError) as e:
+        print(f"Failed to download {filename}: {e}")
+        return False
+    else:
+        print(f"Successfully downloaded {filename}")
+        return True
```
svlandeg, line 82: Instead of mixing an `else` condition with the `try`-`except`, simply add the happy path entirely to the `try` clause. 
```suggestion
    try:
        # pyarrow.fs uses huggingface_hub with builtin exponential backoff
        fs.copy_files(uri, filepath)
        print(f"Successfully downloaded {filename}")
        return True
    except (requests.RequestException, IOError) as e:
        print(f"Failed to download {filename}: {e}")
        return False
```

nanochat/dataset.py
```diff
@@ -68,45 +69,17 @@ def download_single_file(index):
...
-                time.sleep(wait_time)
-            else:
-                print(f"Failed to download {filename} after {max_attempts} attempts")
-                return False
-
-    return False
+    try:
+        # pyarrow.fs uses huggingface_hub with builtin exponential backoff
+        fs.copy_files(uri, filepath)
+    except (requests.RequestException, IOError) as e:
```
svlandeg, line 77: This `except` statement is copy-pasted from the original code, which was using `requests.get()`. Does `fs.copy_files()` raise the exact same type of exceptions, or does this need updating?

pyproject.toml
```diff
@@ -10,6 +10,7 @@ dependencies = [
     "files-to-prompt>=0.6",
     "numpy==1.26.4",
     "psutil>=7.1.0",
+    "pyarrow>=21.0.0",
```
svlandeg, line 13: `pyarrow` automatically comes with `datasets`, right? So we don't really need to specify this separately?

ptindall-rxfoundry: It doesn't look like datasets pins any versions of pyarrow, so this may have been needed to ensure a minimum version for the downloader functionality?

nanochat/dataset.py
ptindall-rxfoundry: Agree with @svlandeg comment.  No need to have the extra else condition here.

ptindall-rxfoundry:
On the logging issue, I am seeing the same verbosity with just INFO setting on Ubuntu, but oddly not on MacOS.  This can likely be controlled through some logger settings.

In general, I have been seeing data corruption issues on both Ubuntu and MacOS for different download files, which makes me think the current implementation has some issues.  After adding the PR changes here to my forked repo, the corruption issues seem to have been fixed on both Ubunutu and MacOS.  As such, I think this is a very worthy PR to pull in to the upstream repo and would help out to support getting it in.

kszucs:
> Hi!

Hi @svlandeg!
> 
> Just a few quick comments/suggestions. You mention that `pyarrow.fs.copy_files()` has builtin handling for retries and automatic backoff. Can you point to where this is documented? 

`pyarrow` supports multiple filesystems, including Hugging Face through fsspec which has a HF implementation https://huggingface.co/docs/huggingface_hub/en/package_reference/hf_file_system

> Can this behaviour be configured? (e.g. number of retries).

Good question, if not we can certainly add support for it. @lhoestq is it configurable?

> Have you done any comparisons whether the current approach is more or less performant than the proposed approach? (in terms of speed & mem usage)

No, though if we want to focus on efficiency we can leverage HF's xet storage by using [hf_hub_download](https://huggingface.co/docs/huggingface_hub/v1.2.3/en/package_reference/file_download#huggingface_hub.hf_hub_download), see the following links for more details:
- https://huggingface.co/docs/huggingface_hub/en/guides/download#faster-downloads 
- https://huggingface.co/docs/hub/en/xet/index
- https://huggingface.co/blog/parquet-cdc

> When running the code from this PR, it looks like the logs get pretty swamped with httpx messages, 
> Is there a way to reduce the output on the console so it's not so overwhelming?

We can probably influence that through the HfApi configuration, cc @lhoestq.

nanochat/dataset.py
kszucs: Sure, going to update after we discussed the right approach. HF provides more utilities for efficient downloads, see below.

pyproject.toml
kszucs: `datasets` depends on `pyarrow` and the pin is required to support `hf://` protocol uri which is available since `pyarrow 21.0.0`, see https://github.com/apache/arrow/pull/45089

kszucs:
Hi @ptindall-rxfoundry!

> On the logging issue, I am seeing the same verbosity with just INFO setting on Ubuntu, but oddly not on MacOS. This can likely be controlled through some logger settings.

We should be able the log level using https://huggingface.co/docs/huggingface_hub/en/package_reference/utilities#huggingface_hub.utils.logging.get_verbosity

> 
> In general, I have been seeing data corruption issues on both Ubuntu and MacOS for different download files, which makes me think the current implementation has some issues. After adding the PR changes here to my forked repo, the corruption issues seem to have been fixed on both Ubunutu and MacOS. As such, I think this is a very worthy PR to pull in to the upstream repo and would help out to support getting it in.

We have multiple options to improve the user experience, like faster downloads thanks to Xet and/or reusing local HF cache. cc @lhoestq for more ideas.

ptindall-rxfoundry:
Putting this at the top of the dataset.py file suppresses the INFO logging statements:

```
import logging
logging.getLogger("httpx").setLevel(logging.ERROR)
```

lhoestq:
> Good question, if not we can certainly add support for it. @lhoestq is it configurable?

default is 5 but it could surely be made configurable cc @wauplin

> We can probably influence that through the HfApi configuration, cc @lhoestq.

indeed as @ptindall-rxfoundry suggested you can set the httpx logger level

> We have multiple options to improve the user experience, like faster downloads thanks to Xet and/or reusing local HF cache. cc @lhoestq for more ideas.

[hf_hub_download](https://huggingface.co/docs/huggingface_hub/v1.2.3/en/package_reference/file_download#huggingface_hub.hf_hub_download) is the best option here as it uses Xet and it has a cache. `pyarrow.fs` on the other hand doesn't cache anything (but it does use Xet)

Wauplin:
> default is 5 but it could surely be made configurable cc @Wauplin

Yes indeed, we do retries but not in a configurable way at the moment. Would this prove useful in your case? Default of 5 is usually enough

svlandeg:
Hi all! This has become a bit stale/outdated, so I'll go ahead and close this for now 🙏

PR #33, opened 2025-10-14
Skip sync across GPUs for all but last gradient accumulation step

aflah02:
Sync is not needed at every gradient accumulation step, and this change introduces the no_sync context when we're not at the final gradient accumulation step.

This optimization is inline with other implementations such as [fairseq](https://github.com/facebookresearch/fairseq/blob/main/fairseq/trainer.py#L820)
, which use model.no_sync() to disable gradient synchronization until the final micro-step.

This reduces redundant gradient communication by roughly $(N - 1) / N$  (where N is the number of gradient accumulation steps) resulting in a substantial efficiency gain for DDP training.

### DIFF
--- a/scripts/chat_sft.py
+++ b/scripts/chat_sft.py
@@ -16,6 +16,7 @@ import copy
 import wandb
 import torch
 import torch.distributed as dist
+from contextlib import nullcontext
 
 from nanochat.common import compute_init, compute_cleanup, get_base_dir, print0, DummyWandb
 from nanochat.checkpoint_manager import load_model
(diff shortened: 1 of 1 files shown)


karpathy:
If I may accuse this PR of vibe coding AI slop?

Note that nanochat is not using the DDP container (otherwise common in PyTorch), so this context manager does not apply. If you look at the implementations of AdamW and Muon used for optimization, you'll see that the synchronization happens inside the `.step()` call, which happens only a single time after all the gradient accumulation is done.

So my impression is that the code is correct as is.

I asked Copilot out of curiosity to review the PR, and it also completely missed it.

We're entering a new era of GitHub and I'm not sure I like it.

aflah02:
Oh I totally missed that

So I knew about this being a thing people miss as I missed it in some of my personal code and I didn't care to look further. My bad I should've done that

Closing it!

aflah02:
Also maybe just for context I did use AI to double check the changes and it missed the same thing but I should've taken a closer look myself

The idea itself is something I remembered from my prior mistakes

PR #35, opened 2025-10-14
fix: return inf instead of crashing when evaluate_bpb has zero total_bytes

bhaskar0210s:
Edge case: all tokens are special tokens or ignored

Return infinity to indicate no meaningful bytes were processed


svlandeg (approved):
Makes sense to make the code more robust here. Alternatively, we could add an epsilon to the divisor on L62.

PR #36, opened 2025-10-14
Refactor: Remove pandas dependency from base_eval.py & Improve DDP detection in common.py

azekowka:
This PR addresses the TODO item in scripts/base_eval.py by removing the dependency on the pandas library. Specifically, the pd.read_csv call used to load eval_meta_data.csv has been replaced with Python's built-in csv module.

Changed is_ddp() function in nanochat/common.py. Before, it only checked the RANK environment variable to infer if Distributed Data Parallel (DDP) was active. The updated function now first imports torch.distributed as dist. Then, inside is_ddp(), it first calls dist.is_initialized(). If that returns True (meaning PyTorch's distributed backend is active), the function immediately returns True. If dist.is_initialized() returns False, it then falls back to the original check of the RANK environment variable.

### DIFF
--- a/nanochat/common.py
+++ b/nanochat/common.py
@@ -76,7 +76,8 @@ def print_banner():
     print0(banner)
 
 def is_ddp():
-    # TODO is there a proper way
+    if dist.is_initialized():
+        return True
     return int(os.environ.get('RANK', -1)) != -1
 
 def get_dist_info():


karpathy:
Like, will get around to later today once i can validate. Sadly probably won't actually remove pandas dep because some of the other deps likely need it.

scripts/base_eval.py
```diff
@@ -38,11 +38,20 @@ def evaluate_model(model, tokenizer, device, max_per_task=-1):
...
+
+    # Load eval metadata
+    eval_metadata = {}
+    with open(eval_meta_data_path, 'r') as f:
+        reader = csv.reader(f)
+        header = next(reader) # Skip header
+        for row in reader:
+            task_name = row[0]
+            random_baseline = float(row[1])
+            eval_metadata[task_name] = {"Random baseline": random_baseline}
```
svlandeg, line 54: It was decided in #192 to keep the call to `pd.read_csv` for now

nanochat/common.py
```diff
@@ -76,7 +76,8 @@ def print_banner():
     print0(banner)
 
 def is_ddp():
-    # TODO is there a proper way
+    if dist.is_initialized():
+        return True
```
svlandeg, line 80: cf comment [here](https://github.com/karpathy/nanochat/pull/54#issuecomment-3406382852)

PR #38, opened 2025-10-14
feat(engine.py): sample unique tokens per row in generation stream

azekowka:
Before, when initiating a batch generation, the first sampled token was broadcasted to all rows. This change updates the logic to ensure that a unique token is sampled for each row, improving the diversity and independence of generated sequences within a batch.

### DIFF
--- a/nanochat/engine.py
+++ b/nanochat/engine.py
@@ -185,7 +185,7 @@ class Engine:
             seq_len=len(tokens),
             **kv_model_kwargs,
         )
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
Thanks for the PR @azekowka! Can you remove the unrelated edits from PR 36 from this PR?

azekowka:
Thank you @svlandeg for reviewing! I closed this PR, and created a new one, without unrelated edits from PR 36. Here is the link - https://github.com/karpathy/nanochat/pull/201

PR #39, opened 2025-10-14
docs: Remove comma before "that" in README pipeline sentence

TheNomadInOrbit:
Improve readability by removing unnecessary comma before "that" in  pipeline sentence for better grammar and flow.


svlandeg (commented):
This feels like a matter of preference, I think either is fine, so let's just keep it as-is.

PR #40, opened 2025-10-14
add modular imports to the init

burtenshaw:
This PR add modular imports to nanochat's `__init__` this will allow users to use nanochat code without relative imports. For example:

```python
from nanochat.tokenizer import RustBPETokenizer

tokenizer = RustBPETokenizer.from_directory(tokenizer_dir="tokenizer")
```

I have copied the same usage as `scripts` to avoid having to change any further code.

### DIFF
--- a/nanochat/__init__.py
+++ b/nanochat/__init__.py
@@ -1,0 +1,33 @@
+# Import all submodules used by scripts
+from . import common
+from . import tokenizer
+from . import checkpoint_manager
+from . import core_eval
+from . import gpt
+from . import dataloader
+from . import loss_eval
+from . import engine
+from . import dataset
+from . import report
+from . import adamw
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
Hi Ben!

Can you elaborate on why exactly you'd need this? Don't the scripts run fine as is?

Also, if we do want to have this, I'd prefer sorting them alphabetically to facilitate future maintenance.

burtenshaw:
Thanks for pinging!

> Can you elaborate on why exactly you'd need this? Don't the scripts run fine as is?

At the time, I was experimenting with importing with functions into new scripts and notebooks. For example, to load the tokenizer as above. I don't think it makes much sense to use the repo like that any more, so I'll close.

PR #43, opened 2025-10-14
fix: add uv to PATH to avoid command not found

amrrs:


Aymane11:
uv is by [default](https://docs.astral.sh/uv/reference/installer/) installed to `~/.local/bin` which is -generally- in `PATH` by default

speedrun.sh
```diff
@@ -20,6 +20,8 @@ mkdir -p $NANOCHAT_BASE_DIR
 
 # install uv (if not already installed)
 command -v uv &> /dev/null || curl -LsSf https://astral.sh/uv/install.sh | sh
+echo 'export PATH="$HOME/.local/bin:$PATH"' >> ~/.bashrc
```
vikhyat, line 23: Don't think it's a good idea to mutate the user's environment.

mahimairaja: And it varies from POSIX to UNIX

Issue #11, opened 2025-10-13
Kerning on logo 't' is too large


mykeln:
The "t" in "nanochat" needs to be shifted one character to the left.

PR #49, opened 2025-10-14
Update logo

obxium:
Fantastic project and excited for the future of Eureka Labs!

This is a small update to the logo to address the kerning distance with the letter 't' as described in #11. 😸

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ __pycache__/
 *.pyc
 rustbpe/target/
 dev-ignore/
+report.md
+eval_bundle/
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2025 Andrej Karpathy
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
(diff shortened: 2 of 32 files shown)


karpathy:
Haha but that makes the logo inconsistent in banner and in code!

obxium:
> Haha but that makes the logo inconsistent in banner and in code!

My bad and I should have tried harder/went to the source first.

I updated the code version to match.

Cheers!

karpathy:
Curious why did the https://manytools.org/hacker-tools/ascii-banner/ render it the original way.

obxium:
> Curious why did the https://manytools.org/hacker-tools/ascii-banner/ render it the original way.

Not sure, the website seems to use Figlet under the hood. I can check the `figlet` CLI render for the font and compare though if that would help. Ultimately, it does look better from a typography point of view, but it's your call to approve.

I'm just happy to be chatting in here with you about it. Mad respect! 🫡

obxium:
Okay, I dove into this a little, and it turns out that the website you linked uses [figlet.js](https://github.com/patorjk/figlet.js) under the hood which like the original figlet CLI, has [kerning options](https://github.com/patorjk/figlet.js?tab=readme-ov-file#understanding-kerning), so if one directly uses the library instead of the website, you could possibly programmatically control the kerning of the output should you need to dynamically generate the logo or recreate it from the source.

That said, I tried the [OG Figlet](https://www.figlet.org/) CLI's `-k` flag for kerning and it does well on kerning the 't' in the output in the same way as I manually did for you, as shown in this screenshot:

kern-example

So the TL;DR is that the website doesn't seem to support kerning, but the underlying JS library does although I did not test it. The OG Figlet definitely supports kerning and seems to do a decent enough job.

One more thing: if you're interested in the original Rebel font, I found [an example of the font](https://raw.githubusercontent.com/ankushbhgaat/figlet-fonts/refs/heads/master/DOS%20Rebel.flf) with comments attributing its creation to Valerie Mates, so shout out to Valerie for an awesome Figlet font!

PR #50, opened 2025-10-14
add support for mps (Metal Performance Shaders) in web and cli to enable GPU accelerated inference on Macs

KODIKAS-NL:
Hi ! nice project.
I made some small changes so you can instruct torch to use MPS ( https://docs.pytorch.org/docs/stable/notes/mps.html ) instead of Cuda to enable accelerated inference on Macs
I added an extra argument (-d) to the web and cli script to support device selection.

### DIFF
--- a/nanochat/common.py
+++ b/nanochat/common.py
@@ -98,7 +98,7 @@ def compute_init(device_type="cuda"):
     # Reproducibility
     torch.manual_seed(42)
     if device_type == "cuda":
-    torch.cuda.manual_seed(42)
+        torch.cuda.manual_seed(42)
     elif device_type == "mps":
         torch.mps.manual_seed(42)
     # skipping full reproducibility for now, possibly investigate slowdown later
--- a/scripts/chat_web.py
+++ b/scripts/chat_web.py
@@ -29,7 +29,7 @@ parser.add_argument('-g', '--model-tag', type=str, default=None, help='Model tag
 parser.add_argument('-s', '--step', type=int, default=None, help='Step to load')
 parser.add_argument('-p', '--port', type=int, default=8000, help='Port to run the server on')
 parser.add_argument('--host', type=str, default='0.0.0.0', help='Host to bind the server to')
-parser.add_argument('-d', '--device', type=str, default='cuda', help='Device to run the model on: cuda|mps')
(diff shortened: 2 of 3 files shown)


jstello:
This would be nice

nanochat/common.py
```diff
@@ -89,15 +89,18 @@ def get_dist_info():
...
     """Basic initialization that we keep doing over and over, so make common."""
 
     # CUDA is currently required
-    assert torch.cuda.is_available(), "CUDA is needed for a distributed run atm"
+    assert torch.cuda.is_available() | torch.mps.is_available(), "CUDA or MPS is needed for a distributed run atm"
 
     # Reproducibility
     torch.manual_seed(42)
+    if device_type == "cuda":
     torch.cuda.manual_seed(42)
```
Aymane11: i think you missed an indent here

scripts/chat_web.py
```diff
@@ -29,10 +29,11 @@
 parser.add_argument('-s', '--step', type=int, default=None, help='Step to load')
 parser.add_argument('-p', '--port', type=int, default=8000, help='Port to run the server on')
 parser.add_argument('--host', type=str, default='0.0.0.0', help='Host to bind the server to')
+parser.add_argument('-d', '--device', type=str, default='cuda', help='Device to run the model on: cuda|mps')
```
Aymane11: you can allow only "mps" or "cuda" by setting `choices=['mps', 'cuda']`
(see: https://docs.python.org/3/library/argparse.html#choices)

nanochat/common.py
KODIKAS-NL: Strange, the indent was there in my local file, but somehow didn't get checked in 🤔
Fixed it

scripts/chat_web.py
KODIKAS-NL: Of course 😊
Done!

karpathy:
ty, this was on my mind too, probably for both training/inference, hopefully relatively doable. will come back around to later today.

sailor239:
Do we need to update the dependencies first?

KODIKAS-NL:
> ty, this was on my mind too, probably for both training/inference, hopefully relatively doable. will come back around to later today.

I'll close my pull request since I saw you merged your own beautiful implementation yesterday 😊

PR #52, opened 2025-10-14
Fix incorrect SFT loss logging during gradient accumulation

notkisk:
This PR fixes a logging bug in the **SFT training script** where `train_loss` only reflected the **last micro-batch** of each gradient accumulation cycle instead of the average across all micro-batches

The fix accumulates detached losses on-device and computes their mean after all accumulation steps  ensuring accurate loss reporting without affecting gradients or training behavior

---

### **Before**

```python
train_loss = loss.detach()  # only logs last micro-batch
```

### **After**

```python
accumulated_loss += loss.detach()
...
train_loss = accumulated_loss / grad_accum_steps
```

---

**Impact:**
✅ Correct average loss logging
✅ No change to gradient correctness or training dynamics

### DIFF
--- a/scripts/chat_sft.py
+++ b/scripts/chat_sft.py
@@ -225,6 +225,8 @@ for step in range(num_iterations):
 
     # average loss across micro-steps for logging (tensor, still on device)
     train_loss = accumulated_loss / grad_accum_steps
+    if ddp:
+        dist.all_reduce(num_tokens, op=dist.ReduceOp.SUM) # sum over ranks
 
     # learning rate scheduler
     lrm = get_lr_multiplier(step)


notkisk:
In the original code:

for micro_step in range(grad_accum_steps):
    ...
    train_loss = loss.detach()  # for logging
    loss = loss / grad_accum_steps
    loss.backward()


The line

train_loss = loss.detach()


overwrites the previous value each micro-step - so after the loop, train_loss only contains the last micro-batch’s loss, not the average across all micro-batches.

That means the loss logging (the one printed and sent to WandB) was incorrect.
It did not represent the true per-step average loss, even though the gradients and training behavior were correct (since the loss was divided before backward()).

karpathy:
It's not a bug, it's a deliberate choice. Doing "the right thing" adds a bunch of added bloat/complexity synchronizing the number. We're only seeing/logging the loss at a single rank but that's ok as just an estimate. The training is all correct otherwise.

PR #54, opened 2025-10-15
ddp check

geronimi73:
huge fix ;-)

### DIFF
--- a/nanochat/common.py
+++ b/nanochat/common.py
@@ -76,8 +76,7 @@ def print_banner():
     print0(banner)
 
 def is_ddp():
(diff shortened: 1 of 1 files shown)


karpathy:
I'm careful with this one because iirc I tried to use `is_initialized` earlier in the development and I recall it breaking something and I changed it back to current version. I can't remember what that was anymore. Maybe I'm making it up, I'll have to re-validate all runs ok.

svlandeg:
Related PR worth checking as well: https://github.com/karpathy/nanochat/pull/256

PR #56, opened 2025-10-15
Add hf-transfer dependency to pyproject.toml

jorahn:
to fix error after pre-training stage

```bash
[rank0]: ValueError: Fast download using 'hf_transfer' is enabled (HF_HUB_ENABLE_HF_TRANSFER=1) but 'hf_transfer' package is not available in your environment. Try `pip install hf_transfer`.
```
originating from [tasks/smoltalk.py](https://github.com/karpathy/nanochat/blob/67aaca98f53a6db9f1f5a2a62893f98d443a090a/tasks/smoltalk.py#L16)
```python
self.ds = load_dataset("HuggingFaceTB/smol-smoltalk", split=split).shuffle(seed=42)
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,7 @@ dependencies = [
     "torch>=2.8.0",
     "uvicorn>=0.36.0",
     "wandb>=0.21.3",
+    "hf-transfer>=0.1.9",
 ]
 
 [build-system]


sachin-aag:
Ran into this issue as well. Would be good to have this merged.

svlandeg (commented):
Hi @jorahn and @sachin-aag, thanks for the report!

As far as I can see, `HF_HUB_ENABLE_HF_TRANSFER` isn't being set on this repo. Could it be that you have this set in your environment from a different project? In a vanilla new `venv`, I'm not running into any errors when using/downloading `SmolTalk`.

According to the [HF documentation](https://huggingface.co/docs/huggingface_hub/en/package_reference/environment_variables#hfhubenablehftransfer), `HF_HUB_ENABLE_HF_TRANSFER` is actually deprecated and using `hf-transfer` not necessary anymore.

jorahn:
@svlandeg thanks for checking. i tested on a pretty clean new machine, but it must have slipped in from some setup scripts then, if its not in this repo.

svlandeg:
Ok, thanks for reporting back! I'll go ahead and close this, but feel free to reopen if this should surface again on a clean `venv`.

sachin-aag:
I was on a clean env, and still got the issue. I checked online and some other people have also gotten this issue: https://github.com/huggingface/hf_transfer/issues/63 in other repos. Not quite sure why it happens. 

I can add a code snippet like: 
```
# Disable hf_transfer if set externally (not needed for nanochat)
  import os
  os.environ.pop('HF_HUB_ENABLE_HF_TRANSFER', None)
```

Let me know if that makes sense. Would be better than installing hf-transfer, i guess as the intended behavior is to rely on xet now as per HF docs. 
@svlandeg @jorahn

svlandeg:
@sachin-aag: do you have env vars set on your system that may include this one from a previous project? What does `print(os.environ.get("HF_HUB_ENABLE_HF_TRANSFER"), None)` give you in a clean `venv` , before executing anything, and again right before calling `SmolTalk`?

I'd be hesitant to alter the user's environment during `nanochat` execution...

sachin-aag:
So just checked. I was on a fresh VM, but for some reason runpod defaults that variable to 1, even on a new VM. I will check with someone from the cloud provider side to understand what is going on. 
image

svlandeg:
Interesting @sachin-aag, thanks for investigating and reporting back! That confirms the suspicion that this is not a `nanochat` issue. I'd definitely refer the `runpod` people to the [HF documentation](https://huggingface.co/docs/huggingface_hub/en/package_reference/environment_variables#hfhubenablehftransfer) where it reads that this env var is now deprecated.

PR #63, opened 2025-10-15
Add error handling for missing checkpoint files in load_checkpoint function

cobanov:
Enhance the load_checkpoint function to raise FileNotFoundError for missing model, optimizer, and metadata files, improving robustness and user feedback.


svlandeg (commented):
Hi!

I'm not sure this PR adds much. If the `model_path` doesn't exist, `torch` will actually throw a very similar `FileNotFoundError` that will be equally clear to the user.

With respect to the `optimizer_path` and `meta_path` - typically those should exist if the model directory is correct.

I'll go ahead and close this one.

Issue #47, opened 2025-10-14
Question: Initializing rope embedding twice


sasikr2:
Why we do initialization of rotary embedding at two places?

```
# line 169 of gpt.py
cos, sin = self._precompute_rotary_embeddings(self.rotary_seq_len, head_dim)


#and at line 185 inside function of init_weights(...)
cos, sin = self._precompute_rotary_embeddings(self.rotary_seq_len, head_dim)
self.cos, self.sin = cos, sin
```

is this typo?

i-zaitsev:
I think the second one is a public method used when building model from a checkpoint:
https://github.com/karpathy/nanochat/blob/master/nanochat/checkpoint_manager.py#L77

karpathy:
So actually the first one is not an init because the init function runs with device set to 'meta'. So it's fake computation, it's just shapes and dtypes metadata. The second call actually runs it. This is all to support the meta device initialization...

PR #81, opened 2025-10-16
remove unused initialization of rotary embeddings

HosLak:
Fixed redundant reinitialization of `cos ` and `sin `buffers in `init_weights()`.
These buffers are already registered during model setup, so the extra initialization call was removed for cleaner weight initialization.


HosLak:
https://github.com/karpathy/nanochat/issues/47#issuecomment-3408356226

PR #85, opened 2025-10-16
Add CPU web server and OpenAI API

evgenys91:
Added script that provides CPU inference and OpenAI SDK compatibility.

### Why
Pretty useful (for us at least) to be able to run the model on a CPU machine (like laptop) and hit it with the OpenAI SDK from another server for testing. Saves spinning up GPU instances just to try things out.

### What works

The server now responds to `/v1/chat/completions` in the OpenAI format, so standard SDK calls just work:

```python
from openai import OpenAI

client = OpenAI(
    api_key="not_set",
    base_url="http://localhost:8000/v1"
)

response = client.chat.completions.create(
    model="nanochat",
    messages=[{"role": "user", "content": "Hello!"}]
)
```

**Supported:**
- `messages`, `temperature`, `max_tokens`, `top_k`, `stream`

Run with:
```bash
python -m scripts.chat_web_cpu --model-dir /tmp/nanochat
```


svlandeg (commented):
Thanks for the PR @evgenys91. This has now become obsolete after https://github.com/karpathy/nanochat/commit/da32e1d657b62cc1110f20f86fbc301d6189726e, so I'll go ahead and close this 🙏

PR #91, opened 2025-10-16
fix: handle None in report.py and correct README typo

MuraliChandran:
**Docs: correct typo in README**
   - Fixed a minor spelling mistake: “moden” -> “modern”.


nanochat/report.py
willhama, line 1: It should rather fail than returing an empty string

MuraliChandran: Thanks, I reverted back

svlandeg (commented):
Thanks @MuraliChandran - I'll go ahead and close this as this typo has already been covered by other PRs.

PR #93, opened 2025-10-17
Enable Proxy/Gateway Compatibility via Uvicorn's --root-path

wpybtw:
Description:

This PR implements the standard solution for deploying FastAPI behind proxies (like Jupyter tunnels, Nginx, etc.) where the public access URL contains a path prefix that is not part of the application itself.

Problem:

When deployed to an environment that forwards requests (e.g., from /notebook/user/proxy/8000/chat/completions to /chat/completions), the client-side API calls were failing with a 404 Not Found. This occurred because the browser didn't know the proxy prefix and was incorrectly calling /chat/completions relative to the domain root. The server was unable to determine the necessary prefix automatically.

Solution:

Configuration: Added the standard --root-path argument to argparse and passed it to uvicorn.run(). This allows users to launch the server with the required public prefix (e.g., python chat_web.py --root-path "/my/proxy/path").

ASGI Compliance: When --root-path is provided, Uvicorn correctly sets the request.scope['root_path'] variable for every incoming request, making the proxy prefix available to the application.

Dynamic UI Fix: The @app.get("/") endpoint now reads this root_path and injects it into the client's API_URL variable, ensuring all subsequent API calls are correctly prefixed and routed through the gateway.

### DIFF
--- a/scripts/chat_web.py
+++ b/scripts/chat_web.py
@@ -70,7 +70,7 @@ parser.add_argument('-g', '--model-tag', type=str, default=None, help='Model tag
 parser.add_argument('-s', '--step', type=int, default=None, help='Step to load')
 parser.add_argument('-p', '--port', type=int, default=8000, help='Port to run the server on')
 parser.add_argument('--host', type=str, default='0.0.0.0', help='Host to bind the server to')
-parser.add_argument('--root-path', type=str, default='', help='ASGI root path for proxy/gateway configurations.')
+parser.add_argument('--root-path', type=str, default='', help='ASGI root path for proxy/gateway configurations')
 args = parser.parse_args()
 
 # Configure logging for conversation traffic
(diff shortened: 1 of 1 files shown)


scripts/chat_web.py
```diff
@@ -224,19 +225,21 @@ async def lifespan(app: FastAPI):
...
-    # Replace the API_URL to use the same origin
+
+    # Get the prefix provided by the proxy/ASGI server.
+    proxy_prefix = request.scope.get('root_path', '').rstrip('/')
+
     html_content = html_content.replace(
-        "const API_URL = `http://${window.location.hostname}:8000`;",
-        "const API_URL = '';"
+        "const API_URL = '';",
+        f"const API_URL = '{proxy_prefix}';"
```
svlandeg, line 239: While reviewing this PR, I've gotten confused with the current state of this code on `master`, which replaces `http://${window.location.hostname}:8000` with an empty string. But `ui.html` already has `const API_URL = ''` in it, so nothing is being replaced on `master`. 

With this PR, the empty string would get replaced with the provided `root_path`.

svlandeg: Update: I think the original `replace` call here on `master` was non-functional and obsolete, but it does make sense to have it in the form this PR has it.

Kludex:
Although this seems right, I also get very confused about the `root-path` and proxy related settings. I suggest creating a small nginx file, and trying this PR out - and sharing it here so we can also try it as part of the review process. 🙏

svlandeg:
@wpybtw: can you look into providing a repro as Kludex mentioned?

svlandeg:
This has gotten a bit stale, and it's not clear how big an issue this really is. I'll go ahead and close this for now.

Issue #96, opened 2025-10-17
TODO: better way to check is_ddp


itsAnanth:
instead of using `os.environ.get('RANK', -1)` use `torch.distributed.is_available()` and `torch.distributed.is_initialized()` to get status regarding whether ddp is being used or not

internally `is_initialized()` function checks if group world is initialized or not

PR #97, opened 2025-10-17
chore: updated is_ddp to use more robust checks

itsAnanth:
Refactor is_ddp function to check whether ddp is being used or not #96


itsAnanth:
Got it 👍🏻

PR #103, opened 2025-10-17
Add automatic GPU detection for accurate MFU calculation

teelaitila:
Fixes inaccurate MFU reporting on non-H100 hardware.
   
   - Add `get_peak_flops()` function to auto-detect GPU and return peak BF16 Tensor FLOPS
   - Supports most common Nvidia GPUs
   - Defaults to H100 for unknown GPUs

### DIFF
--- a/nanochat/common.py
+++ b/nanochat/common.py
@@ -150,26 +150,21 @@ def get_peak_flops(device=None) -> float:
         "L40S": 362e12,
     }
     
-    try:
-        if device is None:
-            device = torch.device("cuda:0")
-        device_name = torch.cuda.get_device_name(device)
-        
-        # Match GPU by substring (case-insensitive). Sort by length to check specific names first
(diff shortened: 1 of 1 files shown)


nanochat/common.py
```diff
@@ -126,6 +126,51 @@ def compute_cleanup():
...
+        "6000 Blackwell": 503.8e12,
+        "6000 Ada": 364e12,
+        "A6000": 154.8e12,
+        "RTX 5090": 209.5e12,
+        "RTX 4090": 165.2e12,
+        "RTX 3090": 71e12,
+        "L40S": 362e12,
+    }
+    
+    try:
```
karpathy: try catch? smells like vibe coding

teelaitila: Inexperienced and using ai for help. Obviously dumb to use it here on second look. Removing it

teelaitila: Also wondering if adding amd gpus or apple chips would be good or add too much bloat?

PR #108, opened 2025-10-18
Fix chat_web.py to support mps device type

shivamojha2:
I was testing the web chat server on MPS after training on CUDA and noticed that `compute_init()` was being called without the required `device_type` parameter.

This PR adds the missing `autodetect_device_type()` call before `compute_init()` and updates the worker pool initialization to properly support CUDA and MPS systems.

Thanks for putting this together!


karpathy:
There is a PR up where I'm trying to implement this properly. Readme links to it.

karpathy:
reference the cpu|mps branch. closing.

PR #113, opened 2025-10-19
export NANOCHAT_BASE_DIR so child processes get it too

tlepoint:
Similar to #48 but for the run1000.sh script.


karpathy:
doh ty

PR #116, opened 2025-10-19
Fix a memory leak and incorrect sort in rustbpe

mitsuhiko:
I came across this while reading the rustbpe sources.

This fixes a memory leak on the iterator (`PyObject_GetIter` returns an owned references) and what I believe to be a bug in the tie-braker sort.  At least the comment implies that it would work the other way round.

### DIFF
--- a/rustbpe/src/lib.rs
+++ b/rustbpe/src/lib.rs
@@ -116,7 +116,7 @@ impl Ord for MergeJob {
             self.count.cmp(&other.count)
         } else {
             // ascending order on the pair when counts tie
-            other.pair.cmp(&self.pair)
+            self.pair.cmp(&other.pair)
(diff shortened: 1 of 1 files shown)


karpathy:
Fixed the memory leak here:

https://github.com/karpathy/nanochat/commit/9467d83cf23dcc9a9b4ca6e35103142f48a55b27

ty. I'm not 100% sure but I think the tie breaking is probably correct as is w.r.t. intended functionality.

mitsuhiko:
Yeah I was not really sure about the sort.  I only looked at it because I usually double and triple check all custom `Ord` implementations in Rust because if they don't have total order, they can cause a panic in the sort function.  (But this sort implements a total order)

PR #123, opened 2025-10-20
fix: remove unnecessary tensor allocation in DistAdamW optimizer

SermetPekin:
This fix optimizes the DistAdamW optimizer by removing unnecessary tensor allocations during gradient reduction and parameter updates.

### DIFF
--- a/nanochat/adamw.py
+++ b/nanochat/adamw.py
@@ -26,7 +26,6 @@ class DistAdamW(torch.optim.Optimizer):
         grad_slices = []
         for group in self.param_groups:
             params: list[Tensor] = group["params"]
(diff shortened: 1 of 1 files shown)


karpathy:
Yeah this was part of a code chunk that I yoinked from modded-nanogpt and it looked wrong but I didn't want to deal with it in that moment so I just left a TODO for later. I agree it's wrong and wasteful. I don't think I'm missing any galaxy-brained reason for it.

SermetPekin:
Hi @karpathy thanks for the review.  In the next PR I actually introduced a Github workflow to test building stage and  pytest. After that PR we can try to produce a minimal test to see if deleting that line is causing any trouble if you do not mind.

PR #127, opened 2025-10-20
Use PyTorch's built-in GQA support

BahadirAkdemir:
Replace custom repeat_kv implementation with enable_gqa=True parameter in scaled_dot_product_attention for better performance and cleaner code.

Changes:
- Added enable_gqa=True to all F.scaled_dot_product_attention calls
- Removed dependency on manual key/value head replication
- Maintains identical functionality

If it isn’t intentionally done that way, I think it’s better to use PyTorch’s optimized implementation of Multi Query and Grouped Query Attention. It makes the code easier to work with and improves performance through more efficient kernel optimizations. I tested it and observed only around (1e-6 difference with torch.allclose()) differences between the current and updated versions.

### DIFF
--- a/nanochat/gpt.py
+++ b/nanochat/gpt.py
@@ -49,18 +49,6 @@ def apply_rotary_emb(x, cos, sin):
     return out
 
 
-def repeat_kv(x, n_rep):
-    """torch.repeat_interleave(x, dim=1, repeats=n_rep)"""
-    if n_rep == 1:
-        return x
-    bs, n_kv_heads, slen, head_dim = x.shape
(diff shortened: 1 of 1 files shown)


karpathy:
doh, i didn't realize there is mainline support in sdpa! It's actually a bit worse because we don't even use MQA right now. Thank you for the PR, will probably merge.

karpathy:
I chose to do it this way ty!
https://github.com/karpathy/nanochat/commit/a088b7a6ec6e144b99a2a89d0b1f772198abcb97

PR #128, opened 2025-10-20
add optional chunked cross-entropy for memory-efficient training

mnehete32:
This PR introduces support for chunked cross-entropy (chunked_ce / --use_chunked_ce) in the model. When enabled, the cross-entropy loss is computed on-the-fly without materializing full logits in GPU memory. This allows training larger models on GPUs with limited VRAM, at the cost of slightly slower training.

The library added is also light weight, it has very minimal dependencies:
```
python >= 3.9
torch>=2.4
triton>=3.0.0 
```


karpathy:
ty i have a todo to investigate it, but i don't want the repo to become a "configuration monster". I want to incorporate it into mainline when (and if) it actually makes sense. i.e. it gives better results in the report.

mnehete32:
I understand, makes sense 👍
I just wanted to make it available as an optional flag for experimentation and for cases with low gpu memory setups, so large model can be trained.

svlandeg:
As this PR has become somewhat stale & out-of-date, I'll go ahead and close this for now.

PR #129, opened 2025-10-21
Gsm8k platinum + pyo3 forward compat 

Qubitium:
@karpathy Two changes:

1. Add `GSM8k-Platinum` to tests.  Dataset: https://huggingface.co/datasets/madrylab/gsm8k-platinum

blog: https://gradientscience.org/gsm8k-platinum/
paper: https://arxiv.org/abs/2502.03461
code: https://github.com/MadryLab/platinum-benchmarks

`GSM8k-Platinum` is a subset of gsm8k and fully compatible with existing gms8k code.  The researchers removed ambiguous questions that were causing models (and humans) to fail those questions. This reduced false fails in the gsm8k test. 

```py
# Simple test to check we are getting the correct rows from the gsm8k datasets.
# It does not verify the actual content of the dataset itself.
EXPECTED_COUNTS = {
    ("main", "train"): 7473,
    ("main", "test"): 1319,
    ("socratic", "train"): 7473,
    ("socratic", "test"): 1319,
    ("platinum", "test"): 1209,
}
```
You can see from the unit test that `platinum` removed 110 ambiguos rows of data from the original `main/test`.

2.  `Pyo3` depend updated to 0.26.0 with two small changes to `rustbpe` so that it compiles and compiles without warning on python 3.13t and above when nogil is enabled for the build. `nogil` builds of python requires pyo3 >= 0.26.0.

```py
# replace deprecated api with new
pyo3::Python::with_gil => pyo3::Python::attach
py.allow_threads => py.detach
 ```

Thank you for the great tool!

### DIFF
--- a/rustbpe/src/lib.rs
+++ b/rustbpe/src/lib.rs
@@ -307,7 +307,7 @@ impl Tokenizer {
         // Helper: refill `buf` with up to `buffer_size` strings from the Python iterator.
         // Returns Ok(true) if the iterator is exhausted, Ok(false) otherwise.
         let refill = |buf: &mut Vec<String>| -> PyResult<bool> {
-            pyo3::Python::with_gil(|py| {
+            pyo3::Python::attach(|py| {
                 buf.clear();
                 let it = py_iter.bind(py);
                 loop {
@@ -345,7 +345,7 @@ impl Tokenizer {
             total_sequences += buf.len() as u64;
 
             let pattern = self.compiled_pattern.clone();
-            let local: AHashMap<CompactString, i32> = py.allow_threads(|| {
+            let local: AHashMap<CompactString, i32> = py.detach(|| {
                 buf.par_iter()
                     .map(|s| {
                         let mut m: AHashMap<CompactString, i32> = AHashMap::new();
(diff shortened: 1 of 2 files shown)


burtenshaw:
I can run the branch and they seem useful. @Qubitium Don't you think these would be better as two PRs? The `Pyo3` changes are functionally useful, but the `GSM8k-Platinum` is a matter of approach. 

*Just thinking about what will get merged fastest*

Qubitium:
@burtenshaw  Thanks for the review. My dev env was 3.14 so I hit the pyo3 issue running small unit tests so pushed this change along side th GSM8k Platinum change as it was just easier to do on myside to make it run. @karpathy If you don't like the pyo3 changes, I can revert them asap. The meat is the GSM8k Platinum eval.

karpathy:
Not a huge fan of compound PRs. I don't fully understand the need for gsm8k-plat atm.

Qubitium:
@karpathy  Fair. But do you still want the the pyo3 compat changes for Python 3.14 and beyond? I can make that a separate micro pr.

PR #133, opened 2025-10-21
 Multi platform CI - Github Workflow tests v2 [numpy restriction was flexed to support win+py3.13]

SermetPekin:
- Github Workflow with multiplatform and tests with multiple versions of python
- Platforms: Ubuntu, macOS, Windows
- Python versions: 3.10, 3.11, 3.12, 3.13
- utf-8 encoding was specified while reading file on test_rustbpe.py
- in order to support py3.13 on windows numpy restriction was changed
```
# pyproject.toml file
"numpy>=2.3.0; python_version>='3.13'",
"numpy>=1.26.4; python_version<'3.13'",
```

### DIFF
--- a/.github/workflows/base.yml
+++ b/.github/workflows/base.yml
@@ -18,8 +18,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [windows-latest] # ubuntu-latest, macos-latest, 
-        python-version: [ '3.13'] # '3.10', '3.11', '3.12' ,
+        os: [ubuntu-latest, macos-latest, windows-latest]   
+        python-version: ['3.10', '3.11', '3.12', '3.13'] 
       fail-fast: false
 
     steps:


.github/workflows/base.yml
```diff
@@ -0,0 +1,75 @@
...
+        source .venv/bin/activate
+
+    - name: Activate virtual environment (Windows)
+      if: runner.os == 'Windows'
+      run: |
+        .venv\Scripts\activate
+
+    - name: Install dependencies with uv
+      run: |
+        uv pip install transformers>=4.0.0
```
burtenshaw: Why do we need `transformers` here?

.github/workflows/base.yml
```diff
@@ -0,0 +1,75 @@
...
+      uses: actions/checkout@v3
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install uv
+      run: |
+        python -m pip install uv
```
burtenshaw, line 36: I think this container ([`astral-sh/setup-uv@v7`](https://github.com/astral-sh/setup-uv) ) has `uv` installed.

burtenshaw (changes requested):
These workflows would be really useful.

.github/workflows/base.yml
SermetPekin: @burtenshaw ,I created a new version for this PR.  Thanks for the review. with the current codebase it was easier to implement. Btw thanks for the astral artifact suggestion it worked better. [View PR](https://github.com/karpathy/nanochat/pull/151)

SermetPekin:
there is an updated version of this PR. good to be closed. [new PR](https://github.com/karpathy/nanochat/pull/151)

PR #137, opened 2025-10-21
Optimize print0 function: Cache DDP rank eval. for better performance

SermetPekin:
I think the current `print0` function performs redundant work on every call. Similar to template meta programming concept of C++ maybe it is better to let script the evaluation (kind of) at compile time. 

- This PR introduces definition of `print0` function with DDP rank evaluation at module import time. 
```python 
_ddp_rank = int(os.environ.get('RANK', 0))

if _ddp_rank == 0:
    def print0(s="", **kwargs):
        print(s, **kwargs)
else:
    def print0(s="", **kwargs):
        pass
```

### DIFF
--- a/nanochat/common.py
+++ b/nanochat/common.py
@@ -56,10 +56,17 @@ def get_base_dir():
     os.makedirs(nanochat_dir, exist_ok=True)
     return nanochat_dir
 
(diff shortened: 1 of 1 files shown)


karpathy:
so many new lines!

PR #141, opened 2025-10-21
Enabled markdown rendering for response by nanochat in ui.html

Goderr:
I realize that the "https://nanochat.karpathy.ai/" is a temporary hosting site for d23 model but I noticed that the response that was being given by nanochat was in raw markdown format. So I decided to tweak the ui.html so that the response generated by nanochat is rendered and get a better output.

The current UI shows raw Markdown; this PR renders assistant messages as safe GitHub‑flavored Markdown for better readability of tables, code blocks, and emphasis.

Changes:

- Parse with Marked (gfm, breaks).

- Sanitize with DOMPurify.

- Stream throttling to reduce DOM churn and prevent malformed interim tables.

- CSS for tables/code, optional language badges on fenced code. 

1. **Bold words** - The words which were supposed to be bold were displayed in their raw format is instead of **bold** it was displayed as "** bold **" 

Current UI -
image

Updated UI -
image

2. **Table formation** - 

Current UI-
image

Update UI-
image

3. **Code Blocks** -

Current UI-
image

Update UI-
image


karpathy:
This is cool, I'll take a look ty!

willhama:
This is a nice adaption to it. I would just be carefull to not recreate a chatGPT-like interface so it becomes not to complex. In that case a switch to a UI framework such as Svelte, Vue or React makes much more sense or attaching it to something like https://github.com/open-webui/open-webui.  The goal of this chat interface is to keep it neat, simple and easily testable.

Goderr:
> This is cool, I'll take a look ty!

@karpathy thank you for this repo and can't wait for what Eureka Labs has to offer

Goderr:
@willhama I totally agree on keeping the UI simple and less complex. The intent for rendering markdown is too make it readable during testing, there is no new framework or behaviour change in it.

willhama:
> @willhama I totally agree on keeping the UI simple and less complex. The intent for rendering markdown is too make it readable during testing, there is no new framework or behaviour change in it.

@Goderr, I really like the changes you made, so it's totally fine and as you point out add no additional dependencies or frameworks.

PR #142, opened 2025-10-21
Fixes Linux CPU support using "extra" indexes, Mac & Linux pin_memory crash fix

lukestanley:
Linux environments without CUDA like GitHub Codespaces or Codex agents can use won't directly work on master.
- Instead of relying on `sys_platform == 'linux'`, this adds direct control with `--extra cpu` / `--extra gpu` flags.
- Fixed pin_memory=True crash on CPU
- Git ignores eval_bundle

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2,13 +2,32 @@ version = 1
 revision = 3
 requires-python = ">=3.10"
 resolution-markers = [
-    "python_full_version >= '3.12' and sys_platform == 'linux'",
-    "python_full_version >= '3.12' and sys_platform != 'linux'",
-    "python_full_version == '3.11.*' and sys_platform == 'linux'",
-    "python_full_version == '3.11.*' and sys_platform != 'linux'",
-    "python_full_version < '3.11' and sys_platform == 'linux'",
-    "python_full_version < '3.11' and sys_platform != 'linux'",
+    "python_full_version >= '3.12' and sys_platform == 'linux' and extra != 'extra-8-nanochat-cpu' and extra == 'extra-8-nanochat-gpu'",
+    "python_full_version >= '3.12' and sys_platform != 'linux' and extra != 'extra-8-nanochat-cpu' and extra == 'extra-8-nanochat-gpu'",
+    "python_full_version == '3.11.*' and sys_platform == 'linux' and extra != 'extra-8-nanochat-cpu' and extra == 'extra-8-nanochat-gpu'",
+    "python_full_version < '3.11' and sys_platform == 'linux' and extra != 'extra-8-nanochat-cpu' and extra == 'extra-8-nanochat-gpu'",
+    "python_full_version == '3.11.*' and sys_platform != 'linux' and extra != 'extra-8-nanochat-cpu' and extra == 'extra-8-nanochat-gpu'",
+    "python_full_version < '3.11' and sys_platform != 'linux' and extra != 'extra-8-nanochat-cpu' and extra == 'extra-8-nanochat-gpu'",
+    "python_full_version >= '3.12' and sys_platform == 'linux' and extra == 'extra-8-nanochat-cpu' and extra != 'extra-8-nanochat-gpu'",
+    "python_full_version >= '3.12' and sys_platform != 'darwin' and sys_platform != 'linux' and extra == 'extra-8-nanochat-cpu' and extra != 'extra-8-nanochat-gpu'",
+    "python_full_version == '3.11.*' and sys_platform == 'linux' and extra == 'extra-8-nanochat-cpu' and extra != 'extra-8-nanochat-gpu'",
+    "python_full_version < '3.11' and sys_platform == 'linux' and extra == 'extra-8-nanochat-cpu' and extra != 'extra-8-nanochat-gpu'",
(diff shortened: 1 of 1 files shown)


jwnsu:
macos (MPS device) has similar crash issue due to pin_memory flag. I tried pin_memory in data_loader.py and mid_train.py only, as other changes are specific to Linux/CPU.

lukestanley:
This is an explicit, small change to introduce reasonable cross-platform support.

Tested on:
 - CPU with dev/runcpu.sh
 - on my Ubuntu system with a 3090 I validated the new proposed install path by pretraining with a depth of 1 and 1 node.
- [jwnsu](https://github.com/jwnsu) reported success on their MPS

@jwnsu I'm glad that works for you too, it is a simple bug, being a platform specific feature, with a simple fix too, please feel free to use the commit for it [(defd124) ](https://github.com/karpathy/nanochat/pull/142/commits/defd1246aa80c99cf1b486ec112b1223916e21df) for a new PR or whatever you feel is most appropriate.

@karpathy 
Cross-platform support for Torch is a tricky pain and Astral's UV is almost there, but always with caveats.
This is the most simply "pluralistic" upgrade I could figure out!
I am interested in reasonable typical dev platform support, hence the index changes.

Hope this helps.

karpathy:
(sorry for dumb uv questions) do we have to now pass the `--extra` flag to every uv sync?

karpathy:
reading... https://docs.astral.sh/uv/guides/integration/pytorch/#configuring-accelerators-with-optional-dependencies

lukestanley:
It's tricky, it's not a dumb question, Astral uv moves fast!
As I understand it, using "--extra" on the sync seems best for now to allow avoiding UV's use of the package file from making uncontrolled OS based dependency assumptions. uv doesn’t install extras by default.
I don't know how defaults and priorities work, it's possible there are better ways that already work.
Astral uv has "--torch-backend=auto" but it only works  with "uv pip" and not with one lockfiles on multiple systems AFAIK.
@charliermarsh did I get that right?

Thanks a million to both of you for your work!
@karpathy

karpathy:
Got it. I think my concern is that a lot of people are habituated to just doing `uv sync` and they expect it to "do the right thing", so the required use of an extra to pick is a little bit gnarly maybe.

karpathy:
This PR seems to work both on my Linux box and my Macbook and contains other useful fixes. (I also had to bump the uv sync command in run1000.sh btw). Merging for now but interested in further discussion. Thank you for the PR!

lukestanley:
> I think my concern is that a lot of people are habituated to just doing `uv sync` and they expect it to "do the right thing", so the required use of an extra to pick is a little bit gnarly maybe.

Yeah, it would be ideal if `uv sync` _just worked_ for a default platform target at least - be it GPU or CPU.

It seems like uv defaults to CPU from PyPI, probably because of the lockfile; on my Ubuntu PC with a 3090 when I just tried a bare `uv sync` (with no index specified), uv installed `torch==2.9.0` (instead of `torch==2.8.0+cu128` which `uv lock --extra gpu` pulled in for me before).

Setting a default package index for PyTorch is likely possible, I'll investigate, and verify how the defaults work (to verify if this is due to the lockfile, or my prior state).

> This PR seems to work both on my Linux box and my Macbook and contains other useful fixes. 
> (I also had to bump the uv sync command in run1000.sh btw). 

Oops, sorry I missed that one, thanks! ;D

> Merging for now but interested in further discussion. Thank you for the PR!

Awesome, you're welcome! Your unified educational framework / testbed is a great service to knowledge, thank you for educating us! I'm pretty happy with GPUs going _brrr_, but I'm curious how minimal a good "cognitive core" could be, maybe a CPU version might _not_ be a joke one day!
@karpathy

charliermarsh:
Another option is that you can use dependency groups instead of extras, which _do_ support a default. Here's a gist to that effect: https://gist.github.com/charliermarsh/bc3182f969a1064cfb9269c79a49de1a. In short, you use `cpu` and `cu128` groups instead of extras; then `uv sync` will, by default, use the `cu128` PyTorch; and `uv sync --no-default-groups --group cpu` would sync the CPU-only PyTorch.

lukestanley:
Thanks for the gist example, @charliermarsh, uv has saved thousands / millions of devs' time, *and* is in many cases allowing hair regrowth for ex-CUDA hell sufferers! ;D

> Another option is that you can use dependency groups instead of extras, which _do_ support a default. Here's a gist to that effect: https://gist.github.com/charliermarsh/bc3182f969a1064cfb9269c79a49de1a. In short, you use `cpu` and `cu128` groups instead of extras; then `uv sync` will, by default, use the `cu128` PyTorch; and `uv sync --no-default-groups --group cpu` would sync the CPU-only PyTorch.

If I understand correctly, uv dependency groups effectively allow default dependency sets, but maybe in a more explicit way than "that's what the lockfile last had set"? But for our multi-device situation I suspect we only need system markers and extras for special cases, but I'll sleep on it.

To _just work_ with `uv sync` for Mac users and Linux CUDA users, the project file could use environment markers to:
- use CUDA on Linux by default (again!)
- _but,_ markers can't detect CUDA on Linux, so we would treat _non-CUDA_ Linux users as a special case by selecting the CPU **extra** (which maps to the CPU index)
- Bash scripts could detect missing `nvidia-smi` and choose the appropriate **extra** flag for plain Linux CPU support
- This would be good for special cases like MIPS in the future too, and could be a common script that gates `uv sync` cases.
@karpathy

PR #88, opened 2025-10-16
add cpu|mps support

karpathy:
WIP, allowing people to run the code either on CPU (any potato) or MPS (Macbook GPUs)

Atm struggling a bit to figure out how to adjust the `pyproject.toml` to switch pytorch to the basic version on demand. Current workaround is to delete these lines from the pyproject.toml:

```toml

# target torch to cuda 12.8
[tool.uv.sources]
torch = [
    { index = "pytorch-cu128" },
]

[[tool.uv.index]]
name = "pytorch-cu128"
url = "https://download.pytorch.org/whl/cu128"
explicit = true

```

PR #145, opened 2025-10-22
fix: MPS/CPU compatibility for training and inference on Mac

jasonkneen:
## Summary

Makes nanochat fully runnable on Apple Silicon (MPS) and CPU after merging the latest upstream changes. All crashes and silent errors on non-CUDA devices are resolved.

Tested on:
- MacBook Pro M-series, torch 2.2.2 and 2.9.1
- 29/29 tests pass, 10 skipped (FA3 hardware only)

---

### `nanochat/engine.py`
- **Replace `signal.SIGALRM` with `concurrent.futures`** for calculator timeout. `SIGALRM` only works on Unix main thread — silently broken when called from FastAPI worker threads
- Guard `torch.cuda.synchronize()` behind `device_type == "cuda"`

### `nanochat/gpt.py`
- **Extract `init_rotary_embeddings()`** from `init_weights()` so checkpoint loading can restore non-persistent `cos`/`sin` buffers without re-randomising all weights
- Cast rotary buffers to `bfloat16` on CUDA only (MPS bfloat16 requires torch≥2.4; float32 used on MPS/CPU)
- **Add `F.rms_norm` fallback** for torch<2.4 (`rms_norm` was added in 2.4)
- Update `forward()` rotary dtype assertion to be device-aware

### `nanochat/optim.py`
- **`_cuda_compile()`**: skip `torch.compile(fullgraph=True)` on MPS/CPU; return function unchanged for eager execution
- **`adamw_step_fused` / `muon_step_fused`**: move 0-D CPU scalar tensors to parameter device at function start (cross-device ops crash in eager mode on MPS)
- **`muon_step_fused`**: use `bfloat16` in polar express on CUDA only; float32 on MPS/CPU
- **`_step_muon`**: replace `torch._foreach_copy_()` with `p.copy_(s)` loop on non-CUDA (`_foreach_copy_` not implemented on MPS in torch<2.4)

### `nanochat/flash_attention.py`
- **Probe SDPA for `enable_gqa` support at import time** (`enable_gqa` added in torch 2.5; `inspect.signature` raises on C builtins in older Python/torch)
- Fall back to manual KV head repetition via `repeat_interleave` when `enable_gqa` is unavailable

### `nanochat/checkpoint_manager.py`
- Call `model.init_rotary_embeddings()` instead of `model.init_weights()` after `load_state_dict()` to restore non-persistent rotary buffers without overwriting loaded weights

### `scripts/base_train.py`
- Guard `torch.compile(model)` behind `device_type == "cuda"`
- Set `mfu = None` on non-CUDA instead of computing `0/inf = 0.00%`
- Handle `mfu is None` cleanly in end-of-run report

### `tests/test_mps_compat.py` *(new)*
- 16 tests covering every fix; all pass on MPS with torch 2.2.2 and 2.9.1

### DIFF
--- a/nanochat/gpt.py
+++ b/nanochat/gpt.py
@@ -40,8 +40,12 @@ class GPTConfig:
 
 
 def norm(x):
-    # Purely functional rmsnorm with no learnable params
-    return F.rms_norm(x, (x.size(-1),))
+    # Purely functional rmsnorm with no learnable params.
+    # F.rms_norm was added in PyTorch 2.4; fall back to manual implementation on older builds.
+    if hasattr(F, 'rms_norm'):
+        return F.rms_norm(x, (x.size(-1),))
+    variance = x.pow(2).mean(-1, keepdim=True)
+    return x * torch.rsqrt(variance + 1e-6)
 
 
 def has_ve(layer_idx, n_layer):
@@ -230,15 +234,22 @@ class GPT(nn.Module):
                 torch.nn.init.zeros_(block.attn.ve_gate.weight)
 
         # Rotary embeddings
-        head_dim = self.config.n_embd // self.config.n_head
-        cos, sin = self._precompute_rotary_embeddings(self.rotary_seq_len, head_dim)
-        self.cos, self.sin = cos, sin
+        self.init_rotary_embeddings()
 
         # Cast embeddings to bf16: optimizer can tolerate it and it saves memory
         if self.transformer.wte.weight.device.type == "cuda":
             self.transformer.wte.to(dtype=torch.bfloat16)
             for ve in self.value_embeds.values():
                 ve.to(dtype=torch.bfloat16)
+
+    def init_rotary_embeddings(self):
+        """Initialize (or re-initialize) only the non-persistent rotary embedding buffers.
+        Call this instead of the full init_weights() when loading a checkpoint, since these
+        buffers have persistent=False and are therefore absent from the saved state_dict.
+        """
+        head_dim = self.config.n_embd // self.config.n_head
+        cos, sin = self._precompute_rotary_embeddings(self.rotary_seq_len, head_dim)
+        self.cos, self.sin = cos, sin
 
     def _precompute_rotary_embeddings(self, seq_len, head_dim, base=10000, device=None):
         # TODO: bump base theta more? e.g. 100K is more common more recently
@@ -253,7 +264,9 @@ class GPT(nn.Module):
         # calculate the rotation frequencies at each (time, channel) pair
         freqs = torch.outer(t, inv_freq)
         cos, sin = freqs.cos(), freqs.sin()
-        cos, sin = cos.bfloat16(), sin.bfloat16() # keep them in bfloat16
+        # bfloat16 saves memory on CUDA; MPS only supports it in torch>=2.4 so use float32 there
+        if device.type == "cuda":
+            cos, sin = cos.bfloat16(), sin.bfloat16()
         cos, sin = cos[None, :, None, :], sin[None, :, None, :] # add batch and head dims for later broadcasting
         return cos, sin
 
@@ -391,7 +404,9 @@ class GPT(nn.Module):
         # Grab the rotary embeddings for the current sequence length (they are of shape (1, seq_len, 1, head_dim/2))
         assert T <= self.cos.size(1), f"Sequence length grew beyond the rotary embeddings cache: {T} > {self.cos.size(1)}"
         assert idx.device == self.cos.device, f"Rotary embeddings and idx are on different devices: {idx.device} != {self.cos.device}"
-        assert self.cos.dtype == torch.bfloat16, "Rotary embeddings must be in bfloat16"
+        expected_rot_dtype = torch.bfloat16 if self.cos.device.type == "cuda" else torch.float32
+        assert self.cos.dtype == expected_rot_dtype, \
+            f"Rotary embeddings dtype mismatch: expected {expected_rot_dtype}, got {self.cos.dtype}"
         # if kv cache exists, we need to offset the rotary embeddings to the current position in the cache
         T0 = 0 if kv_cache is None else kv_cache.get_pos()
         cos_sin = self.cos[:, T0:T0+T], self.sin[:, T0:T0+T] # truncate cache to current sequence length
--- a/nanochat/checkpoint_manager.py
+++ b/nanochat/checkpoint_manager.py
@@ -100,8 +100,10 @@ def build_model(checkpoint_dir, step, device, phase):
     with torch.device("meta"):
         model = GPT(model_config)
     # Load the model state
+    # Only init the rotary embedding buffers (persistent=False, absent from checkpoint),
+    # not the full weights — everything else comes from load_state_dict.
     model.to_empty(device=device)
-    model.init_weights() # note: this is dumb, but we need to init the rotary embeddings. TODO: fix model re-init
+    model.init_rotary_embeddings()
     model.load_state_dict(model_data, strict=True, assign=True)
     # Put the model in the right training phase / mode
     if phase == "eval":
--- a/nanochat/engine.py
+++ b/nanochat/engine.py
@@ -13,9 +13,8 @@ The whole thing is made as efficient as possible.
 
 import torch
 import torch.nn.functional as F
-import signal
 import warnings
-from contextlib import contextmanager
+import concurrent.futures
 from collections import deque
 from nanochat.common import compute_init, autodetect_device_type
 from nanochat.checkpoint_manager import load_model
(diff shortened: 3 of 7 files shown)


jwnsu:
Very nice changes for macos, thanks! Suggest we have an additional MEMORY_SIZE parameter to calculate device_batch_size and total_batch_size, eval_tokens, split_tokens.

Many macos devices have 64GB or 128GB unified memory. For example, 64GB memory can support device_batch_size=16.

jasonkneen:
Good point! will update

jwnsu:
Thanks, will test your revised PR and hope it will help accelerate it merging into repository -- currently running my own set of modifications, yours is more comprehensive.

jasonkneen:
Updated! was using x86 python so updated that and currently running the training to 1000 steps

dev/README_MACOS.md
burtenshaw, line 1: I think multiple README's will be confusing to students. Why not open another PR with a separate guide on MPS use. Some guides are also shared [here](https://github.com/karpathy/nanochat/discussions) and [here](https://huggingface.co/nanochat-students).

dev/runcpu.sh
burtenshaw, line 1: This is useful for dealing with all the MPS options. @jasonkneen Can you share running logs?

pyproject.toml
```diff
@@ -14,7 +14,7 @@ dependencies = [
     "setuptools>=80.9.0",
     "tiktoken>=0.11.0",
     "tokenizers>=0.22.0",
-    "torch>=2.8.0",
+    "torch>=2.0.0,!=2.3.*,!=2.4.*,!=2.5.*,!=2.6.*,!=2.7.*,!=2.8.*,!=2.9.*",
```
burtenshaw: `mps` should run on the cpu variant or `torch`. I think the use of older torch is caused by the lack of index `https://download.pytorch.org/whl/cpu`.

nanochat/gpt.py
```diff
@@ -35,7 +35,13 @@ class GPTConfig:
 
 def norm(x):
     # Purely functional rmsnorm with no learnable params
-    return F.rms_norm(x, (x.size(-1),))
+    # Fallback for older PyTorch versions that don't have F.rms_norm
+    if hasattr(F, 'rms_norm'):
+        return F.rms_norm(x, (x.size(-1),))
+    else:
```
burtenshaw: Could you try this out with the latest change son `main`? Mainly the `toml`. We should be able to use a `torch` version with `F.rms_norm`.

burtenshaw (changes requested):
I took a look at this and added some inline comments. In general:
- `dev/README_MACOS.md` is a guide on nanochat for mps and I think there are better ways for the community to share guides/ tutorials than this repo.
- The mps python environment depends on older version of torch which make meaningful changes to the ml code. 
- The are a number of shell scripts that I think will confuse students. @karpathy Maybe we could deal with this in another PR by making a `recipes` directory that contains alternatives to `speedrun.sh` for alternative hardware. That will cut down on config, and maintain the 'just run it' vibe.

jwnsu:
I've started a testing run, will share logs once it's done. It hang during Evaluating base model after lambda_openai, seems struggling with memory (on a macos 64GB).  Adding flag **--max-per-task=32** to base_eval resolved the eval issue, also  **--max-problems 100** for chat_eval.

Below is the report card for runmacos_overnight.sh:
## Summary

- Characters: 396,325
- Lines: 9,826
- Files: 49
- Tokens (approx): 99,081
- Dependencies (uv.lock lines): 2,365

| Metric          | BASE     | MID      | SFT      | RL       |
|-----------------|----------|----------|----------|----------|
| CORE            | 0.0175   | -        | -        | -        |
| ARC-Challenge   | -        | -        | 0.2700   | -        |
| ARC-Easy        | -        | -        | 0.2200   | -        |
| GSM8K           | -        | -        | 0.0000   | -        |
| HumanEval       | -        | -        | 0.0000   | -        |
| MMLU            | -        | -        | 0.3200   | -        |
| ChatCORE        | -        | -        | 0.0160   | -        |

pavelklymenko:
👍🏻

jasonkneen:
Will this get merged in?

dev/continue_training.sh
```diff
@@ -0,0 +1,230 @@
...
+fi
+
+if [ "$SFT_CHECKPOINT" != "none" ]; then
+    echo "  ✓ SFT: $SFT_CHECKPOINT"
+else
+    echo "  ✗ SFT: Not found"
+fi
+echo ""
+
+# Determine what to do
```
karpathy, line 120: This script is being very clever and it's bloating the file a lot. It seems like it's designed to be run over and over and reliably push the whole pipeline through. I was thinking of these scripts as more partly documentation, and partly an example of a thing you can run and wait a long time to get result. All of this logic is complexifying it quite a bit and makes it harder to read and understand what the sequence of steps is. I'm ok with some of the "autodetect hyperparams" up top. But I don't think all of the other logic here is a good idea unless you can justify.

nanochat/gpt.py
```diff
@@ -35,7 +35,13 @@ class GPTConfig:
 
 def norm(x):
     # Purely functional rmsnorm with no learnable params
-    return F.rms_norm(x, (x.size(-1),))
+    # Fallback for older PyTorch versions that don't have F.rms_norm
+    if hasattr(F, 'rms_norm'):
```
karpathy, line 45: Surprised this wouldn't exist. Are we trying to support way too old version of pytorch or so

nanochat/muon.py
```diff
@@ -19,7 +18,11 @@ def zeropower_via_newtonschulz5(G: Tensor, steps: int) -> Tensor:
     """
     assert G.ndim >= 2 # batched Muon implementation by @scottjmaddox, and put into practice in the record by @YouJiacheng
     a, b, c = (3.4445, -4.7750,  2.0315)
-    X = G.bfloat16()
+    # Use bfloat16 on CUDA for speed, float32 on MPS/CPU (MPS doesn't support bfloat16)
+    if G.device.type == "cuda":
+        X = G.bfloat16()
+    else:
```
karpathy: possibly able to skip the else block here.

pyproject.toml
```diff
@@ -14,7 +14,7 @@ dependencies = [
     "setuptools>=80.9.0",
     "tiktoken>=0.11.0",
     "tokenizers>=0.22.0",
-    "torch>=2.8.0",
+    "torch>=2.0.0,!=2.3.*,!=2.4.*,!=2.5.*,!=2.6.*,!=2.7.*,!=2.8.*,!=2.9.*",  # Excludes versions without macOS arm64 wheels
```
karpathy: Why not leave as before?

pyproject.toml
```diff
@@ -44,19 +44,6 @@ python_files = ["test_*.py"]
 python_classes = ["Test*"]
 python_functions = ["test_*"]
 
-# target torch to cuda 12.8
```
karpathy: Wait isn't all of this needed to correctly route the indices

pyproject.toml
jubruckne: I'm running torch 2.9 so wheels seem to have been added.

README.md
```diff
@@ -95,7 +95,38 @@ And a bit more about computing environments that will run nanochat:
...
+- 128GB: batch_size=16, total_batch=16384 (16× faster)
+- 64GB: batch_size=8, total_batch=8192 (8× faster)
+- 32GB: batch_size=4, total_batch=4096 (4× faster)
+- <32GB: batch_size=1, total_batch=1024 (baseline)
+
+**Quick start on Mac:**
+```bash
+bash dev/runcpu.sh              # Quick test (~15-30 min on 128GB Mac)
+bash dev/runmac_overnight.sh    # Better quality (~2-3 hours on 128GB Mac)
+bash dev/continue_training.sh   # Resume interrupted training
```
karpathy: I don't love the continue training script, per my comment also above.

karpathy:
@jasonkneen  Thank you I want to merge most of this. I am most suspicious of all the uv / dependency changes. I'd prefer we not touch the dependencies unless we have to. And I think I don't wish to support continue training, it feels too bloating.

karpathy:
One more question on my mind - how come all of this worked on my macbook fine without all of the changes?

jasonkneen:
It did? That's awkward 🧐 I can check on the uv changes and see if it's necessary

svlandeg:
Hmmm wait. So this PR got closed when https://github.com/karpathy/nanochat/pull/88 was merged (Oct 21) and I subsequently removed the feature branch `cpu-mps-dev` I guess to clean up. That automatically closed this PR on Oct 30 because it targets that (merged) feature branch. But there's plenty of review discussion still going on here from within that time period Oct 21-30 so maybe this was closed prematurely by accident, should target `master` instead and should get another look at. 🤔

svlandeg:
@jasonkneen : are you still around to have a look at the reviewer comments and to update this PR for the latest on `master`?

jasonkneen:
sure I can take a look

pyproject.toml
icetraxx: All of those versions do contain arm64 wheels:

https://pypi.org/project/torch/2.3.0/#files
https://pypi.org/project/torch/2.4.0/#files
https://pypi.org/project/torch/2.5.0/#files
https://pypi.org/project/torch/2.6.0/#files
https://pypi.org/project/torch/2.7.0/#files
https://pypi.org/project/torch/2.8.0/#files
https://pypi.org/project/torch/2.9.1/#files

jasonkneen:
Apologies - I am looking at this now

svlandeg:
@jasonkneen: there are plenty of open reviewer's comments, could you look into those as well?

jasonkneen:
will do

PR #151, opened 2025-10-22
Multi platform CI - Github Workflow tests v3

SermetPekin:
>This PR introduces a comprehensive CI/CD pipeline without requiring any changes to the existing poproject.toml file. All tests are passing successfully on my branch. [View Test Results](https://github.com/SermetPekin/nanochat/actions/runs/18713928841)

- Github Workflow with multiplatform and tests with multiple versions of python [for now only 3.10]
- Platforms: Ubuntu, macOS
- Python versions: 3.10

## Details
- Uses astral-sh/setup-uv@v7 for fast and reliable dependency management
- Comprehensive test coverage with pytest --maxfail=5
- Platform-specific environment variable handling
- No breaking changes to existing codebase

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,7 +32,6 @@ jobs:
     - name: Install dependencies with uv
       run: |
         uv sync --extra cpu 
-        uv pip install transformers>=4.0.0
 
     - name: Add nanochat to PYTHONPATH (Unix)
       run: |


.github/workflows/base.yml
```diff
@@ -0,0 +1,50 @@
...
+    branches:
+      - master
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]   
+        python-version: ['3.10', '3.11', '3.12', '3.13']
```
karpathy: we dont have to support so many. there is a .python-version in the repo.

.github/workflows/base.yml
```diff
@@ -0,0 +1,50 @@
...
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Set up uv
+      uses: astral-sh/setup-uv@v7
+
+    - name: Install dependencies with uv
+      run: |
+        uv sync
```
karpathy: this changed as of PR this morning, i think now would be --extra cpu i believe

karpathy:
People love CI and tests.

SermetPekin:
@karpathy you would also love it if you happen to break any code forgetting a line :) 👑 King Karpathy.

.github/workflows/base.yml
SermetPekin: sure. I am going to limit this workflow file to 3.10

.github/workflows/base.yml
SermetPekin: I will check this out.

SermetPekin:
@karpathy this PR is ready to review. CI tests are passing. [View Test Results](https://github.com/SermetPekin/nanochat/actions/runs/18725984296)

.github/workflows/base.yml
```diff
@@ -0,0 +1,49 @@
...
+    - name: Set up uv
+      uses: astral-sh/setup-uv@v7
+
+    - name: Install dependencies with uv
+      run: |
+        uv sync --extra cpu 
+        uv pip install transformers>=4.0.0
+
+    - name: Add nanochat to PYTHONPATH (Unix)
+      if: runner.os != 'Windows'
```
svlandeg: can be removed, as `windows` isn't in the `os` matrix (anymore)
```suggestion
```

.github/workflows/base.yml
```diff
@@ -0,0 +1,49 @@
...
+      run: |
+        echo "PYTHONPATH=$(pwd):$PYTHONPATH" >> $GITHUB_ENV
+
+    - name: Add nanochat to PYTHONPATH (Windows)
+      if: runner.os == 'Windows'
+      run: |
+        echo "PYTHONPATH=$PWD;$env:PYTHONPATH" >> $env:GITHUB_ENV
+
```
svlandeg: can be removed, as `windows` isn't in the `os` matrix (anymore)
```suggestion
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,49 @@
+name: Test
```
svlandeg, line 1: I would personally put this in a file `test.yml`, for clarity.

.github/workflows/base.yml
```diff
@@ -0,0 +1,49 @@
+name: Test
+
+on:
+  push:
+    branches:
+      - '**'
```
svlandeg: I don't think you want to run this for every branch? Typically, you'd do `master` or `main` only, and then any PR that gets opened (which you have in the next step). 
```suggestion
      - master
```

.github/workflows/base.yml
```diff
@@ -0,0 +1,49 @@
...
+        os: [ubuntu-latest, macos-latest]   
+        python-version: ['3.10'] 
+      fail-fast: false
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
```
svlandeg:
```suggestion
      uses: actions/setup-python@v6
```

.github/workflows/base.yml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]   
+        python-version: ['3.10'] 
+      fail-fast: false
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
```
svlandeg:
```suggestion
      uses: actions/checkout@v5
```

.github/workflows/base.yml
SermetPekin: @svlandeg your first question is an answer to this. "What else is missing here that the tests aren't actually run yet on this PR? (maybe GH actions are currently disabled for this repo?)" I want to see it on my branch before pushing here. This is just to trigger easily. It can be changed while pushing by reviewers.

svlandeg: Ah gotcha, so you basically just set this so that your branch would test, but this should be reverted to `master` before merging this PR. Makes sense.

.github/workflows/test.yml
```diff
@@ -0,0 +1,43 @@
...
+        uv sync --extra cpu 
+        uv pip install transformers>=4.0.0
+
+    - name: Add nanochat to PYTHONPATH (Unix)
+      run: |
+        echo "PYTHONPATH=$(pwd):$PYTHONPATH" >> $GITHUB_ENV
+
+    - name: Run pytest
+      run: |
+        uv run pytest tests/ --maxfail=5
```
svlandeg:
```suggestion
        uv run pytest tests/
```
5 times seems an awful lot of times to re-run a failing test suite. I would remove this option and only consider it if we often run into flaky behaviour (connection issues etc)

SermetPekin: --max-fail does not re-run a failing test suite. if fewer than 3 different tests failures it will continue to show all these tests to help developers to see which ones are failing in a complete run.  Suppose we have 10 different tests for different functions sometimes we may want to proceed even if we had a failure in some other unrelated functions to examine which parts are affected. Helps us to see the big picture. (We may also use -x parameter to stop running other tests after any one failure. But regardless of what we choose there is no rerun. There is only one pass of all tests.)

svlandeg: Ah, sorry, should have looked it up 🥲  Agree with you that it's nice to see the full picture - I'd usually just let the whole test suite finish so you can get a sense of how many tests and which ones are failing. So I'd personally still drop the option: just running every test once seems fine by me.

.github/workflows/test.yml
```diff
@@ -0,0 +1,43 @@
...
+    branches:
+      - master
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]   
+        python-version: ['3.10']
```
gsaponaro, line 17: It would be nice to get the supported version number from the file `.python-version` in the repo, rather than hard-coding and duplicating that information here.
Something like this, after `actions/checkout`:

```
      - name: Read Python version
        id: pyver
        run: echo "python-version=$(cat .python-version)" >> $GITHUB_OUTPUT

      - name: Set up Python
        uses: actions/setup-python@v5
        with:
          python-version: ${{ steps.pyver.outputs.python-version }}
```

SermetPekin: thanks. I think it is better as it is for a few reasons. We can test multiple versions when necessary, it is better to be hard coded than reading a (random) file from the repository. This file can be ignored in `.gitignore` file at some point which would result failing to test the tests.

gsaponaro: OK, I guess it's fine to have some information duplicated in different parts of the repository :)

The first reason you mention, to be able to test multiple versions (one day) in the `matrix` definition is compelling. Apart from that, we could set a default fallback value when `$(cat .python-version)` fails... but it would be equivalent to hard-coding `python-version: ['3.10']`.

svlandeg (commented):
Hmm, I'm guessing that after the recent refactors and the addition of Flash Attention 3, these tests won't run anymore on a CPU node.

karpathy:
Looks like a bunch of people will have trouble with FA3. I might try to implement a fallback.

.github/workflows/test.yml
```diff
@@ -0,0 +1,43 @@
...
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Set up uv
+      uses: astral-sh/setup-uv@v7
+
+    - name: Install dependencies with uv
+      run: |
+        uv sync --extra cpu 
+        uv pip install transformers>=4.0.0
```
svlandeg: `transformers` has recently been added to the repo's deps

```suggestion
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,42 @@
...
+    - name: Set up uv
+      uses: astral-sh/setup-uv@v7
+
+    - name: Install dependencies with uv
+      run: |
+        uv sync --extra cpu 
+
+    - name: Add nanochat to PYTHONPATH (Unix)
+      run: |
+        echo "PYTHONPATH=$(pwd):$PYTHONPATH" >> $GITHUB_ENV
```
svlandeg, line 38: @karpathy: I personally think it would be nicer to be able to run `uv pip install -e .` here, instead of messing with env vars. But if we want that then we'd have to add

```
[tool.setuptools]
packages = ["nanochat"]
```

to `pyproject.toml`.

svlandeg:
Superseded by https://github.com/karpathy/nanochat/pull/536

PR #154, opened 2025-10-22
FEAT: Allow CPU-only execution in compute_init

SyedaAnshrahGillani:
Modified compute_init in nanochat/common.py to allow for CPU-only execution by checking the NANOCHAT_DEVICE environment variable. This enhances flexibility and hackability, aligning with the project's goals of accessibility.


karpathy:
use --device_type

PR #156, opened 2025-10-22
Export the base dir variable in runcpu.sh

tlepoint:
Similar to #48 and #113  but for dev/runcpu.sh

### DIFF
--- a/dev/runcpu.sh
+++ b/dev/runcpu.sh
@@ -10,7 +10,7 @@
 
 # all the setup stuff
 export OMP_NUM_THREADS=1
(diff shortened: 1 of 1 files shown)


willhama:
@tlepoint could explain quickly why it does not work without exporting the env? Thanks. :)

tlepoint:
> @tlepoint could explain quickly why it does not work without exporting the env? Thanks. :)

In bash, when you set a variable without export, it only exists in the current shell process. Child processes (scripts called from this script) cannot see these variables.

willhama (approved):

svlandeg (approved):
Thanks @tlepoint! This seems to be the last `export` that was forgotten 👍

PR #159, opened 2025-10-23
 Improve configurator: add testable parse_args() and ConfigManager class

SermetPekin:
Refactored configurator for better architecture and control over global variable injection.
- Introduced ConfigManager class for explicit configuration handling
- Added parse_args() function that can be tested independently


svlandeg:
The `configurator` has been removed in favor of using `argparse`, so I'm afraid this one has become obsolete. Closing, but thanks!

PR #162, opened 2025-10-24
Add Adaptive Gradient Clipping (AGC) to pretraining

ulanch:
This PR adds Adaptive Gradient Clipping (AGC) as an optional training technique for pretraining, addressing the TODO comment in base_train.py about experimenting with gradient clipping methods.

**What is AGC?**

Standard gradient norm clipping clips all gradients globally. AGC instead clips gradients based on the ratio of gradient norm to parameter norm, adapting the threshold per-parameter based on parameter scale. This can help stabilize training in deep networks where different layers have different parameter scales.

Implementation follows [Brock et al., 2021](https://arxiv.org/abs/2102.06171) with simplifications: uses per-tensor norms instead of unit-wise (per-row/channel) norms, and applies to all parameters including biases/LayerNorm. This keeps the implementation minimal and hackable while preserving the core benefit of scale-adaptive clipping.

**Changes**

- Add `agc_clip` hyperparameter to config section (disabled by default with 0.0)
- Implement `clip_grad_agc()` function inline in base_train.py
- Integrate AGC into training loop after standard gradient clipping

**Usage**

Set `agc_clip=0.01` (or similar threshold) when running base_train.py to enable. Disabled by default to preserve existing behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,8 @@ nanochat cn be run on CPU or on MPS (if you're on Macbook), and will automatical
 
 To customize your nanochat, see [Guide: infusing identity to your nanochat](https://github.com/karpathy/nanochat/discussions/139) in Discussions, which describes how you can tune your nanochat's personality through synthetic data generation and mixing that data into midtraining and SFT stages.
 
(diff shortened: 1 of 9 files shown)


karpathy:
I'm not going to start adding various random features to make nanochat into a "configuration monster". If AGC is indeed better, please attach a report demonstrating better CORE score for the same wall clock time.

PR #172, opened 2025-10-25
FIX: uv workspace

Majdoddin:
### Problem
The original installation does not work properly:
- `uv pip show nanochat` worked but importing from a script outside the project directory failed.
- `rustbpe` got installed as a module not a package.
```bash
  $ uv sync --extra cpu
  $ . .venv/bin/activate
  $ python -c "import rustbpe"
  $ uv pip show rustbpe   # installed as module, not package
  warning: Package(s) not found for: rustbpe
  $ uv pip show nanochat
  Name: nanochat
  Version: 0.1.0
  Location: /home/ruhollah/ai/Boolformer/nanochat/.venv/lib/python3.10/site-packages
  Editable project location: /home/ruhollah/ai/Boolformer/nanochat
  Requires: datasets, fastapi, files-to-prompt, numpy, psutil, regex, setuptools, tiktoken, tokenizers, torch, uvicorn, wandb
  Required-by:
  $ python -c "import nanochat"    # Works inside nanochat directory
  $ cd ~
  $ python -c "import nanochat"    # Fails outside nanochat directory
  Traceback (most recent call last):
    File "<string>", line 1, in <module>
  ModuleNotFoundError: No module named 'nanochat'
```
It seems it is using maturin for a [Mixed Rust/Python package](https://www.maturin.rs/project_layout.html#mixed-rustpython-project), but this project has in fact 2 pacaages: `nanochat` and `rustbpe`. 
Similar problem is reported [here](https://github.com/PyO3/maturin/issues/490) (the `.pth` files are not generated for `nanochat`)

### Fix 
- Creating a uv workspace with:
-- `nanochat`: Python package (uv_build backend, depends on rustbpe)
-- `rustbpe`: Rust extension package (maturin backend)
- moving `nanochat` to `src/`, as[ required by uv workspace](https://docs.astral.sh/uv/concepts/projects/workspaces/#workspace-layouts).
- updating `speedrun.sh`
```bash
  $ uv sync --extra cpu
  $ . .venv/bin/activate
  $ python -c "import rustbpe"
  $ uv pip show rustbpe      # installed as a package, not a module
  Name: rustbpe
  Version: 0.1.0
  Location: /home/ruhollah/ai/Boolformer/nanochat/.venv/lib/python3.10/site-packages
  Editable project location: /home/ruhollah/ai/Boolformer/nanochat/rustbpe
  Requires:
  Required-by: nanochat
  $ cd ~
  $ python -c "import nanochat"    # Now works from any directory!
  $ uv pip show nanochat
  Name: nanochat
  Version: 0.1.0
  Location: /home/ruhollah/ai/Boolformer/nanochat/.venv/lib/python3.10/site-packages
  Editable project location: /home/ruhollah/ai/Boolformer/nanochat
  Requires: datasets, fastapi, files-to-prompt, numpy, psutil, regex, rustbpe, setuptools, tiktoken, tokenizers, torch, uvicorn, wandb
  Required-by:
```

The other alternative would be [setting path](https://docs.astral.sh/uv/concepts/projects/dependencies/#path)  for `rustbpe`:
```
[tool.uv.sources]
rustbpe = { path = "./ruspbpe/" }
```
_maybe_ with no need to move `nanochat` to `src/`

PS - updated versions for `maturin`, `pyo3`, `pyo3-log`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ dependencies = [
     "torch>=2.8.0",
     "uvicorn>=0.36.0",
     "wandb>=0.21.3",
-    "rustbpe==0.1.0"
+    "rustbpe~=0.1.0"
 ]
 
 [build-system]


karpathy:
Hi @charliermarsh (or @zanieb), are you by any chance able to briefly advise on the best uv setup for nanochat? I believe the repo is a bit unique in that 1) we take dependency on pytorch and wish it to work on multiple platforms and 2) the custom `rustbpe` rust tokenizer is a bit of a "project within a project". I'm not able to confidently review this PR or variations. The current pyproject toml is here https://github.com/karpathy/nanochat/blob/master/pyproject.toml

rustbpe/Cargo.toml
```diff
@@ -8,8 +8,8 @@ dary_heap = "0.3"
 indexmap = "2.2"
 fancy-regex = "0.16.1"
 log = "0.4.28"
-pyo3 = { version = "0.23.3", features = ["extension-module"] }
-pyo3-log = "0.12.4"
+pyo3 = { version = "0.26", features = ["extension-module"] }
+pyo3-log = "0.13.1"
```
zanieb, line 12: Why do these dependency versions change? That doesn't seem like it belongs here unless it's needed.

zanieb: Is this for the `pyo3/extension-module` feature?

rustbpe/pyproject.toml
```diff
@@ -0,0 +1,17 @@
...
+# when uv/pip invoke maturin through the PEP 517 interface, it always builds in release mode by default.
+[build-system]
+requires = ["maturin>=1.9.6,<2.0"]
+build-backend = "maturin"
+
+[tool.maturin]
+module-name = "rustbpe"
+bindings = "pyo3"
+manifest-path = "Cargo.toml"
+features = ["pyo3/extension-module"]
```
zanieb, line 17: Why did you need to add this feature?

zanieb:
@karpathy do you ever intend to publish the package? i.e., so it's installable from PyPI rather than from a local checkout?

rustbpe/Cargo.toml
Majdoddin: Thank you very much for your review. 
Yes, the dependencies are needed for the feature.

And I just set the versions to match the `maturin`'s version in `rustbpe/pyproject.toml` (actually `pyo3` should be `0.27`, but `pyo3-log` still does not support it)
```
requires = ["maturin>=1.9.6,<2.0"]
```
Is it OKs so?

rustbpe/pyproject.toml
Majdoddin: This feature is needed if the `rustbpe`, as a member of the worksapce, is to be installed in editable mode (which is the default mode). 
Or we should specify `rustbpe` in the top `pyproject.toml` to be installed in  noneditable mode. 

On the other hand, this feature is needed, 
-- as you mentioned, if the package is to be published on PyPI, 
-- and for a docker image of the project, if any:
For a smaller image size, the `Dockerfile` often uses a separate build stage, possibly with a different `FROM` and Python version. Without the `extension-module` feature, the module is dynamically linked to the build stage's `libpython`, which won't exist in the final image, causing a runtime error. With the feature, the module is not linked to `libpython`, and its symbols are resolved at runtime by the host Python process that loads it.

rustbpe/pyproject.toml
```diff
@@ -0,0 +1,17 @@
+[project]
+name = "rustbpe"
+version = "0.1.0"
+description = "A very lightweight Rust library for training a GPT tokenizer"
+readme = "README.md"
+requires-python = ">=3.10"
+
+# when uv/pip invoke maturin through the PEP 517 interface, it always builds in release mode by default.
```
konstin, line 8: This is unfortunately a gap in PEP 517 that this isn't configurable (https://discuss.python.org/t/pep-517-debug-vs-release-builds/1924), though rustbpe still compiles fast even in release mode.

rustbpe/pyproject.toml
```diff
@@ -0,0 +1,17 @@
...
+[build-system]
+requires = ["maturin>=1.9.6,<2.0"]
+build-backend = "maturin"
+
```
konstin, line 12: You can add rebuilds through `uv sync` with the following:

```toml
[tool.uv]
cache-keys = [
    { file = "pyproject.toml" },
    { file = "src/**/*.rs" },
    { file = "Cargo.toml" },
    { file = "Cargo.lock" }
]
```

This makes uv check whether any rust source file has changed and rerun the maturin build on `uv sync` or `uv run`.

pyproject.toml
```diff
@@ -46,33 +40,39 @@ python_functions = ["test_*"]
...
+
+[tool.uv]
+conflicts = [
+    [
+        { extra = "cpu" },
+        { extra = "gpu" },
+    ],
+]
+# rebuild if rustbpe is modified
+cache-keys = [{ file = "pyproject.toml" }, {file = "rustbpe/**"}]
```
konstin: This rebuilds the nanochat package (this `pyproject.toml`) on change, while we want to rebuild rustbpe.

pyproject.toml
```diff
@@ -17,21 +17,15 @@ dependencies = [
     "torch>=2.8.0",
     "uvicorn>=0.36.0",
     "wandb>=0.21.3",
+    "rustbpe==0.1.0"
```
konstin: If you plan on publishing both packages to PyPI, this bound may need to be adjusted.

konstin (commented):
Hi @karpathy, I'm working with Charlie and Zanie 👋 

The new workspace structure looks ideal for what you describe. rustbpe being its own top level import and not a submodule of nanochat, it fits well in its own package in a workspace. With this setup, we don't need to call `maturin develop` manually, `uv sync` can do this internally. By also using `tool.uv.sources`, `uv sync`/`uv run` will rebuild when any Rust sources change.

I can't comment on this inline, but we need to change `run1000.sh` and `runcpu.sh` to install Rust before calling `uv sync`, and we can drop the `maturin develop`.

The `uv.lock` version changes look unrelated, do we need the version bumps in this PR? Related to this, it seems that on master, both torch 2.8.0 and 2.9.0 are used, depending whether GPU or CPU are used. The lockfile update in this PR fixes this, but it's probably better to separate those changes out.

Majdoddin:
@konstin Thank you very much for your review.

I updated `runcpu.sh` and `run1000.sh`.

About the `uv.lock`, AFIK, the common practice is that if a PR changes the dependency metadata (like `pyproject.toml`, `Cargo.toml`), it should commit the lock files. 
Or am I missing sth here?

rustbpe/pyproject.toml
Majdoddin: Yes, exactly. I had seen that discussion, I just left a comment here to emphasis that it does the same as the manual `maturin develop --release`

rustbpe/pyproject.toml
Majdoddin: done, thanks!

pyproject.toml
Majdoddin: done, thanks!

pyproject.toml
Majdoddin: done, thanks!

konstin:
The lockfile looks as you ran with `--upgrade` or in some other way caused an update, which changes many dependency versions. We usually try to separate dependency updates from other changes from dependency updates are independent of features and bugfixes, and to separate potential regressions from updates into their own PRs.

karpathy:
Thank you all I'll try to find time today to look at this PR. I have to sit down and read more uv docs. It's definitely a bit anxiety inducing given that it's a large PR and it moves a lot of files, which always causes chaos. By default, I don't want to do this unless there is a great reason to.

I'd like to better understand what the concrete issue is in the current master and what the most surgical changes are to improve it. This PR looks possibly a lot more bloated than it has to be.

- I don't think I intend to support pip install of nanochat, I'm not sure why someone would want to do that or if that's a goal of the repo. Doing so suggests more of a library or so, which nanochat is not.
- I do like that the maturin part could become part of uv sync internally.

zanieb:
> I don't think I intend to support pip install of nanochat, I'm not sure why someone would want to do that or if that's a goal of the repo.

If you don't intend to publish it to PyPI, the approach in this pull request seems pretty reasonable. (If you did want to publish it, separating these into two separate packages would make it a bit more onerous to manage).

This breakage described in the pull request body seems sufficiently motivating:

> uv pip show nanochat worked but importing from a script outside the project directory failed.
> [...]
> Similar problem is reported https://github.com/PyO3/maturin/issues/490 (the .pth files are not generated for nanochat)

Whereas the following doesn't seem like a big deal to me:

> rustbpe got installed as a module not a package.

konstin:
A less intrusive alternative is having maturin manage both the Rust module and the Python sources. Currently, maturin sees the `rustbpe` module and ignores the `nanochat` directory. `import nanochat` works because Python imports from the current directory by default, causing the import errors in other directories in the OP of this PR. If we make the Rust code a submodule of nanochat, maturin can manage both: https://github.com/karpathy/nanochat/compare/master...konstin:konsti/better-maturin-structure

With this change, `uv sync` will manage both Rust and Python change, and rebuild when Rust sources change. You don't need to call `maturin develop` either, only `uv sync`.

svlandeg:
With `rustbpe` now being its own repo/package, I think this can probably be closed?

karpathy:
yes! closing ty

PR #175, opened 2025-10-26
Fix tok/sec metrics for base_train and mid_train when gradient accumulation is not 1

water-vapor:
Currently, when`grad_accum_steps` is not 1, the reported tokens-per-second (tok/sec) metric is incorrect. The issue arises because `world_tokens_per_fwdbwd` represents the number of tokens per step, while `dt `measures the total time for all gradient accumulation steps. As a result, the reported tok/sec value is underestimated by a factor of `1/grad_accum_steps`.

The fix is to replace `world_tokens_per_fwdbwd` with `total_batch_size` when computing tok/sec. Notably, the subsequent line
```python
flops_per_sec = num_flops_per_token * total_batch_size / dt
```
already uses total_batch_size correctly.


dendritic:
+1
I was just about propose an identical fix, then I found this PR

karpathy:
Not sure how this slipped thank you.

PR #182, opened 2025-10-26
Add Modal Deployment

svilupp:
- adds modal deployment scripts
- adds synthetic data gen pipeline


svilupp:
Apologies! Wrong origin!

PR #192, opened 2025-10-28
Remove need for pandas in the base eval script

vjdad4m:
The `scripts/base_eval.py` script no longer uses pandas. `pandas.read_csv` was replaced with `csv.DictReader` and the `pandas` import was removed.

*Note: we cannot remove `pandas` from the project entirely yet because it's still required transitively by the `datasets` package.*

### DIFF
--- a/scripts/base_eval.py
+++ b/scripts/base_eval.py
@@ -15,9 +15,9 @@ import time
 import json
 import random
 import yaml
+import csv
 from contextlib import nullcontext
 
-import pandas as pd
 import torch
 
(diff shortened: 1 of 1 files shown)


karpathy:
Yeah I'm not sure what to do here tbh. We're not removing the dependency anyway, so I'm not 100% sure if it's useful to bloat the script to remove the import here. And I think getting rid of `datasets` as a dependency might be harder. I'd still love to do it, but it might be non-trivial and it gives up a lot of flexibility.

PR #195, opened 2025-10-28
Remove explicit numpy dependency

Aj-esh:
The numpy library `numpy==1.26.4` is currently listed as a dependency for `nanochat`, but it isn’t being directly used in the project.    

* NumPy is not a direct dependency of `nanochat` — no module within the codebase imports or relies on it.
* Keeping it in `pyproject.toml` can lead to unnecessary dependency conflicts, especially since it pins `numpy==1.26.4`.
* Removing it is helping to reduce dependency footprint and avoid potential version conflicts for users or forks.

## Proposed Change:

Remove the numpy entry from the dependency list.

## Impact:

* No functional change to NanoChat’s behavior.
* Cleaner dependency management for developers and users forking or extending the repo.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,6 @@ dependencies = [
     "datasets>=4.0.0",
     "fastapi>=0.117.1",
     "files-to-prompt>=0.6",
-    "numpy==1.26.4",
     "psutil>=7.1.0",
     "regex>=2025.9.1",
     "setuptools>=80.9.0",


Aj-esh:
I felt the same, having said that since `numpy` isn’t currently being used, removing could helps avoid unnecessary conflicts and keeping the dependency list lean. If future extensions end up needing it, we can easily add it back.

Also, just a thought, do we really need to stick with `numpy==1.26` when the latest stable release is 2.3?

karpathy:
numpy is installed through pytorch as its dependency anyway right? So this line is a bit of a noop.

svlandeg:
> numpy is installed through pytorch as its dependency anyway right? So this line is a bit of a noop.

Yep, we get `numpy` from pytorch, so we don't need this line. But by having it, we're actively restring the `numpy` version, which is unnecessary.

PR #201, opened 2025-10-29
feat(engine.py): Sample unique initial tokens for each sequence in a batch

azekowka:
…batch

Before, when initiating a batch generation, the first sampled token was broadcasted to all sequences. This change now expands the output logits after a single efficient prefill, allowing for the sampling of a unique starting token for each sequence in the batch.

### DIFF
--- a/nanochat/engine.py
+++ b/nanochat/engine.py
@@ -216,6 +216,8 @@ class Engine:
         ids = torch.tensor([tokens], dtype=torch.long, device=device)
         logits = self.model.forward(ids, kv_cache=kv_cache_prefill)
         logits = logits[:, -1, :]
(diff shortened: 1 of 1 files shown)


karpathy:
Is this wrong? It's not actually sampling unique tokens per row?
I have a TODO to fix this bug, it's been too long sry ;\

karpathy:
Fixed in 8f979a8.

The fix expands the prefill logits to `num_samples` and samples independently for each row, plus simplifies the generation loop by moving the forward pass to the end of the loop (eliminating the `first_iteration` flag entirely).

Added a test `test_multi_sample_first_token_diversity` that verifies tokens are sampled independently.

PR #205, opened 2025-10-30
fix(eval): inference_mode, drop pandas in base_eval.py, typo & DDP note

dipeshbabu:
This PR makes scripts/base_eval.py lighter and more robust during CORE evaluation:

1. Fixes a docstring typo: “Evlauate” → “Evaluate”.

2. Wraps per-task evaluation in torch.inference_mode() and keeps device moves explicit. This reduces autograd overhead and helps avoid CUDA eval hiccups some users have reported.

3. Removes the pandas dependency and loads eval_meta_data.csv via the stdlib csv module (mirrors ongoing work to keep the repo dependency-lite and addresses the TODO already noted in this file). 

4. Hardens the metadata lookup: if a task row or the “Random baseline” field is missing, we raise a clear KeyError with the file path and task label.

5. Adds a brief comment clarifying that the task-level evaluation handles per-rank work so base_eval.py doesn’t try to re-shard.

### DIFF
--- a/scripts/base_eval.py
+++ b/scripts/base_eval.py
@@ -28,7 +28,6 @@ from nanochat.core_eval import evaluate_task
 # -----------------------------------------------------------------------------
 # nanoChat specific function dealing with I/O etc.
 
-
 def evaluate_model(model, tokenizer, device, max_per_task=-1):
     """
     Evaluate a base model on the CORE benchmark.
@@ -44,7 +43,7 @@ def evaluate_model(model, tokenizer, device, max_per_task=-1):
     with open(config_path, 'r') as f:
         config = yaml.safe_load(f)
     tasks = config['icl_tasks']
-
+    eval_metadata = pd.read_csv(eval_meta_data)
     eval_metadata = {}
     with open(eval_meta_data, "r", newline="", encoding="utf-8") as f:
         reader = csv.DictReader(f)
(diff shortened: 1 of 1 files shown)


karpathy:
I like a lot of these ideas but I really dislike the use of black formatter. The kinds of line breaks it creates are monstrous to my eyes.

karpathy:
I'm not sure if this PR is doing something substantial enough.

The pandas dep I agree we may want to delete.

The inference_mode feels less necessary given that we already have:

```python
@torch.no_grad()
def evaluate_example(idx, model, tokenizer, data, device, task_meta):
```

I used no_grad instead of inference mode because earlier in the repo I got some weird errors about inference mode tensors and I didn't want to deal with it. I realize it's a bit more modern preferred choice though, if possible. But then I'd be inclined to change the decorator on evaluate_example instead.

dipeshbabu:
Thanks for the feedback, @karpathy! I’ve narrowed the PR:

Kept the pandas → csv change, the safer metadata lookup, the docstring typo fix, and a short DDP note.

Removed inference_mode entirely, evaluation remains on @torch.no_grad() as you prefer.

This keeps the change substantial but safe (dependency trim + clearer errors) without altering eval behavior. If you ever want to revisit inference_mode, I’m happy to open a tiny follow-up that switches only the decorator on evaluate_example, but I won’t bundle that here.

dipeshbabu:
But the thing is there's an open CUDA usage error during eval issue; running eval in inference_mode is safe, idiomatic change that can prevent accidental grad/allocator work during evaluation.

scripts/base_eval.py
```diff
@@ -44,6 +44,11 @@ def evaluate_model(model, tokenizer, device, max_per_task=-1):
         config = yaml.safe_load(f)
     tasks = config['icl_tasks']
     eval_metadata = pd.read_csv(eval_meta_data)
```
karpathy, line 46: bug.
ok let me take a look

karpathy:
Pushed my own version with https://github.com/karpathy/nanochat/commit/cf587acb1a51003463c7eda250e95842802b80fd

PR #224, opened 2025-11-01
CPU Support, as bfloat16 params breaks inference

msaelices:
Error fixed in this PR:
```
RuntimeError: expected m1 and m2 to have the same dtype, but got: c10::BFloat16 != float
```


svlandeg (commented):
Hi @msaelices, thanks for the PR! Could you remove the formatting changes so the diff only shows the core of your contribution? 🙏

karpathy:
I see, ok I think possibly this makes sense. Will merge for now.

PR #228, opened 2025-11-02
Added tinyrun for minimal configuration

willhama:
I have seen many people discuss and try to run it on single GPUs. I myself have a 3080 GPU with 10GB VRAM workstation at home and was not able to run the speedrun.sh. Therefore, I have created tinyrun.sh, which is the same as speedrun.sh and run1000, but with adjusted parameters. I was able to run the whole pipeline end-to-end in roughly one hour.

I think it brings it a bit closer to people who do not have the budget of 100$ or are hesitating to rent a cluster for 25$ an hour or having a simple GPU setup at home.

@karpathy, what do you think?

### DIFF
--- a/tinyrun.sh
+++ b/tinyrun.sh
@@ -50,16 +50,6 @@ DATASET_DOWNLOAD_PID=$!
 python -m scripts.tok_train --max_chars=2000000000
 python -m scripts.tok_eval
 
-# Download the eval_bundle from s3 to evaluate CORE metric during training (~162MB)
-EVAL_BUNDLE_URL=https://karpathy-public.s3.us-west-2.amazonaws.com/eval_bundle.zip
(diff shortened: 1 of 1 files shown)


karpathy:
Possible I'd start a new Discussion specifically around this use case and paste it there. This way other people can also ask questions, contribute their versions, etc. I don't think I want this to be "officially supported" in master atm. People will try to run nanochat on all kinds of potatoes of large diversity and I don't wish to add them all to master .sh files. Please create a Discussion ty!

willhama:
No harsh feelings there. Makes sense to not bloat the repo with .sh files. Just out of curiosity: how do you run your code while you are developing? I guess you also run it on a lower-end hardware before you go for the cloud runners? @karpathy

PR #234, opened 2025-11-03
Fix subtle model-loading bug

odomobo:
lstrip will strip out more than just `_orig_mod.` . I encountered this bug on my own fork when trying to add a layer named `output_projection`, and the 'o' got stripped to `utput_projection`.

### DIFF
--- a/nanochat/checkpoint_manager.py
+++ b/nanochat/checkpoint_manager.py
@@ -72,7 +72,7 @@ def build_model(checkpoint_dir, step, device, phase):
             for k, v in model_data.items()
         }
     # Hack: fix torch compile issue, which prepends all keys with _orig_mod.
-    model_data = {k.lstrip("_orig_mod."): v for k, v in model_data.items()}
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
Oooh, nice catch @odomobo!

`lstrip` indeed works on a set of characters, and thus will strip additional characters beyond the `_orig_mod.` prefix if those characters match, like `o` does.

Definitely much better to use `removeprefix` here, which only removes the given word once exactly as it appears.

svlandeg:
FYI, quick repro:
```
mystrings = ["omg_other", "_orig_mod.omg_other"]
for mystring in mystrings:
    print(f"lstrip: {mystring} --> {mystring.lstrip('_orig_mod.')}")
    print(f"removeprefix: {mystring} --> {mystring.removeprefix('_orig_mod.')}")
```
gives
> lstrip: omg_other --> ther
removeprefix: omg_other --> omg_other
lstrip: _orig_mod.omg_other --> ther
removeprefix: _orig_mod.omg_other --> omg_other

odomobo:
@svlandeg thank you thank you!

PR #236, opened 2025-11-03
fix(eval): use UTF-8 when reading CORE JSONL and writing CSV

dipeshbabu:
This PR makes scripts/base_eval.py deterministic across platforms by opening the CORE JSONL with encoding='utf-8' (JSON defaults to UTF-8). On Windows/macOS, locale defaults (e.g., cp1252) can cause UnicodeDecodeError or silent mis-decoding.

Optional addition: write the output CSV with encoding='utf-8' and newline='' for consistent output.

### DIFF
--- a/scripts/base_eval.py
+++ b/scripts/base_eval.py
@@ -88,7 +88,7 @@ def evaluate_model(model, tokenizer, device, max_per_task=-1):
 
         # Load data for this task
         data_path = os.path.join(data_base_path, task_meta['dataset_uri'])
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
Yep, makes sense. I've often seen things fail on Windows & Mac over these kind of missing UTF8 encoding statements. Kind of silly to still have to do this in 2025, but best to avoid issues.

There's several other `open` statements in this file as well as others - should we update all of them?

PR #242, opened 2025-11-03
log grad norm during training

nitishpandey04:
log grad norm too during training, helpful to track training stability, learning dynamics, etc.

tested training after making the changes - 
image
wandb-
image

### DIFF
--- a/scripts/base_train.py
+++ b/scripts/base_train.py
@@ -274,7 +274,7 @@ for step in range(num_iterations + 1):
     # gradient clipping (TODO possibly experiment with)
     if grad_clip > 0.0:
         grad_norm = torch.nn.utils.clip_grad_norm_(orig_model.parameters(), grad_clip)
-        grad_norm = grad_norm.item()
+        grad_norm = grad_norm.item() # takes cpu and gpu sync, call it once and use everywhere
     # step the optimizers
     lrm = get_lr_multiplier(step)
     for opt in optimizers:
@@ -301,14 +301,10 @@ for step in range(num_iterations + 1):
     mfu = 100 * flops_per_sec / promised_flops_per_sec_h100 # in %
     if step > 10:
         total_training_time += dt # only count the time after the first 10 steps
-    print0(
-        f"step {step:05d}/{num_iterations:05d} ({pct_done:.2f}%) | "
-        f"loss: {debiased_smooth_loss:.6f} | "
-        + (f"grad_norm: {grad_norm.item():.5f} | " if grad_clip > 0.0 else "")
-        + f"lrm: {lrm:.2f} | dt: {dt * 1000:.2f}ms | "
-        f"tok/sec: {tok_per_sec:,} | mfu: {mfu:.2f} | "
-        f"total time: {total_training_time/60:.2f}m"
-    )
+    print_grad_norm = ""
+    if grad_clip > 0.0:
+        print_grad_norm = f" grad_norm: {grad_norm:.5f} |"
+    print0(f"step {step:05d}/{num_iterations:05d} ({pct_done:.2f}%) | loss: {debiased_smooth_loss:.6f} |{print_grad_norm} lrm: {lrm:.2f} | dt: {dt * 1000:.2f}ms | tok/sec: {tok_per_sec:,} | mfu: {mfu:.2f} | total time: {total_training_time/60:.2f}m")
     if step % 100 == 0:
         log_data = {
             "step": step,


svlandeg (approved):
Agreed that it might be nice to have. Will leave final decision with Andrej though.

karpathy:
- This will create error if grad clip is not used because grad_norm won't exist.
- Calls `.item()` twice is bad because it costs time and GPU sync to communicate between GPU and CPU, should ideally only be done a single time

cc @svlandeg

nitishpandey04:
> * This will create error if grad clip is not used because grad_norm won't exist.
> * Calls `.item()` twice is bad because it costs time and GPU sync to communicate between GPU and CPU, should ideally only be done a single time
> 
> cc @svlandeg

sorry, missed that edge case. will make the required changes.

nitishpandey04:
@karpathy, how should we handle the case when grad_clip is not used ? logging grad_norm as 0 if it doesn't exist has no point, so should we skip it in that case ? and log it if it is used ?

karpathy:
yeah... it's kind of annoying isn't it

scripts/base_train.py
```diff
@@ -300,19 +301,28 @@ def get_muon_momentum(it):
     mfu = 100 * flops_per_sec / promised_flops_per_sec_h100 # in %
     if step > 10:
         total_training_time += dt # only count the time after the first 10 steps
-    print0(f"step {step:05d}/{num_iterations:05d} ({pct_done:.2f}%) | loss: {debiased_smooth_loss:.6f} | grad_norm: {grad_norm.item():.5f} | lrm: {lrm:.2f} | dt: {dt * 1000:.2f}ms | tok/sec: {tok_per_sec:,} | mfu: {mfu:.2f} | total time: {total_training_time/60:.2f}m")
+    print0(
+        f"step {step:05d}/{num_iterations:05d} ({pct_done:.2f}%) | "
+        f"loss: {debiased_smooth_loss:.6f} | "
+        + (f"grad_norm: {grad_norm.item():.5f} | " if grad_clip > 0.0 else "")
```
karpathy: 👀

nitishpandey04:
> yeah... it's kind of annoying isn't it

@karpathy it is actually, but this is the cleanest way i know to make it look reasonable lol. any suggestions ?

scripts/base_train.py
svlandeg: Note that in the current code you still have the `item()` call twice - the second one should be deleted.

One option to make this a bit cleaner: define the string to be printed beforehand, then paste it into the large `print0` statement as-is.

Something like 
```
print_grad_norm = ""
if grad_clip > 0.0:
    print_grad_norm = f"grad_norm: {grad_norm:.5f} | "
```
and then in the middle of the original `print0` statement put `{print_grad_norm}`.

Or something such. Still not ideal, but the `if` in the middle of the printed f-strings feel worse for readability 🙈

nitishpandey04: agreed, that is cleaner. `if` statement in the middle of f-strings feels awkward

scripts/base_train.py
```diff
@@ -300,9 +301,12 @@ def get_muon_momentum(it):
     mfu = 100 * flops_per_sec / promised_flops_per_sec_h100 # in %
     if step > 10:
         total_training_time += dt # only count the time after the first 10 steps
-    print0(f"step {step:05d}/{num_iterations:05d} ({pct_done:.2f}%) | loss: {debiased_smooth_loss:.6f} | lrm: {lrm:.2f} | dt: {dt * 1000:.2f}ms | tok/sec: {tok_per_sec:,} | mfu: {mfu:.2f} | total time: {total_training_time/60:.2f}m")
+    print_grad_norm = ""
+    if grad_clip > 0.0:
+        print_grad_norm = f" grad_norm: {grad_norm:.5f} |"
```
nitishpandey04, line 306: if `grad_clip` is 0, `print_grad_norm` will be invisible in the main f-string
otherwise it will appear normally, like loss, lrm, etc.

karpathy:
I had some more little nits, the most important being that I don't like to copy paste conditions around (grad_clip > 0), because inevitably you always change one of them and leave the others unchanged, leading to bugs. So i made a variable for it. 

https://github.com/karpathy/nanochat/commit/c6b7ab744055d5915e6ccb61088de80c10cbaff9

closing this but ty

nitishpandey04:
learned some things from this pr, thanks @karpathy @svlandeg 
happy to contribute :)

PR #244, opened 2025-11-03
fix(checkpoint): handle bf16 on MPS by casting to fp32 during load

dipeshbabu:
Apple MPS lacks full bfloat16 support. If checkpoints were saved with bf16
weights, loading on MPS can error or upcast unpredictably. Mirror the existing
CPU behavior by casting bf16 tensors to float32 when device.type == "mps".

Scope: checkpoint loading only. CUDA paths unchanged. No metric/logic changes.

### DIFF
--- a/nanochat/checkpoint_manager.py
+++ b/nanochat/checkpoint_manager.py
@@ -65,7 +65,7 @@ def build_model(checkpoint_dir, step, device, phase):
     """
     assert phase in ["train", "eval"], f"Invalid phase: {phase}"
     model_data, optimizer_data, meta_data = load_checkpoint(checkpoint_dir, step, device, load_optimizer=False)
-    if device.type == "cpu":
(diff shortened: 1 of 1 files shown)


svlandeg:
Maintenance note: follow-up of https://github.com/karpathy/nanochat/pull/224

svlandeg (approved):
Hi @dipeshbabu, thanks for the PR! Makes sense to me.

jwnsu:
Your Apple silicon model and macos version? There is no issue with bf16 for M4.max and MacOS 26. Autocast to bf16 during training is also fine, but didn't see speed improvement over f32 either.

dipeshbabu:
> Your Apple silicon model and macos version? There is no issue with bf16 for M4.max and MacOS 26. Autocast to bf16 during training is also fine, but didn't see speed improvement over f32 either.

M1 max macOS 15.0

PR #247, opened 2025-11-04
Replace fcntl with filelock for Windows compatibility

barmag:
This pull request fixes an issue where the nanochat.dataset script fails on Windows due to the use of the fcntl module, which is not available on Windows.

The fcntl module is used in the download_file_with_lock function in nanochat/common.py for file locking. This pull request replaces the fcntl-based file locking with the filelock library, which is a cross-platform solution and is already a dependency of the project. This change allows the nanochat.dataset script to run successfully on Windows, making the project more accessible to a wider range of users.

I have tested this change on Windows and can confirm that the nanochat.dataset and scripts.tok_train scripts run without any issues.

### DIFF
--- a/dev/gen_synthetic_data.py
+++ b/dev/gen_synthetic_data.py
@@ -37,7 +37,7 @@ from concurrent.futures import ThreadPoolExecutor, as_completed
 
 from nanochat.common import get_base_dir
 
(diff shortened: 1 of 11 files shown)


karpathy:
TIL, ty

PR #251, opened 2025-11-05
Make NPROC_PER_NODE customizable in run1000.sh and speedrun.sh

vinjn:
Some nodes have less than 8 GPUs :)


run1000.sh
```diff
@@ -72,7 +72,7 @@ python -m scripts.tok_eval
 # 5) That's it, everything else (e.g. the learning rates) is adjusted automatically by the training script.
 
 # Number of processes/GPUs to use
-NPROC_PER_NODE=8
+NPROC_PER_NODE=${NPROC_PER_NODE:-8}
```
nitishpandey04, line 75: can you explain how will it pick the number of gpus now ?

PR #253, opened 2025-11-05
Add centralized wandb initialization utility to reduce code duplication

SermetPekin:
- This change centralizes the current approach of creating wandb or DummyWandb instances based on script variables, helping to avoid code repetition across scripts

### DIFF
--- a/scripts/mid_train.py
+++ b/scripts/mid_train.py
@@ -64,7 +64,7 @@ synchronize = torch.cuda.synchronize if device_type == "cuda" else lambda: None
 get_max_memory = torch.cuda.max_memory_allocated if device_type == "cuda" else lambda: 0
 
 # wandb logging init
-wandb_run = get_wandb("nanochat-mid" , run=run, master_process=master_process, user_config=user_config)
+wandb_run = get_wandb("nanochat-mid", run=run, master_process=master_process, user_config=user_config)
 
 # Load the model and tokenizer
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
I'm usually a fan of avoiding code duplication, but in this particular case I'm a bit on the fence. This PR changes only 2 lines to a single line (in 4 scripts), but the single line is somewhat less self-explanatory, as you don't see the `wandb.init` call explicitely anymore. So there's a trade-off with clarity here. I don't feel strongly either way though.

PS: it would be better to avoid non-functional edits such as moving imports & existing code, so that the PR's diff focuses only on the actual contribution and is easier to review.

SermetPekin:
> I'm usually a fan of avoiding code duplication, but in this particular case I'm a bit on the fence. This PR changes only 2 lines to a single line (in 4 scripts), but the single line is somewhat less self-explanatory, as you don't see the `wandb.init` call explicitely anymore. So there's a trade-off with clarity here. I don't feel strongly either way though.
> 
> PS: it would be better to avoid non-functional edits such as moving imports & existing code, so that the PR's diff focuses only on the actual contribution and is easier to review.

- My initial purpose was to make wandb optional by adding an optional import (using try - catch) with using wandb that would help some users not to be forced to install it. This centralization would help to use other options or non at all as it is the case in nanochat. 
- PS: those non functional edits make sense cause imports were not in good order and they are just too few to pollute the changes.

PR #255, opened 2025-11-06
Fix  torch.dtype mismatching  when running engine test.

howardgao96:
When running engine.py to view the result of kv cache, it raises the following error: 

```shell
RuntimeError: expected mat1 and mat2 to have the same dtype, but got: c10::BFloat16 != float
```

I add the `autocast_ctx` for `model.generate` and `engine.generate` to ensure that the data type of variables are the same. Here is the result: 

promt:  "The chemical formula of water is"

```shell
 H2O. It is made up of two hydrogen atoms and one oxygen atom. The chemical formula of water is H2O because it has two hydrogen atoms and one oxygen atom. The chemical formula of water is H2O because it has two hydrogen atoms and one oxygen atom. The chemical formula of water is H
Reference time: 1.10s
 H2O. It is made up of two hydrogen atoms and one oxygen atom. The chemical formula of water is H2O because it has two hydrogen atoms and one oxygen atom. The hydrogen atoms are bonded to the oxygen atom. The oxygen atom is bonded to the hydrogen atoms. The oxygen atom is bonded to the
Engine time: 0.83s
```

### DIFF
--- a/nanochat/engine.py
+++ b/nanochat/engine.py
@@ -332,7 +332,7 @@ if __name__ == "__main__":
     autocast_ctx = torch.amp.autocast(device_type=device_type, dtype=torch.bfloat16) if device_type == "cuda" else nullcontext()
 
     # load the model and tokenizer
-    model, tokenizer, meta = load_model("sft", device, phase="eval")
+    model, tokenizer, meta = load_model("base", device, phase="eval")
     bos_token_id = tokenizer.get_bos_token_id()
     # common hyperparameters
     kwargs = dict(max_tokens=64, temperature=0.0)


svlandeg (approved):
I can confirm that this error is being raised when executing `python engine.py`. Makes sense to me to use `autodetect_device_type` and `with autocast_ctx` as we do elsewhere. With these edits, the command runs correctly again.

nanochat/engine.py
```diff
@@ -327,8 +328,11 @@ def generate_batch(self, tokens, num_samples=1, **kwargs):
     import time
     # init compute
     ddp, ddp_rank, ddp_local_rank, ddp_world_size, device = compute_init()
+    device_type = autodetect_device_type()
+    autocast_ctx = torch.amp.autocast(device_type=device_type, dtype=torch.bfloat16) if device_type == "cuda" else nullcontext()
+
     # load the model and tokenizer
-    model, tokenizer, meta = load_model("base", device, phase="eval")
+    model, tokenizer, meta = load_model("sft", device, phase="eval")
```
svlandeg: Just noting that this edit is unrelated to the main fix of this PR, though I can see how you'd like to see the output of the SFT model instead of the base one. It doesn't really matter in the scheme of things here though.

svlandeg: Playing with this more, there are prompts that break this script for me using `sft`, e.g. when asking a question ("Is the sky blue?") and creating a turn-based conversation. The `model.generate` function will create a longer "conversation" in these cases, while `engine.generate` will break off after the first `<|assistant_end|>`. To be on the safe side, I'll revert this particular edit.

svlandeg:
```suggestion
    model, tokenizer, meta = load_model("base", device, phase="eval")
```

megan6969love:
Hey there Sent from my iPhoneOn Nov 14, 2025, at 10:29 AM, Andrej ***@***.***> wrote:﻿
Merged #255 into master.

megan6969love:
Sent from my iPhone

PR #256, opened 2025-11-06
fix(common): avoid destroying non-initialized DDP process group

dipeshbabu:
Split DDP checks into:
- is_ddp_requested(): env-based (torchrun) for pre-init decisions
- is_ddp_initialized(): true runtime state for cleanup

This avoids calling dist.destroy_process_group() on CPU/MPS runs launched
via torchrun where we never initialized a PG (we only init on CUDA).


karpathy:
Thank you!

PR #258, opened 2025-11-07
dataset downloader: add progress bar, essential info to stdout, and debug logs to `logs/` (no functional change)

h3nock:
This PR improves the dataset download experience by adding a progress bar, refining stdout to show only essential information, and redirecting detailed DEBUG logs to a dedicated file. 

Previously, the script streamed interleaved output from all worker processes directly to stdout. While this is helpful for debugging when things go wrong, such cases are far less common than normal runs. So, for most cases, it just floods the terminal with noise and leaves developers waiting blindly, unsure how much progress has been made or how long it might take.

What’s new:
- Progress bar: the download process now includes a progress bar that shows completion percentage and a rough ETA.
- Readable stdout: only essential information is shown like shard destinations, log location, and a clear progress bar for run status.
- Dedicated logs: full debug details go to logs/dataset_download.log.
- No performance change (verified on the same setup):

        - Before (no progress bar): 13:30:59 → 13:34:14 (~3m15s)
        - After (with progress bar): 16:40:09 → 16:43:24 (~3m15s)

Screenshots (showing progress display, completion output, and log file details):

progress

done

log_file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,12 @@ __pycache__/
 dev-ignore/
 report.md
 eval_bundle/
+logs/
+
+# Secrets
+.env
+
+# Local setup
+.claude
+CLAUDE.md
+wandb/


h3nock:
@svlandeg made the work share factor configurable and also added details about its effect including an example

.gitignore
```diff
@@ -4,4 +4,5 @@ __pycache__/
 rustbpe/target/
 dev-ignore/
 report.md
-eval_bundle/
\ No newline at end of file
+eval_bundle/
+logs/
```
svlandeg, line 7:
```suggestion
```

Issue #257, opened 2025-11-07
Getting cuda error running speedrun.sh on H100:8


ggilley:
Want to do my own run. Using lambda.ai. H100:8

git clone from 20 minutes ago.

ran speedrun.sh. Getting cuda error Any ideas:
```

Step 00000 | Validation bpb: 3.3013
[rank4]: Traceback (most recent call last):
[rank4]:   File "/home/ubuntu/.local/share/uv/python/cpython-3.10.19-linux-x86_64-gnu/lib/python3.10/runpy.py", line 196, in _run_module_as_main
[rank4]:     return _run_code(code, main_globals, None,
[rank4]:   File "/home/ubuntu/.local/share/uv/python/cpython-3.10.19-linux-x86_64-gnu/lib/python3.10/runpy.py", line 86, in _run_code
[rank4]:     exec(code, run_globals)
[rank4]:   File "/home/ubuntu/test/nanochat/scripts/base_train.py", line 272, in <module>
[rank4]:     loss.backward()
[rank4]:   File "/home/ubuntu/test/nanochat/.venv/lib/python3.10/site-packages/torch/_tensor.py", line 625, in backward
[rank4]:     torch.autograd.backward(
[rank4]:   File "/home/ubuntu/test/nanochat/.venv/lib/python3.10/site-packages/torch/autograd/__init__.py", line 354, in backward
[rank4]:     _engine_run_backward(
[rank4]:   File "/home/ubuntu/test/nanochat/.venv/lib/python3.10/site-packages/torch/autograd/graph.py", line 841, in _engine_run_backward
[rank4]:     return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
[rank4]:   File "/home/ubuntu/test/nanochat/.venv/lib/python3.10/site-packages/torch/autograd/function.py", line 315, in apply
[rank4]:     return user_fn(self, *args)
[rank4]:   File "/home/ubuntu/test/nanochat/.venv/lib/python3.10/site-packages/torch/_functorch/_aot_autograd/runtime_wrappers.py", line 2303, in backward
[rank4]:     return impl_fn()
[rank4]:   File "/home/ubuntu/test/nanochat/.venv/lib/python3.10/site-packages/torch/_functorch/_aot_autograd/runtime_wrappers.py", line 2289, in impl_fn
[rank4]:     out = CompiledFunction._backward_impl(ctx, all_args)
[...]

PR #259, opened 2025-11-08
Fix GitHub Issue #257: Getting cuda error running speedrun.sh on H100:8

dvvaliya369:
**Agent:** blackbox
**Model:** blackboxai/blackbox-pro


ggilley:
Interesting. I'm confused because he specifically said the config was for H100:8 on lambda.ai. So I'm surprised I failed. Any idea why it's failing now (I did try torch 2.8 which was in the original uv.lock file)

PR #262, opened 2025-11-08
make promised_flops_per_gpu global, configurable variable

nitishpandey04:
**Problem:**
I have seen people discuss their `mfu` in the Discussions, but they are often stating the wrong numbers as they are not running on H100s. 

**Change made:**
Make the variable more visible to users so they can set it and log correct `mfu` for their setup. 
Users need to do one search to get the bfloat16 non sparsity flops for their gpu and set the value to the variable (properly explained as comments in code).

A small change to allow `nanochatters` to discuss true `mfu` numbers.
Let's go 🚀

### DIFF
--- a/scripts/base_train.py
+++ b/scripts/base_train.py
@@ -59,7 +59,7 @@ core_metric_every = 2000 # every how many steps to evaluate the core metric (-1 
 core_metric_max_per_task = 500 # examples per task in estimating the core metric
 sample_every = 2000 # every how many steps to sample from the model
 # Logging
-promised_flops_per_gpu = 989e12 # Only used for logging. Set it to "without sparsity, bfloat16 flops for a single gpu", to log correct `mfu`. Default is set for H100 SXM, bfloat16, without 2:4 sparsity.
+promised_flops_per_gpu = 989e12 # To log the correct `mfu`, set this to the value of your GPU, "without sparsity, bfloat16 flops for a single gpu". Default is set for H100 SXM, bfloat16, without 2:4 sparsity.
 # Output
 model_tag = "" # optionally override the model tag for the output checkpoint directory name
 # now allow CLI to override the settings via the configurator lol


nitishpandey04:
here's an example of checking the flops for your gpu - https://www.perplexity.ai/search/what-is-estimated-flops-for-h1-Dgd_4PLKRVGghD58IDx51w#0

scripts/base_train.py
```diff
@@ -58,6 +58,8 @@
 core_metric_every = 2000 # every how many steps to evaluate the core metric (-1 = disable)
 core_metric_max_per_task = 500 # examples per task in estimating the core metric
 sample_every = 2000 # every how many steps to sample from the model
+# Logging
+promised_flops_per_gpu = 989e12 # Only used for logging. Set it to "without sparsity, bfloat16 flops for a single gpu", to log correct `mfu`. Default is set for H100 SXM, bfloat16, without 2:4 sparsity.
```
svlandeg: No need to repeat "logging" twice + some little tweaks to the description
```suggestion
promised_flops_per_gpu = 989e12 # To log the correct `mfu`, set this to the value of your GPU, "without sparsity, bfloat16 flops for a single gpu". Default is set for H100 SXM, bfloat16, without 2:4 sparsity.
```

nitishpandey04:
> The change looks good in principle, and makes it easier for users to set this according to their system. I know Andrej is wary of having too many parameters/configurations in the code though, so I'll defer to him whether he wants to merge this or not.

Sure. Thanks @svlandeg

nitishpandey04:
@svlandeg @karpathy 
i feel here it is about correctness also
when the logs are incorrect, users should know about it
and they should have the lever to correct it

karpathy:
There is a way to make this better but it involves doing the mfu like torchtitan does, not like this.

PR #282, opened 2025-11-12
Clarify assertion comments for KV cache dimensions

YanivDorGalron:


svlandeg (approved):
LGTM, thanks!

karpathy:
actually i'm confused by the new comment

svlandeg:
The main point is the removal of "batch_size" in the comment for L110, as the batch size is compared at L113 (`ix == 2`). 

The additional comment is about `ix=1`, in which case the values should be `2` because this is hard-coded at L90: 
```
self.kv_shape = (num_layers, 2, batch_size, num_heads, seq_len, head_dim)
```

nanochat/engine.py
```diff
@@ -108,7 +108,7 @@ def prefill(self, other):
         assert other.kv_cache is not None, "Cannot prefill with a None KV cache"
         for ix, (dim1, dim2) in enumerate(zip(self.kv_shape, other.kv_shape)):
             if ix in [0, 1, 3, 5]:
-                # num_layers, batch_size, num_heads, head_dim must match
+                # num_layers, num_heads, head_dim must match, ix=1 should always be 2.
```
svlandeg, line 111:
```suggestion
                # num_layers, num_heads, head_dim must match
```

YanivDorGalron: I added the comment
> ix = 1 should always be 2

to emphasize why the list in the if statement includes four indices that must always be identical and that the dimension at ix = 1 is fixed at 2, corresponding to the caching of both keys and values.

svlandeg: Yep, I know ;-) But the main contribution of this PR is removing `batch_size` from the comment, as it's incorrect here, so I suggested to focus on that ;-)

svlandeg:
Now fixed on `master` [here](https://github.com/karpathy/nanochat/commit/91f09ccd0d48daf89eee6ef7fcec05977fd87068)

YanivDorGalron:
Great!

PR #286, opened 2025-11-13
update eval_tokens calc, remove hardcode

nitishpandey04:
change is significant enough to call this nanochat v2 i think 🤔


nitishpandey04:
Understood. Let's place this on hold until I have some clarity on `total_batch_size` discussion - https://github.com/karpathy/nanochat/discussions/287

karpathy:
I specifically don't think this is a good idea at all. This makes the reported validation loss different and incomparable for different people who train at different batch sizes.

PR #294, opened 2025-11-14
Manage the Python module with maturin

svlandeg:
Created as an alternative to https://github.com/karpathy/nanochat/pull/172, with thanks to @konstin 🙏 

Thought it would be easier to have this as an actual PR to compare both approaches.

@konstin: as I understand, we could/should then also remove the `maturin develop` lines in `run1000.sh`, `runcpu.sh` and `speedrun.sh`? And move the rust install to before `uv sync`?

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -311,7 +311,7 @@ name = "exceptiongroup"
 version = "1.3.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "typing-extensions", marker = "python_full_version < '3.12' or (extra == 'extra-8-nanochat-cpu' and extra == 'extra-8-nanochat-gpu')" },
+    { name = "typing-extensions", marker = "python_full_version < '3.11' or (extra == 'extra-8-nanochat-cpu' and extra == 'extra-8-nanochat-gpu')" },
 ]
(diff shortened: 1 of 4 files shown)


uv.lock
```diff
@@ -311,7 +311,7 @@ name = "exceptiongroup"
 version = "1.3.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "typing-extensions", marker = "python_full_version < '3.12' or (extra == 'extra-8-nanochat-cpu' and extra == 'extra-8-nanochat-gpu')" },
+    { name = "typing-extensions", marker = "python_full_version < '3.11' or (extra == 'extra-8-nanochat-cpu' and extra == 'extra-8-nanochat-gpu')" },
```
svlandeg, line 314: It looks like most (if not all) of the edits here in `uv.lock` will disappear from the diff when merging in the most recent `master` (but I don't have the permission to push to this branch directly)

konstin:
> as I understand, we could/should then also remove the `maturin develop` lines in `run1000.sh`, `runcpu.sh` and `speedrun.sh`? And move the rust install to before `uv sync`?

Yep!

svlandeg:
With `rustbpe` now being its own repo/package, this has become obsolete.

PR #296, opened 2025-11-16
stats: include model_tag and model_step in /stats

intently:
Adds model_tag and model_step to /stats. How to test: start server with --model-tag d20 and curl /stats.

### DIFF
--- a/scripts/chat_web.py
+++ b/scripts/chat_web.py
@@ -45,8 +45,8 @@ from pydantic import BaseModel
 from typing import List, Optional, AsyncGenerator
 from dataclasses import dataclass
 from contextlib import nullcontext
-from nanochat.common import compute_init, autodetect_device_type
-from nanochat.checkpoint_manager import load_model
+from nanochat.common import compute_init, autodetect_device_type, get_base_dir
+from nanochat.checkpoint_manager import load_model, find_largest_model, find_last_step
 from nanochat.engine import Engine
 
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
I see how this could be useful, but the current implementation replicates a lot of code that already happens under the hood when loading the model (L227-L254).

One potential other solution would be to store some additional meta data about the model in the `Worker` class, for instance after loading the model in `initialize` we could fetch the `meta_data` (what is currently `_`) and set it as a var `model_meta_data` on the `Worker` instance. 

```
            model, tokenizer, meta_data = load_model(source, device, phase="eval", model_tag=model_tag, step=step)
            ...
            worker = Worker(
                gpu_id=gpu_id,
                device=device,
                engine=engine,
                model_meta_data=meta_data,
                tokenizer=tokenizer,
                autocast_ctx=autocast_ctx
            )
```

Then in `stats`, we could do something like 
```
"model_meta_data": worker_pool.workers[0].model_meta_data,
```
which would pull in all the meta data or we could select something like the `model_meta_data["step"]` and `model_meta_data["model_config"]["n_layer"]` values, which would then be functionaly equivalent to the implementation here, but without having to duplicate all the tag/step determination code. I think I would personally just print the whole `model_meta_data` dict, which contains interesting other stats like loss & accuracy numbers, vocab size, etc.

PR #301, opened 2025-11-18
Fix a TypeError from NoneType

Zack-RT:
run_command may return None, add a if to avoid "object of type 'NoneType' has no len()"


megan6969love:
Stop writing this email take me off of your list of emailsSent from my iPhoneOn Nov 19, 2025, at 4:45 AM, Zack-RT ***@***.***> wrote:﻿
Closed #301.

PR #306, opened 2025-11-19
change test/train split approach to fix bug in spelling bee task

ericsilberstein1:
I noticed seeing the same words for the train and test splits for the SpellingBee and SimpleSpelling tasks, just offset by 1. The cause seems to be this line:

```
seed = index if self.split == 'train' else -(index + 1)
```

but 

```
>>> import random
>>> random.seed(5)
>>> random.random()
0.6229016948897019
>>> random.seed(-5)
>>> random.random()
0.6229016948897019
```

Proposing the fix in this PR.

Noticed the issue while investigating chat eval after midtraining d20 [here](https://github.com/ericsilberstein1/learn-nanochat/blob/master/challenge-28-midtrain-d20/investigate-questions.ipynb).


karpathy:
wait WHAT :O

PR #311, opened 2025-11-20
Validate checkpoint file format in find_last_step

pyrytakala:
Fixes `find_last_step` to properly validate checkpoint file format using regex, preventing crashes when non-standard files exist in the checkpoint directory.

Example crash:
```
# Given files: model_000100.pt, model_000200.pt, model_000200_backup.pt
# Processing model_000200_backup.pt:
split("_")[-1]  # → "backup.pt"
split(".")[0]   # → "backup"
int("backup")   # → ValueError: invalid literal for int() with base 10: 'backup'
```

### DIFF
--- a/nanochat/checkpoint_manager.py
+++ b/nanochat/checkpoint_manager.py
@@ -117,7 +117,7 @@ def find_last_step(checkpoint_dir):
     checkpoint_files = [f for f in os.listdir(checkpoint_dir) if re.search(r'model_(\d+)\.pt$', f)]
     if not checkpoint_files:
         raise FileNotFoundError(f"No checkpoints found in {checkpoint_dir}")
-    last_step = int(max(f.split("_")[-1].split(".")[0] for f in checkpoint_files))
+    last_step = max(int(f.split("_")[-1].split(".")[0]) for f in checkpoint_files)
     return last_step
 
 # -----------------------------------------------------------------------------
--- a/nanochat/checkpoint_manager.py
+++ b/nanochat/checkpoint_manager.py
@@ -117,7 +117,7 @@ def find_last_step(checkpoint_dir):
     checkpoint_files = [f for f in os.listdir(checkpoint_dir) if re.search(r'model_(\d+)\.pt$', f)]
     if not checkpoint_files:
         raise FileNotFoundError(f"No checkpoints found in {checkpoint_dir}")
-    last_step = int(max(re.search(r'model_(\d+)\.pt$', f).group(1) for f in checkpoint_files))
+    last_step = int(max(f.split("_")[-1].split(".")[0] for f in checkpoint_files))
     return last_step
 
 # -----------------------------------------------------------------------------
--- a/nanochat/checkpoint_manager.py
+++ b/nanochat/checkpoint_manager.py
@@ -118,7 +118,16 @@ def find_last_step(checkpoint_dir):
     checkpoint_files = glob.glob(os.path.join(checkpoint_dir, "model_*.pt"))
     if not checkpoint_files:
         raise FileNotFoundError(f"No checkpoints found in {checkpoint_dir}")
-    last_step = int(max(os.path.basename(f).split("_")[-1].split(".")[0] for f in checkpoint_files))
+    # Use regex to match only valid checkpoint files (model_<digits>.pt) and ignore malformed files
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
Hi, thanks for the PR!

Just out of curiosity, how does a file like `model_000200_backup.pt` end up in your directory? I don't think the current scripts do anything like that, right?

Anyway, if we want to make this more robust, we should simply refactor L118 only, there's no need for 10 additional code lines 🙂 . We could simply rewrite L118 into

```
checkpoint_files = [f for f in os.listdir(checkpoint_dir) if re.search(r'model_(\d+)\.pt$', f)]
```
and then we'd apply the more stringent/correct pattern immediately. 

I don't see a way to do this directly with `glob.glob()` as that doesn't support `*` for a variable sequence like `[0-9]*`. (and repeating `[0-9]` 6 times in the pattern would be pretty ugly)

pyrytakala:
Thanks for the suggestion @svlandeg . I've implemented your suggestion, much cleaner!

To answer your question: the current scripts don't create backup files, but I wanted to make the function robust against edge cases where someone might manually create or rename checkpoint files

nanochat/checkpoint_manager.py
```diff
@@ -115,10 +114,10 @@ def find_largest_model(checkpoint_dir):
 
 def find_last_step(checkpoint_dir):
     # Look into checkpoint_dir and find model_<step>.pt with the highest step
-    checkpoint_files = glob.glob(os.path.join(checkpoint_dir, "model_*.pt"))
+    checkpoint_files = [f for f in os.listdir(checkpoint_dir) if re.search(r'model_(\d+)\.pt$', f)]
     if not checkpoint_files:
         raise FileNotFoundError(f"No checkpoints found in {checkpoint_dir}")
-    last_step = int(max(os.path.basename(f).split("_")[-1].split(".")[0] for f in checkpoint_files))
+    last_step = int(max(re.search(r'model_(\d+)\.pt$', f).group(1) for f in checkpoint_files))
```
svlandeg: I don't even think this edit is needed. We've already ensured that the `checkpoint_files` have names such as `model_123456.pt` and nothing else, so we can go ahead and keep the original simple string `split` and avoid running another regex search.

Actually, one more thing - we can even remove `os.path.basename` as our new `checkpoint_files` don't contain the full paths, just the file names (as we switched from `path.join` to `f in os.listdir` ...)

So this line can become
```suggestion
    last_step = int(max(f.split("_")[-1].split(".")[0] for f in checkpoint_files))
```
Could you double check on your dummy example? On my end this works fine. 🙏

pyrytakala: Nice one! Works also on my end

nanochat/checkpoint_manager.py
```diff
@@ -115,10 +114,10 @@ def find_largest_model(checkpoint_dir):
 
 def find_last_step(checkpoint_dir):
     # Look into checkpoint_dir and find model_<step>.pt with the highest step
-    checkpoint_files = glob.glob(os.path.join(checkpoint_dir, "model_*.pt"))
+    checkpoint_files = [f for f in os.listdir(checkpoint_dir) if re.search(r'model_(\d+)\.pt$', f)]
     if not checkpoint_files:
         raise FileNotFoundError(f"No checkpoints found in {checkpoint_dir}")
-    last_step = int(max(os.path.basename(f).split("_")[-1].split(".")[0] for f in checkpoint_files))
+    last_step = int(max(f.split("_")[-1].split(".")[0] for f in checkpoint_files))
```
svlandeg:
```suggestion
    last_step = max(int(f.split("_")[-1].split(".")[0]) for f in checkpoint_files)
```
Apply suggestion from #323 by @kuizhiqing

svlandeg (approved):
I've now also included @kuizhiqing's suggestion to apply the `max` operator after converting to `int`. This will come down to the exact same thing, because the file names are formatted as `"model_{step:06d}.pt"`, with each `int` encoded using the same number of digits, so the `max` will in fact work correctly on the strings as well. It's just that semantically it does make more sense to me to apply `int` first, and then the `max` - it's also more robust.

So all in all this PR looks good to me - all changes are mostly just for robustness of the code, but they are minimal changes that do no obfuscate any behaviour IMO.

PR #312, opened 2025-11-20
Fix: Validate stop parameter against dataset size

pyrytakala:
Fixes issue where training would crash when stop > dataset_size.

Sample test script below.
```
"""
Run with:
    uv run python test_training_bug_real.py
"""

import os
os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"

import torch
from tasks.common import Task, TaskMixture
from nanochat.gpt import GPT, GPTConfig
from nanochat.tokenizer import get_tokenizer

class MinimalTask(Task):
    def __init__(self, name, dataset_size=20, **kwargs):
        super().__init__(**kwargs)
        self.name = name
        self.data = [
            {
                "messages": [
                    {"role": "user", "content": f"{name} question {i}?"},
                    {"role": "assistant", "content": f"{name} answer {i}."}
                ]
            }
            for i in range(dataset_size)
        ]
        self.length = len(self.data)
    
    def num_examples(self):
        return self.length
    
    def get_example(self, index):
        return self.data[index]

def create_tiny_model(device, vocab_size=1000):
    config = GPTConfig(
        sequence_len=128,
        vocab_size=vocab_size,
        n_layer=2,
        n_head=2,
        n_kv_head=2,
        n_embd=64,
    )
    with torch.device("meta"):
        model = GPT(config)
    model.to_empty(device=device)
    model.init_weights()
    return model

def sft_data_generator(dataset, tokenizer, device, batch_size, ddp_rank=0, ddp_world_size=1, max_steps=None):
    """
    Parameterized version of sft_data_generator from scripts/chat_sft.py.
    """
    pad_token_id = tokenizer.encode_special("<|assistant_end|>")
    def collate_and_yield(batch):
        nrows = len(batch)
        ncols = max(len(ids) for ids, mask in batch) - 1
        inputs = torch.full((nrows, ncols), pad_token_id, dtype=torch.long)
        targets = torch.full((nrows, ncols), -1, dtype=torch.long)
        for i, (ids, mask) in enumerate(batch):
            n = len(ids)
            ids_tensor = torch.tensor(ids, dtype=torch.long)
            inputs[i, :n-1] = ids_tensor[:-1]
            row_targets = ids_tensor[1:]
            mask_tensor = torch.tensor(mask[1:], dtype=torch.long)
            row_targets[mask_tensor == 0] = -1
            targets[i, :n-1] = row_targets
        inputs = inputs.to(device)
        targets = targets.to(device)
        return inputs, targets
    batch = []
    steps = 0
    while True:
        for i in range(ddp_rank, len(dataset), ddp_world_size):
            if max_steps and steps >= max_steps:
                if batch:
                    yield collate_and_yield(batch)
                return
            doc = dataset[i]
            ids, mask = tokenizer.render_conversation(doc)
            batch.append((ids, mask))
            if len(batch) == batch_size:
                yield collate_and_yield(batch)
                batch = []
                steps += 1

def train_model(dataset, tokenizer, device, num_steps=5):
    model = create_tiny_model(device, vocab_size=tokenizer.get_vocab_size())
    model.train()
    model.to(device)
    
    optimizer = torch.optim.AdamW(model.parameters(), lr=1e-4)
    train_loader = sft_data_generator(dataset, tokenizer, device, batch_size=2, ddp_rank=0, ddp_world_size=1, max_steps=num_steps)
    
    for step, (inputs, targets) in enumerate(train_loader):
        optimizer.zero_grad()
        loss = model(inputs, targets=targets)
        loss.backward()
        optimizer.step()
        
        print(f"Step {step+1}/{num_steps}: loss = {loss.item():.4f}")
        if step + 1 >= num_steps:
            break

device = torch.device("cpu")
tokenizer = get_tokenizer()

# Training run 1
train_ds = TaskMixture([
    MinimalTask("TaskA", dataset_size=20, stop=15),
    MinimalTask("TaskB", dataset_size=15, stop=10),
])
train_model(train_ds, tokenizer, device, num_steps=5)

# Training run 2
val_ds = TaskMixture([
    MinimalTask("ValTask1", dataset_size=20, stop=15),
    MinimalTask("ValTask2", dataset_size=10, stop=15) # // stop > dataset_size
])
train_model(val_ds, tokenizer, device, num_steps=20)

```

### DIFF
--- a/tasks/common.py
+++ b/tasks/common.py
@@ -39,12 +39,12 @@ class Task:
         start = self.start
         num_ex = self.num_examples()
         if self.stop is not None and self.stop > num_ex:
-                # Warn once, then cap stop
-                logger.warning(
-                    f"Stop parameter ({self.stop}) exceeds dataset size ({num_ex}). "
-                    f"Using {num_ex} examples instead."
-                )
-                self.stop = num_ex
+            # Warn once, then cap stop
+            logger.warning(
+                f"Stop parameter ({self.stop}) exceeds dataset size ({num_ex}). "
+                f"Using {num_ex} examples instead."
+            )
+            self.stop = num_ex
         stop = num_ex if self.stop is None else self.stop
         step = self.step
         span = stop - start


pyrytakala:
Thanks for the review.

> So basically you want to guard against a user providing a stop value that is larger than the dataset has examples, right?

Yes, exactly

I went with option 2 (graceful handling) and added a warning. The challenge with option 1 (assert in constructor) is that num_examples() is abstract in the base Task class, and subclasses load their data after calling super().__init__(), so we can't validate stop <= num_examples() in the base constructor. That would require updating each subclass.

tasks/common.py
```diff
@@ -34,7 +34,18 @@ def get_example(self, index):
...
+            stop = min(self.stop, num_ex)  # Gracefully cap at dataset size
+            if self.stop > num_ex:
+                import warnings
+                warnings.warn(
+                    f"Stop parameter ({self.stop}) exceeds dataset size ({num_ex}). "
+                    f"Using {num_ex} examples instead.",
+                    UserWarning
+                )
+        else:
+            stop = self.num_examples()
```
svlandeg: I see what you mean about the constructor.

We do want to avoid emitting this warning every single time when `len()` is called though. 

What if we do something like this
```suggestion
        num_ex = self.num_examples()
        if self.stop is not None and self.stop > num_ex:
            self.stop = num_ex
            import warnings
                warnings.warn(
                    f"Stop parameter ({self.stop}) exceeds dataset size ({num_ex}). "
                    f"Using {num_ex} examples instead.",
                    UserWarning
                )
        stop = num_ex if self.stop is None else self.stop    
```
This caps `self.stop` once and emits the warning once, and from then on when this task is used, this issue shouldn't occur anymore.

I'd also move the `warnings` import to the top of the file, or better yet, use a `logger = logging.getLogger()` as the rest of the code base does.

pyrytakala: Good suggestions! Updated now.

tasks/common.py
```diff
@@ -34,7 +37,18 @@ def get_example(self, index):
...
+            if self.stop > num_ex:
+                # Warn once, then cap stop
+                logger.warning(
+                    f"Stop parameter ({self.stop}) exceeds dataset size ({num_ex}). "
+                    f"Using {num_ex} examples instead."
+                )
+                self.stop = num_ex
+            stop = self.stop
+        else:
+            stop = self.num_examples()
```
svlandeg:
```suggestion
        stop = self.num_examples() if self.stop is None else self.stop
```

tasks/common.py
```diff
@@ -34,6 +37,15 @@ def get_example(self, index):
...
+        if self.stop is not None:
+            num_ex = self.num_examples()
+            if self.stop > num_ex:
+                # Warn once, then cap stop
+                logger.warning(
+                    f"Stop parameter ({self.stop}) exceeds dataset size ({num_ex}). "
+                    f"Using {num_ex} examples instead."
+                )
+                self.stop = num_ex
         stop = self.num_examples() if self.stop is None else self.stop
```
svlandeg:
```suggestion
        num_ex = self.num_examples()
        if self.stop is not None and self.stop > num_ex:
                # Warn once, then cap stop
                logger.warning(
                    f"Stop parameter ({self.stop}) exceeds dataset size ({num_ex}). "
                    f"Using {num_ex} examples instead."
                )
                self.stop = num_ex
        stop = num_ex if self.stop is None else self.stop
```

tasks/common.py
```diff
@@ -34,7 +37,15 @@ def get_example(self, index):
...
         start = self.start
-        stop = self.num_examples() if self.stop is None else self.stop
+        num_ex = self.num_examples()
+        if self.stop is not None and self.stop > num_ex:
+                # Warn once, then cap stop
+                logger.warning(
+                    f"Stop parameter ({self.stop}) exceeds dataset size ({num_ex}). "
+                    f"Using {num_ex} examples instead."
+                )
+                self.stop = num_ex
```
svlandeg:
```suggestion
            # Warn once, then cap stop
            logger.warning(
                f"Stop parameter ({self.stop}) exceeds dataset size ({num_ex}). "
                f"Using {num_ex} examples instead."
            )
            self.stop = num_ex
```

tasks/common.py
```diff
@@ -34,7 +37,15 @@ def get_example(self, index):
...
-        stop = self.num_examples() if self.stop is None else self.stop
+        num_ex = self.num_examples()
+        if self.stop is not None and self.stop > num_ex:
+            # Warn once, then cap stop
+            logger.warning(
+                f"Stop parameter ({self.stop}) exceeds dataset size ({num_ex}). "
+                f"Using {num_ex} examples instead."
+            )
+            self.stop = num_ex
+        stop = num_ex if self.stop is None else self.stop
```
svlandeg, line 48: @karpathy: Alternatively, if we don't want this warning code here at this place and blow up the function implementation, we could just rewrite the original L37 to
```
        stop = self.num_examples() if self.stop is None else min(self.num_examples(), self.stop)
```
and cap it silently.

PR #326, opened 2025-11-23
fix(model): apply float32 cast before logits softcapping

spjosyula:
This change ensures that the logits softcapping operation (tanh) is performed in float32 precision rather than bfloat16. Previously, the code cast to float32 after the tanh operation, which meant the non-linearity was computed with bfloat16 precision.
I feel the overhead is statistically negligible compared to the stability gained for extended training runs.

### DIFF
--- a/nanochat/gpt.py
+++ b/nanochat/gpt.py
@@ -265,13 +265,14 @@ class GPT(nn.Module):
             # training mode: compute and return the loss
             # TODO: experiment with Liger Kernels / chunked cross-entropy etc.
             logits = self.lm_head(x)
-            logits = softcap * torch.tanh(logits / softcap) # logits softcap
             logits = logits.float() # use tf32/fp32 for logits
+            logits = softcap * torch.tanh(logits / softcap) # logits softcap
             loss = F.cross_entropy(logits.view(-1, logits.size(-1)), targets.view(-1), ignore_index=-1, reduction=loss_reduction)
             return loss
         else:
             # inference mode: compute and return the logits
             logits = self.lm_head(x)
+            logits = logits.float() # use tf32/fp32 for logits
             logits = softcap * torch.tanh(logits / softcap) # logits softcap
             return logits
 


spjosyula:
> I don't have a good intuition on how much this would affect stability of longer training runs. Have you done any experiments by any chance, measuring the impact of this? Just out of curiosity.
> 
> And what's the impact on the inference mode?

I ran simple numerical precision tests to check how tanh saturates. Currently the lm_head is on AdamW with zero weight decay. As far as I have analysed there is no constraint from the logits growing more than the bfloat16 saturation point (~55). After 55, tanh becomes 1 so gradients become 0. In float32 gradients wont vanish
Regarding the inference mode, high-confidence tokens collapse to identical softcap values due to bfloat16 saturation. (raw logits of 60 and 58 both become exactly 15.0). So the model's preferences between them is gone. Float32 preserves these distinctions.

karpathy:
"I feel the overhead is statistically negligible compared to the stability gained for extended training runs."

This is a fairly large tensor though right? Of shape (B, T, V) where V is vocab size - it's huge. I'm not 100% sure about the memory impact here with torch compile. Technically you could backprop through all of this in place and not incur overhead but I'm not sure if torch compile would do it too. I think I'll have to profile this a bit more carefully.

karpathy:
Composer 1 helped me take a look. Its code:

```python
"""
Test script to measure memory usage difference between:
1. Converting to fp32 AFTER tanh (current)
2. Converting to fp32 BEFORE tanh (proposed)

This helps understand the memory impact when using torch.compile.
"""

import torch
import torch.nn as nn
import torch.nn.functional as F

def get_memory_mb():
    """Get current GPU memory usage in MB"""
    if torch.cuda.is_available():
        return torch.cuda.memory_allocated() / 1024 / 1024
    return 0

def reset_memory():
    """Reset memory tracking"""
    if torch.cuda.is_available():
        torch.cuda.reset_peak_memory_stats()
        torch.cuda.empty_cache()

class TestModelCurrent(nn.Module):
    """Current approach: convert to fp32 AFTER tanh"""
    def __init__(self, vocab_size, n_embd):
        super().__init__()
        self.lm_head = nn.Linear(n_embd, vocab_size, bias=False)
        self.softcap = 15

    def forward(self, x):
        logits = self.lm_head(x)
        logits = self.softcap * torch.tanh(logits / self.softcap)  # tanh in bfloat16
        logits = logits.float()  # convert to fp32 AFTER tanh
        return logits

class TestModelProposed(nn.Module):
    """Proposed approach: convert to fp32 BEFORE tanh"""
    def __init__(self, vocab_size, n_embd):
        super().__init__()
        self.lm_head = nn.Linear(n_embd, vocab_size, bias=False)
        self.softcap = 15

    def forward(self, x):
        logits = self.lm_head(x)
        logits = logits.float()  # convert to fp32 BEFORE tanh
        logits = self.softcap * torch.tanh(logits / self.softcap)  # tanh in fp32
        return logits

def test_memory_usage(model, x, targets, compiled=False, num_iterations=10):
    """Test memory usage for a given model"""
    reset_memory()

    if compiled:
        model = torch.compile(model, dynamic=False)
        # Warmup for compilation
        for _ in range(3):
            _ = model(x)
            if targets is not None:
                _ = F.cross_entropy(_.view(-1, _.size(-1)), targets.view(-1))
        reset_memory()

    # Run forward passes
    for _ in range(num_iterations):
        logits = model(x)
        if targets is not None:
            loss = F.cross_entropy(logits.view(-1, logits.size(-1)), targets.view(-1))
            loss.backward()
            model.zero_grad()

    peak_memory_mb = torch.cuda.max_memory_allocated() / 1024 / 1024 if torch.cuda.is_available() else 0
    return peak_memory_mb

def main():
    if not torch.cuda.is_available():
        print("CUDA not available. This test requires a GPU.")
        return

    # Model config (similar to your GPT model)
    batch_size = 4
    seq_len = 1024
    vocab_size = 50304
    n_embd = 768

    # Create dummy input (in bfloat16 to match your model)
    x = torch.randn(batch_size, seq_len, n_embd, dtype=torch.bfloat16, device='cuda')
    targets = torch.randint(0, vocab_size, (batch_size, seq_len), device='cuda')

    print(f"Test configuration:")
    print(f"  Batch size: {batch_size}")
    print(f"  Sequence length: {seq_len}")
    print(f"  Vocab size: {vocab_size}")
    print(f"  Embedding dim: {n_embd}")
    print(f"  Logits tensor size: {batch_size * seq_len * vocab_size * 2 / 1024 / 1024:.2f} MB (bfloat16)")
    print(f"  Logits tensor size: {batch_size * seq_len * vocab_size * 4 / 1024 / 1024:.2f} MB (fp32)")
    print()

    # Test without compilation
    print("=" * 60)
    print("Testing WITHOUT torch.compile:")
    print("=" * 60)

    model_current = TestModelCurrent(vocab_size, n_embd).to('cuda').to(torch.bfloat16)
    model_proposed = TestModelProposed(vocab_size, n_embd).to('cuda').to(torch.bfloat16)

    mem_current = test_memory_usage(model_current, x, targets, compiled=False)
    mem_proposed = test_memory_usage(model_proposed, x, targets, compiled=False)

    print(f"Current (fp32 AFTER tanh):  {mem_current:.2f} MB")
    print(f"Proposed (fp32 BEFORE tanh): {mem_proposed:.2f} MB")
    print(f"Difference: {mem_proposed - mem_current:.2f} MB ({((mem_proposed / mem_current - 1) * 100):+.2f}%)")
    print()

    # Test with compilation
    print("=" * 60)
    print("Testing WITH torch.compile:")
    print("=" * 60)

    model_current = TestModelCurrent(vocab_size, n_embd).to('cuda').to(torch.bfloat16)
    model_proposed = TestModelProposed(vocab_size, n_embd).to('cuda').to(torch.bfloat16)

    mem_current_compiled = test_memory_usage(model_current, x, targets, compiled=True)
    mem_proposed_compiled = test_memory_usage(model_proposed, x, targets, compiled=True)

    print(f"Current (fp32 AFTER tanh):  {mem_current_compiled:.2f} MB")
    print(f"Proposed (fp32 BEFORE tanh): {mem_proposed_compiled:.2f} MB")
    print(f"Difference: {mem_proposed_compiled - mem_current_compiled:.2f} MB ({((mem_proposed_compiled / mem_current_compiled - 1) * 100):+.2f}%)")
    print()

    # Summary
    print("=" * 60)
    print("Summary:")
    print("=" * 60)
    print(f"Without compile - Memory increase: {mem_proposed - mem_current:.2f} MB")
    print(f"With compile - Memory increase: {mem_proposed_compiled - mem_current_compiled:.2f} MB")
    print()
    print("Note: torch.compile may optimize away intermediate tensors,")
    print("so the difference might be smaller than expected.")

if __name__ == "__main__":
    main()
```

output:

```
Test configuration:
  Batch size: 4
  Sequence length: 1024
  Vocab size: 50304
  Embedding dim: 768
  Logits tensor size: 393.00 MB (bfloat16)
  Logits tensor size: 786.00 MB (fp32)

============================================================
Testing WITHOUT torch.compile:
============================================================
Current (fp32 AFTER tanh):  3755.03 MB
Proposed (fp32 BEFORE tanh): 4148.03 MB
Difference: 393.00 MB (+10.47%)

============================================================
Testing WITH torch.compile:
============================================================
Current (fp32 AFTER tanh):  3755.72 MB
Proposed (fp32 BEFORE tanh): 3755.72 MB
Difference: 0.00 MB (+0.00%)

============================================================
Summary:
============================================================
Without compile - Memory increase: 393.00 MB
With compile - Memory increase: 0.00 MB

Note: torch.compile may optimize away intermediate tensors,
so the difference might be smaller than expected.
```

Confirming that torch compile is able to fuse this operation ok without any memory penalty (which should very well be possible).

I also examined the kernels generated using TORCH_LOGS env var and verified this is done nice and correctly and fused.

TLDR I'll merge this change thanks.

spjosyula:
That’s awesome, thanks for running the benchmarks to verify. Happy to contribute!

PR #327, opened 2025-11-23
Tok train

kiankyars:
Going through tok_train and there is redundancy in the creation of the token_bytes.pt


svlandeg (approved):
Fair enough, considering the definition of `token_strings` at L78, it looks like this can indeed be shortened.

kiankyars:
Alternatively, one doesn't define token strings outside of the for loop and instead creates the iterator in the for loop statement.

karpathy:
fair enough but not material enough to merge imo, keeping commit history clean

Issue #332, opened 2025-11-25
Add a few basic code linters (and a matching github action) via pre-commit


eyalf-nvidia:
Was thinking it would be nice if the project/repo had and enforced basic linting and formatting. Specifically, i like working with [pre-commit](https://pre-commit.com/). This could be added both locally and in CI via GitHub Actions.

For now, on my fork, i added these -
- [autoflake](https://github.com/PyCQA/autoflake): strips unused imports and variables.
- [pyupgrade](https://github.com/asottile/pyupgrade): rewrites code to the latest supported Python syntax.
- [isort](https://github.com/PyCQA/isort): keeps import blocks consistently ordered and grouped.
- [pre-commit-hooks](https://github.com/pre-commit/pre-commit-hooks): repo hygiene (trim trailing whitespace, enforce LF endings/newlines, detect merge conflicts, block oversized files).
- [black](https://github.com/psf/black): applies the canonical Python formatting.
- [codespell](https://github.com/codespell-project/codespell): catches common spelling mistakes in code and docs (add false positives to [tool.codespell].ignore-words-list in pyproject.toml).

If this is acceptable, i would be happy to contribute my implementation via PR (might already create a PR and wait for an approval, or not).

svlandeg:
Considering https://github.com/karpathy/nanochat/pull/333#issuecomment-3694441256, I'll go ahead and close this one 😉

PR #333, opened 2025-11-25
CI: add pre-commit workflow and dev tooling

eyalf-nvidia:
This PR implements/solves issue #332

### Changes

- Add `.pre-commit-config.yaml` with ruff-based linting and formatting hooks that run consistently locally and in CI
- Wire up the GitHub Action (Pre-commit) that installs the dev dependency group via `uv sync --group dev` and runs `uv run pre-commit run --all-files` on pushes, PRs, and manual dispatch
- Extend the dev dependency group in `pyproject.toml` with `pre-commit>=3.8.0` and configure ruff and codespell
- Normalize line endings in the touched files so the `mixed-line-ending` hook stays quiet
- Add pre-commit documentation to README.md

### Hook Coverage

- **[`ruff`](https://github.com/astral-sh/ruff)**: Fast Rust-based linter and formatter replacing multiple tools:
  - `ruff-check`: removes unused imports, upgrades syntax, sorts imports (replaces autoflake, pyupgrade, isort)
  - `ruff-format`: consistent code formatting with quote style preserved (replaces black)
- **[`pre-commit-hooks`](https://github.com/pre-commit/pre-commit-hooks)**: repo hygiene (trailing whitespace, LF endings, merge conflicts, large files)
- **[`codespell`](https://github.com/codespell-project/codespell)**: catches common spelling mistakes

### Usage (also in README.md)

- Make sure the dev extras are installed: `uv sync --group dev`
- Run the suite on demand: `uv run pre-commit run --all-files`
- (optional) Install git hook: `uv run pre-commit install`

**NOTE:** All `.py` file changes were auto-generated by the introduced linters/formatters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,22 @@ I haven't invested too much here but some tests exist, especially for the tokeni
 python -m pytest tests/test_rustbpe.py -v -s
 ```
 
+## Pre-commit hooks
+
+Linting and formatting are enforced with [pre-commit](https://pre-commit.com/) both locally and in CI via GitHub Actions. To match the checks that run in PRs:
+
+- Make sure the dev extras are installed (`uv sync --group dev`).
+- Run the suite on demand: `uv run pre-commit run --all-files`.
+- (optional) Install the git hook once (for automation during `git commit`): `uv run pre-commit install`.
+
+Hook coverage (auto-fixes most issues; review and stage the changes afterward):
+
+- [`ruff`](https://github.com/astral-sh/ruff): a fast Rust-based linter and formatter that replaces multiple tools:
+  - **Linting** (`ruff-check`): removes unused imports (like autoflake), upgrades syntax (like pyupgrade), and sorts imports (like isort).
+  - **Formatting** (`ruff-format`): applies consistent code formatting (like black), with quote style preserved.
+- [`pre-commit-hooks`](https://github.com/pre-commit/pre-commit-hooks): repo hygiene (trim trailing whitespace, enforce LF endings/newlines, detect merge conflicts, block oversized files).
+- [`codespell`](https://github.com/codespell-project/codespell): catches common spelling mistakes in code and docs (add false positives to `[tool.codespell].ignore-words-list` in `pyproject.toml`).
+
 ## File structure
 
 ```
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,44 @@
+name: Pre-commit
+
+on:
+  push:
+    branches:
+      - main
+      - "release/**"
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  run-pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+
+      - name: Cache uv & pre-commit
+        uses: actions/cache@v4
+        with:
+          path: |
+            .venv
+            ~/.cache/uv
+            ~/.cache/pre-commit
+          key: ${{ runner.os }}-uv-${{ hashFiles('uv.lock', '.pre-commit-config.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-uv-
+
+      - name: Install dev dependencies
+        run: uv sync --group dev
+
+      - name: Run pre-commit
+        run: uv run pre-commit run --all-files


eyalf-nvidia:
> I know Andrej is not a huge fan of `black`, so I'm not sure we'll want to merge this 😬

I personally learned to appreciate `black`, but I'm fine with any other formatting standard/tool as i believe its best to have some standard in place rather than not.

I would still love to "save" this PR and can offer any one of these -
1. Replace `black` with any other alternative (e.g. `ruff` formatter).
2. Remove the `black` hook from the `pre-commit` - as an initial `pre-commit`/`ci`, i think its still valuable.

@svlandeg, what do you say? Is it a lost cause?

svlandeg:
The output of the `ruff` formatter is pretty similar (if not virtually identical) to `black` though, right?

eyalf-nvidia:
> The output of the `ruff` formatter is pretty similar (if not virtually identical) to `black` though, right?

Yes, on any already-Black-formatted codebase the output is usually indistinguishable (~99.9 % identical). But if Andrej’s (and likely, also others’) real grudge with Black isn’t auto-formatting itself or even Black’s specific look but rather it’s **rigid “zero config, my way or the highway” philosophy** - then `ruff format` directly eases exactly that pain without throwing away the stability people like:

- **Black forces double quotes on everything** - Ruff lets you pick single quotes or preserve existing style  
- **Black locks you to 88 characters** - Ruff defaults to 88 but bumping to 100/120 is one line and extremely common  
- **Black gives you almost no knobs** - Ruff adds just the handful of options people begged for (quote style, indent style, docstring code formatting)  
- **Black can slow down on big repos** - Ruff is 30-100× faster (Rust)

In short, if the resentment is aimed at Black’s uncompromising dogma rather than the idea of an opinionated formatter, `ruff format` removes the dogma while keeping (and improving) everything else. That’s why a huge chunk of the former “I hate Black” crowd find `ruff format` reasonable.

svlandeg:
Right. As I understand it, it's mostly about the actual format that the code ends up being in, which is probably fixed (at least partly) by setting a (really) long EOL cut-off. But until we hear back from Andrej I can't give any guarantees whether such a change will result in the PR getting merged 😉 Though maybe it helps if there wouldn't be changes in 44+ files - in that sense maybe also preserving quotations helps, and then we can see from there what changes are still left. (normalizing spaces in front of a comment for instance seems harmless to me)

eyalf-nvidia:
> Right. As I understand it, it's mostly about the actual format that the code ends up being in, which is probably fixed (at least partly) by setting a (really) long EOL cut-off. But until we hear back from Andrej I can't give any guarantees whether such a change will result in the PR getting merged 😉 Though maybe it helps if there wouldn't be changes in 44+ files - in that sense maybe also preserving quotations helps, and then we can see from there what changes are still left. (normalizing spaces in front of a comment for instance seems harmless to me)

@svlandeg, totally fair, let’s wait for Andrej’s final word of course.

Just to set expectations if we go the Ruff route (happy to do the switch in this PR):

- I can configure `line-length = 120` (or whatever he prefers) and `quote-style = "preserve"` which will immediately eliminate **all** quote changes and most purely line-length-related wraps.
- Even with those two settings, I’d still expect ~30–40 of the current 44 files to be touched. Black’s internal formatting decisions go far beyond just quotes and line length: different parenthesis placement, argument splitting logic, f-string handling, blank lines around functions/classes, magic trailing commas, etc.

My main motivation for the PR is exactly what you hinted at: having **any** stable, automated, opinionated formatter (whether Black or Ruff) is vastly better for a public/collaborative project than no formatter at all, IMO. It removes endless nit-picking in reviews and keeps the codebase consistent for everyone.

Just let me know what is decided...

README.md
```diff
@@ -123,6 +123,23 @@ I haven't invested too much here but some tests exist, especially for the tokeni
...
+- (optional) Install the git hook once (for automation during `git commit`): `uv run pre-commit install`.
+
+Hook coverage (auto-fixes most issues; review and stage the changes afterward):
+
+- [`autoflake`](https://github.com/PyCQA/autoflake): strips unused imports and variables.
+- [`pyupgrade`](https://github.com/asottile/pyupgrade): rewrites code to the latest supported Python syntax.
+- [`isort`](https://github.com/PyCQA/isort): keeps import blocks consistently ordered and grouped.
+- [`pre-commit-hooks`](https://github.com/pre-commit/pre-commit-hooks): repo hygiene (trim trailing whitespace, enforce LF endings/newlines, detect merge conflicts, block oversized files).
+- [`black`](https://github.com/psf/black): applies the canonical Python formatting.
+- [`codespell`](https://github.com/codespell-project/codespell): catches common spelling mistakes in code and docs (add false positives to `[tool.codespell].ignore-words-list` in `pyproject.toml`).
```
IsaacGemal: This might be quite a lot for an OSS project.
Also, we don't need half of these to achieve the same result, at least with Ruff.

Per the [docs](https://docs.astral.sh/ruff/)
> Ruff can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [Black](https://github.com/psf/black), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [pyupgrade](https://pypi.org/project/pyupgrade/), [autoflake](https://pypi.org/project/autoflake/), and more,

The details on that might interest you.

eyalfrish: @IsaacGemal , thanks for the comment. I'm kinda used to using these set of hooks, but i have used and am 100% fine to transfer to alternative/equivalent hooks. 

As for this being "a lot" :). I agree that it seems like a long list, but these are mainly automatic formatter, meaning, all you need to do is to run `pre-commit run --all-files` and you are good to go. Moreover, i added a github workflow named "pre-commit" which should help not forget to process the hooks :). For those who REALLY want to make sure its never forgotten, they can even run `pre-commit install` which would install the `pre-commit` execution into git's pre-commit hook.

Anyway, i would love make changes (e.g. move to ruff), if i knew that this would increase the chances of this PR ever being approved. @svlandeg, is there anything proactive i can do to understand what/if i can do to make this PR worth considering? I saw its tagged `suggest/close` which i assume means it can be considered as closed, right?)

eyalfrish: @IsaacGemal @svlandeg , i understand its low chances for me to get this PR approved, but both of you made it clear for me that we would prefer ruff which both comes with built-in linters (allowing to remove most of the hooks i had originally) as well as being more configurable (less strict). Anyway, i re-implemented the hooks. It's cleaner and shorter now. I set line-length to 120 (rather than black's 88) and made it preserve quotes (though, i dont think this is a great decision).

BTW, as i mentioned, if this PR is approved, it also includes a GitHub workflow which is kinda nice/satisfying to have. It would looks like this -
image

I guess, all i have now is to cross fingers and wait :D.

IsaacGemal: Really my nitpick was that Karpathy wants this repo to be really minimal and hackable, so in my mind each new tool that is proposed to be added is a potential longterm overhead on the main goal of the project, even if it is very light.

> It is a single, cohesive, minimal, readable, hackable, maximally-forkable "strong baseline" codebase designed to run start to end and produce a concrete ChatGPT clone and its report card.

Except, at least personally for me, I think ruff with default settings for lint + formatting is the ideal balance, which was what I was suggesting. The less overhead, the more likely to be approved imo.

Also off topic - I think codespell is eventually going to fade in obscurity because of how much AIs can just cover that efficiently, with more nuance in understanding languages than static lists ever could.

karpathy:
Ou, my eyes!

karpathy:
Thank you for the PR but yes I can't stand black and its formatting, or the obsession with a few typos or whitespaces. I like PRs that make the code actually better. Like - trains 50% faster for the same quality of results. Let's go.

PR #316, opened 2025-11-20
Allow any rank to create the checkpoint_dir

marked23:
fixes  #315

Issue #315, opened 2025-11-20
In save_checkpoint, non-zero rank might try to use dir that doesn't yet exist.


marked23:
as-is:
```python
def save_checkpoint(checkpoint_dir, step, model_data, optimizer_data, meta_data, rank=0):
    if rank == 0:
        os.makedirs(checkpoint_dir, exist_ok=True)
```

suggested:
```python
def save_checkpoint(checkpoint_dir, step, model_data, optimizer_data, meta_data, rank=0):
    os.makedirs(checkpoint_dir, exist_ok=True)
    if rank == 0:
```

Logs:
```text
step 26085/26087 (99.99%) | loss: 2.712348 | grad norm: 0.1680 | lrm: 0.00 | dt: 5157.80ms | tok/sec: 83,384 | mfu: 4.91 | total time: 2243.30m
step 26086/26087 (100.00%) | loss: 2.724704 | grad norm: 0.1787 | lrm: 0.00 | dt: 5156.08ms | tok/sec: 83,412 | mfu: 4.91 | total time: 2243.38m
Step 26087 | Validation bpb: 0.8113
Evaluating: hellaswag_zeroshot (0-shot, type: multiple_choice)... accuracy: 0.4540 | centered: 0.2720 | time: 1.87s
Evaluating: jeopardy (10-shot, type: language_modeling)... accuracy: 0.1280 | centered: 0.1280 | time: 1.81s
Evaluating: bigbench_qa_wikidata (10-shot, type: language_modeling)... accuracy: 0.5280 | centered: 0.5280 | time: 1.80s
Evaluating: arc_easy (10-shot, type: multiple_choice)... accuracy: 0.6540 | centered: 0.5387 | time: 3.48s
Evaluating: arc_challenge (10-shot, type: multiple_choice)... accuracy: 0.3460 | centered: 0.1280 | time: 4.01s
Evaluating: copa (0-shot, type: multiple_choice)... accuracy: 0.6600 | centered: 0.3200 | time: 0.35s
Evaluating: commonsense_qa (10-shot, type: multiple_choice)... accuracy: 0.2840 | centered: 0.1050 | time: 4.63s
Evaluating: piqa (10-shot, type: multiple_choice)... accuracy: 0.6960 | centered: 0.3920 | time: 2.33s
Evaluating: openbook_qa (0-shot, type: multiple_choice)... accuracy: 0.3760 | centered: 0.1680 | time: 1.73s
[...]

PR #336, opened 2025-11-26
Fix race condition in save_checkpoint for non-zero ranks

zZzZ9zZ9:
When running `scripts/base_train.py` with multiple ranks, I hit the following error at the very end of training:

```text
RuntimeError: Parent directory /data/home/zizhuofu/.cache/nanochat/base_checkpoints/d20 does not exist.
  File "nanochat/checkpoint_manager.py", line 38, in save_checkpoint
    torch.save(optimizer_data, optimizer_path)
````

This happens because in `save_checkpoint` only `rank == 0` calls `os.makedirs(checkpoint_dir, exist_ok=True)`, while **all** ranks always call `torch.save` for their optimizer shard. If a non-zero rank reaches `torch.save` before rank 0 has created the directory, `torch.save` fails with the “Parent directory does not exist” error.

At the end of `base_train`, this can also corrupt the last checkpoint. Subsequent scripts that try to load that checkpoint (e.g. `scripts/base_loss.py`, `scripts/base_eval.py`, `scripts/mid_train.py`) then fail with:

```text
RuntimeError: PytorchStreamReader failed reading zip archive: failed finding central directory
```

## Fix

This PR removes the race condition by:

* Moving `os.makedirs(checkpoint_dir, exist_ok=True)` **outside** of the `rank == 0` branch so that every rank ensures the directory exists before saving.

This way, all ranks see a valid `checkpoint_dir`, and only rank 0 writes the shared model/metadata files, avoiding concurrent writes and checkpoint corruption.

### DIFF
--- a/nanochat/checkpoint_manager.py
+++ b/nanochat/checkpoint_manager.py
@@ -21,8 +21,8 @@ def log0(message):
         logger.info(message)
 
 def save_checkpoint(checkpoint_dir, step, model_data, optimizer_data, meta_data, rank=0):
+    os.makedirs(checkpoint_dir, exist_ok=True)
     if rank == 0:
-        os.makedirs(checkpoint_dir, exist_ok=True)
         # Save the model state parameters
         model_path = os.path.join(checkpoint_dir, f"model_{step:06d}.pt")
         torch.save(model_data, model_path)


marked23:
Hi.  This looks like a duplicate of #315, which also has a PR #316. 
I got bit by this too,   ;)

svlandeg:
Thanks for the note, @marked23! You're right, and the edits are exactly the same ones, so I'll go ahead and close this one.

Thanks for the detailed explanation though @zZzZ9zZ9 !

zZzZ9zZ9:
Thanks for pointing that out! I just noticed this overlaps exactly with #315 / #316 — appreciate the quick clarification. 😊

PR #342, opened 2025-11-27
Add append-mode logging option

kiankyars:
Add an `append` mode option to `Report.log()` to allow logs to accumulate instead of overwriting.


karpathy:
Doesn't super explain why you'd do this.

PR #345, opened 2025-11-28
Improve KV cache code readability

deepbuilder:
Cosmetic improvements to KVCache for better readability and explicit tensor operations.

**Changes:**

1. **Explicit tensor indexing in insert_kv():**
   - Add missing head dimension indexing: `t0:t1, :` instead of `t0:t1`
   - Makes tensor assignment semantically correct and eliminates reliance on broadcasting
   - Improves robustness for chunked processing scenarios

2. **Explicit dimension validation in prefill():**
   - Replace magic-index loop with explicit dimension unpacking
   - Use named variables for each dimension comparison
   - Provides clearer error messages and better code readability


karpathy:
yeah, you're right thanks

PR #348, opened 2025-12-01
Fix conversation scroll to bottom on some browsers

pkrefta:
Hi,

This pull request fixes a small issue related to the web interface in certain browsers. When a conversation became longer, the input bar for typing a new message would simply move downward, and after adding a new message to the conversation, you had to scroll the browser to reach it.

It also removes duplicated `padding` property for one of the CSS classes.

### DIFF
--- a/nanochat/ui.html
+++ b/nanochat/ui.html
@@ -14,6 +14,11 @@
             box-sizing: border-box;
         }
 
+        html, body{
+            height: 100%;
+            margin: 0;
+        }
+
         body {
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
Thanks @pkrefta, this is nice!

On my system/browser as well, on `master` the conversation doesn't "automatically scroll" and you need to manually scroll down after a long answer. With this PR, it scrolls automatically and the UX is so much nicer.

The duplicate `padding` property can indeed also be removed.

PR #350, opened 2025-12-02
Refactor vertex ai pipelines

javasoup:


javasoup:
I agree with your suggestion. Closing.

PR #358, opened 2025-12-06
Fix dataloader infinity loop end of corpus, no reset file counter

kowalikmarcel:
Dataloader goes into infinity loop when it reaches end of the corpus. There is no counter reset.


karpathy:
I think fixed this just now, please lmk if you run into more issues, closing.

Issue #360, opened 2025-12-08
Bug: tokenizing_distributed_data_loader_with_state never resets pq_idx and breaks multi-epoch training


SAMSUNYUJUN:
### Bug description

When training long enough to consume all Parquet files, the distributed dataloader `tokenizing_distributed_data_loader_with_state` eventually stops yielding batches and the training loop hangs.

This happens because in the inner `document_batches()` generator, `pq_idx` is initialized once before the outer `while True` loop and is never reset inside that loop. Once `pq_idx` reaches `len(parquet_paths)`, the inner `while pq_idx < len(parquet_paths):` loop never runs again, so no more `(doc_batch, (pq_idx, rg_idx))` are yielded even though the outer loop is `while True`.

So effectively the loader only supports a single pass over the Parquet files (plus approximate resume), but not true multi-pass / infinite streaming.

### Code location
The issue comes from the document_batches() function inside tokenizing_distributed_data_loader_with_state from dataloader.py

PR #361, opened 2025-12-08
Fix distributed Parquet dataloader resume for multi-epoch training

SAMSUNYUJUN:
### Summary

This PR fixes an issue in `tokenizing_distributed_data_loader_with_state`
where the internal `document_batches()` generator effectively only makes a
single pass over all Parquet files. After `pq_idx` reaches
`len(parquet_paths)`, the inner loop `while pq_idx < len(parquet_paths):`
never runs again, so the generator stops yielding batches even though the
outer loop is `while True`.

This can cause long training runs (or large `TARGET_PARAM_DATA_RATIO`)
to hang when the dataloader tries to call `next()` but no more batches
are produced.

### What this change does

- Treats `resume_pq_idx` / `resume_rg_idx` as a **one-time** approximate
  resume offset for the first pass after a restart.
- Moves the initialization of `pq_idx` inside the outer `while True`
  loop, so that:
  - On the first pass, we start from `resume_pq_idx` and finish that pass.
  - On later passes, we start from `0` and iterate over all Parquet files
    again (true multi-epoch / infinite streaming).
- Keeps the existing DDP row-group sharding (`rg_idx = ddp_rank`,
  `rg_idx += ddp_world_size`) and only uses `resume_rg_idx` when
  resuming on the same file for the first pass.

This preserves the current “approximate resume (may skip a few documents
but not repeat them)” behavior, while allowing the dataloader to loop
over the dataset multiple times for longer training runs.

Fixes (https://github.com/karpathy/nanochat/issues/360#issue-3704834205)

### DIFF
--- a/nanochat/dataloader.py
+++ b/nanochat/dataloader.py
@@ -29,17 +29,22 @@ def tokenizing_distributed_data_loader_with_state(B, T, split, tokenizer_threads
         parquet_paths = parquet_paths[:-1] if split == "train" else parquet_paths[-1:]
         resume_pq_idx = resume_state_dict["pq_idx"] if resume_state_dict is not None else 0
         resume_rg_idx = resume_state_dict["rg_idx"] if resume_state_dict is not None else None
+        first_pass = True
         pq_idx = resume_pq_idx # we kick off parquet files at the resume index (or by default just 0)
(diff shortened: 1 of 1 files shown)


karpathy:
Sigh, thanks.

PR #364, opened 2025-12-09
Pad vocab size to 64 for DDP optimizers and efficiency

Kripner:
If user decides to not train a tokenizer but instead use e.g. gpt-2's one, and `vocab_size` happens to not be divisible by `world_size` in DDP (e.g. with 50,257 tokens in gpt-2), the `DistAdamW` optimizer will fail with an obscure error. This PR:
- adds an assert to `DistAdamW` with a nicer error message
- defaults to padding `wte` & `lm_head` to size 64, which additionally has potential efficiency benefits on NVIDIA (see https://huggingface.co/docs/transformers/main_classes/model#transformers.PreTrainedModel.resize_token_embeddings)

Additionally, I added an assert to the dataloader which prevents infinite looping when there is some problem with parquet files download/discovery.


karpathy:
good, ty!

PR #375, opened 2025-12-15
HuggingFace wrapped up, eval realized via lm-eval

bill810975:


bill810975:
Accidental PR; closing. No action needed.

PR #387, opened 2025-12-21
bugfix save and load ckpt from model_tag dir

Kurt232:
* We can set the `model_tag` in scripts.base_train, but scripts.mid_train and scripts.chat_sft use default model tag `f"d{n_layer}"`.


svlandeg (commented):
Hi, thanks for the PR! Could you limit the edits on this PR to contain only what is supposed to be the bug fix(es), and omit the additional configuration stuff? It'll be easier to review & merge if it's focused.

Kurt232:
Done. However, I suggest converting the args reading style in `scripts/base_eval.py` into `scripts/base_train.py`.

Kurt232:
And, I also notice that `model_step` is used in `scripts/base_loss.py` while `step` is used in other files.

PR #390, opened 2025-12-26
Replace os.path with pathlib for clearer code

tsvikas:
## Replace os.path with pathlib for clearer code

This PR replaces `os.path` operations with `pathlib.Path` throughout the codebase.
This syntax tends to be more intuitive for most Python users, especially those newer to the language.

For example:

```python
# Before
checkpoint_dir = os.path.join(base_dir, "checkpoints", model_tag)
os.makedirs(checkpoint_dir, exist_ok=True)

# After
checkpoint_dir = base_dir / "checkpoints" / model_tag
checkpoint_dir.mkdir(parents=True, exist_ok=True)
```

```python
# Before
with open(section_file, "r", encoding="utf-8") as in_file:
    section = in_file.read()

# After
section = section_file.read_text(encoding="utf-8")
```

### Notes

- Zero functional changes - same behavior, clearer syntax
- `pathlib` has been in Python stdlib since 3.4 (2014), very stable

### DIFF
--- a/nanochat/checkpoint_manager.py
+++ b/nanochat/checkpoint_manager.py
@@ -5,7 +5,6 @@ import os
 import re
 import json
 import logging
-from pathlib import Path
 import torch
 
 from nanochat.common import get_base_dir


tsvikas:
I know this is a wide change touching many files, and I've seen that formatting PRs haven't been well-received in the past. However, I hope this one is different because:

1. It keeps the code clean and hackable, without adding a commit hook or changing formatting
2. Many Python users find this syntax more readable
3. `pathlib` is a stable, well-established pattern

I understand if you prefer to keep things as-is, but I thought it was worth proposing. Happy to discuss or close if you feel it's not worth the churn.

nanochat/checkpoint_manager.py
```diff
@@ -20,42 +20,42 @@ def log0(message):
     if int(os.environ.get('RANK', 0)) == 0:
         logger.info(message)
 
-def save_checkpoint(checkpoint_dir, step, model_data, optimizer_data, meta_data, rank=0):
+def save_checkpoint(checkpoint_dir: Path, step, model_data, optimizer_data, meta_data, rank=0):
```
svlandeg: We can debate whether or not we want these type annotations. If the repo switches to mainly using `pathlib.Path`, it would seem obvious to me that these will be `Path` instances, and we wouldn't need the explicit type annotation (which the other parameters don't have either) and we wouldn't even need the import at the top of this file.

nanochat/checkpoint_manager.py
```diff
@@ -110,27 +110,27 @@ def find_largest_model(checkpoints_dir):
...
 
-def find_last_step(checkpoint_dir):
+def find_last_step(checkpoint_dir: Path):
     # Look into checkpoint_dir and find model_<step>.pt with the highest step
-    checkpoint_files = glob.glob(os.path.join(checkpoint_dir, "model_*.pt"))
+    checkpoint_files = list(checkpoint_dir.glob("model_*.pt"))
     if not checkpoint_files:
         raise FileNotFoundError(f"No checkpoints found in {checkpoint_dir}")
-    last_step = int(max(os.path.basename(f).split("_")[-1].split(".")[0] for f in checkpoint_files))
+    last_step = int(max(f.name.split("_")[-1].split(".")[0] for f in checkpoint_files))
```
svlandeg: By using `stem` I believe we can drop the splitting on `"."` entirely:
```suggestion
    last_step = int(max(f.stem.split("_")[-1] for f in checkpoint_files))
```

nanochat/checkpoint_manager.py
```diff
@@ -148,5 +148,5 @@ def load_model(source, *args, **kwargs):
         "rl": "chatrl_checkpoints",
     }[source]
     base_dir = get_base_dir()
-    checkpoints_dir = os.path.join(base_dir, model_dir)
+    checkpoints_dir = Path(base_dir) / model_dir
```
svlandeg: At this point, `base_dir` should already be a `Path`
```suggestion
    checkpoints_dir = base_dir / model_dir
```

nanochat/common.py
```diff
@@ -47,35 +48,35 @@ def setup_default_logging():
...
+def download_file_with_lock(url, filename, postprocess_fn=None) -> Path:
     """
     Downloads a file from a URL to a local path in the base directory.
     Uses a lock file to prevent concurrent downloads among multiple ranks.
     """
     base_dir = get_base_dir()
-    file_path = os.path.join(base_dir, filename)
-    lock_path = file_path + ".lock"
+    file_path = base_dir / filename
+    lock_path = file_path.with_name(f"{file_path.name}.lock")
```
svlandeg, line 69: This is arguably the only location where the "Pathlib solution" is slightly more verbose than the "os solution". There's no better way to do this one though than using `with_name`, which is the correct way to do this.

nanochat/checkpoint_manager.py
```diff
@@ -20,42 +20,42 @@ def log0(message):
...
 
-def save_checkpoint(checkpoint_dir, step, model_data, optimizer_data, meta_data, rank=0):
+def save_checkpoint(checkpoint_dir: Path, step, model_data, optimizer_data, meta_data, rank=0):
     if rank == 0:
-        os.makedirs(checkpoint_dir, exist_ok=True)
+        checkpoint_dir.mkdir(parents=True, exist_ok=True)
         # Save the model state parameters
-        model_path = os.path.join(checkpoint_dir, f"model_{step:06d}.pt")
+        model_path = checkpoint_dir / f"model_{step:06d}.pt"
         torch.save(model_data, model_path)
```
svlandeg, line 27: I can confirm that `torch` works fine with `Path` objects and doesn't need the `str()` conversion here.

nanochat/checkpoint_manager.py
tsvikas: OK. I removed these type hints (both from function parameters and function return values) from the whole repo.
function return values are less obviously of type Path, but they are out of style with the rest of the repo.

nanochat/checkpoint_manager.py
tsvikas: Note that `.stem()` only removes the last suffix (e.g., `"file.1.pt" → "file.1"`) while the original code split the filename on the first `.`. However, since `save_checkpoint()` only creates `.pt` files, `.stem` might be fine.

Which do you prefer?

svlandeg: There shouldn't be more than one suffix, so it should be fine

tsvikas:
All feedback addressed. Thanks for the thorough review!

tsvikas:
and also - thank you for the kind words! 🌞

karpathy:
I don't like Path tbh. It creates a new object that kind of behaves like a string but not always and it's annoying to keep track of what is Path and what is str and it needs imports. It's really not that bad to just simply use str and a few functions that operate over them. Doesn't really improve too much at all.

PR #392, opened 2025-12-29
Add optional SHA256 checksum verification for FineWeb-Edu shards

### LINKED ISSUE #344: New script: add checksum verification for FineWeb-Edu shards
### Summary

The current `nanochat/dataset.py` downloader streams FineWeb-Edu shards from HuggingFace but does not verify file integrity. This means truncated or corrupted downloads can remain on disk unnoticed, especially if they come from mirrors or resumable downloads.

### Motivation

Training or evaluation runs may silently use bad data if there are network issues, partial files, or mirror-side problems. The dataset is large enough that redownloading everything is expensive. Many HuggingFace datasets already expose `sha256` in the Git LFS pointer files; it would be nice if `nanochat` made it easy to validate shards against those.

### Solution

I wrote a small helper script `scripts/verify_shards.py` that:

- fetches the Git LFS pointer for each `shard_XXXXX.parquet` from HuggingFace and parses the `sha256:` token
- computes the local `sha256` incrementally to avoid loading big files into memory
- lets the user select a range of shard indices and a local directory
- summarizes missing files, mismatches, and remote errors

Script for reference: [verify_shards.py](https://github.com/Coldmooon/nanochat/blob/master/scripts/verify_shards.py)

```
python scripts/verify_shards.py \
    --repo karpathy/fineweb-edu-100b-shuffle \
    --local-dir ~/.cache/nanochat/base_data \
    --start 0 --end 239
```

I think it could be helpful if integrating more direct and offer a PR.

Image

Vishal-Krishna-Kumar:
This PR- adds the SHA256 checksum verification when downloading
FineWeb-Edu dataset shards.

If a `checksums_sha256.txt` file is present, downloaded shards are
verified before being accepted, preventing any type of silent corruption.

The verification is optional and does not change existing behavior
when no checksums file is available.

Note: running the dataset module requires the torch dependency.

Fixes #344

### DIFF
--- a/nanochat/dataset.py
+++ b/nanochat/dataset.py
@@ -6,7 +6,7 @@ This file contains utilities for:
 
 For details of how the dataset was prepared, see `repackage_data_reference.py`.
 """
-
+import hashlib
 import os
 import argparse
 import time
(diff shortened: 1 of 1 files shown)


nanochat/dataset.py
```diff
@@ -85,6 +117,13 @@ def download_single_file(index):
                         f.write(chunk)
             # Move temp file to final location
             os.rename(temp_path, filepath)
+            #Optional CheckSum Verification
+            
+            checksums = load_checksums()
+            if checksums and filename in checksums:
+                verify_checksum(filepath, checksums[filename])
```
svlandeg, line 124: When the checksum doesn't match, the file should be removed instead of just raising an error at this point. Because with the current implementation, the file would just remain on disk and would not be checked again in a subsequent run (because then it exists and wouldn't be redownloaded / rechecked).

Probably the process should then (try to) continue instead of halting entirely with a raw exception.

PR #393, opened 2025-12-29
fix(install): add nvidia-nvshmem to GPU extra to prevent libnvshmem_host.so.3 error

### LINKED ISSUE #183: `libnvshmem_host.so.3` not found errors after commit `dd9387b`
## Environments
Two types of environments are used:
1. Lambda: lambda instances launched with the Lambda Stack, which comes with many preinstalled packages
2. Others: standard ubuntu linux machines with GPU drivers installed. These can be reproduced using lambda instances with the GPU base image selected

## Behavior
Before commit `dd9387b362ac2c72a345592458a668a753a1516a`, the code ran without errors on both lambda and other machines.

After that commit, the lambda stack still runs fine, but other environments would emit `libnvshmem_host.so.3` not found errors before training starts. 

## Workaround
I found that adding a line of `"nvidia-nvshmem-cu12>=3.3"` in 
```python
[project.optional-dependencies]  
cpu = [  
    "torch>=2.8.0",  
]  
gpu = [  
    "torch>=2.8.0",  
    "nvidia-nvshmem-cu12>=3.3"
]  
```
would fix the problem for other environments. But I am unsure if how robust this fix is across different environments. Thanks.

Vishal-Krishna-Kumar:
### What this PR does
Adds `nvidia-nvshmem-cu12` to the `gpu` optional dependency group to prevent runtime
errors such as: libnvshmem_host.so.3: cannot open shared object file

on non-Lambda Linux environments.

### Why this is needed

After commit dd9387b, some environments fail at runtime because NVSHMEM is no longer
installed implicitly when GPU extras are enabled.

Explicitly listing this dependency makes GPU installs more robust and prevents
silent runtime failures.

### Scope & Safety

- Only affects users installing with `--extra gpu`
- CPU-only installs are unchanged
- Matches behavior observed before dd9387b

### Related issue

Fixes #183


svlandeg (commented):
The fix proposed here doesn't seem ideal to me, as it would influence any `gpu` installation while it seems that for many of them, this is all still going fine right now and they don't need this explicit `nvidia-nvshmem-cu12` requirement.

What I'd love to understand is what exactly caused the issue reported in https://github.com/karpathy/nanochat/issues/183 to manifest after commit https://github.com/karpathy/nanochat/commit/dd9387b362ac2c72a345592458a668a753a1516a? I guess I just don't see it.

svlandeg:
As there has been no follow-up / further clarification around this issue & the proposed fix, I'll go ahead and close this for now.

PR #395, opened 2025-12-29
Small rewrite of base cache dir helper

helloaidank:
**Description:** Small cleanup in `get_base_dir()` in `common.py`

Simplify construction of nanochat_dir from three lines to a single os.path.join(...) expression.
Preserve existing behavior (still uses `~/.cache/nanochat `by default, or `NANOCHAT_BASE_DIR `when set).
No functional changes expected. This is just a minor readability refactor.


svlandeg (commented):
Hi! The matter of "readability" is subjective IMO, and I don't think this particular case matters much either way. I would just keep as is.

PR #396, opened 2025-12-29
Add high value engine tests for core invariants (33 LoC)

barisozmen:
## Problem

Engine lacks tests for core generation contracts. Refactoring or modifying the generation logic has no safety net for these invariants:
- Seed reproducibility
- Temperature=0 determinism
- max_tokens limit
- num_samples output count

## Solution

Add 4 minimal, interface-focused tests using the existing `MockModel`/`ByteTokenizer` fixtures:

| Test | Contract |
|------|----------|
| test_seed_reproducibility | Same seed produces identical output |
| test_temperature_zero_determinism | temperature=0 is deterministic regardless of seed |
| test_max_tokens_respected | Generation stops at max_tokens limit |
| test_num_samples_count | num_samples=N produces exactly N sequences |

They give confidence that refactoring or modifying the generation logic won’t break core invariants.

Added 33 lines of code in total (counted by non-comment non-blank lines).

## Verification
Ran test_engine.py with new tests. All passed.

```bash
$ PYTHONPATH=. uv run pytest tests/test_engine.py -v

====================== test session starts ======================
platform darwin -- Python 3.10.18, pytest-8.4.2, pluggy-1.6.0 -- ~/dev/github/barisozmen/nanochat/.venv/bin/python
cachedir: .pytest_cache
rootdir: ~/dev/github/barisozmen/nanochat
configfile: pyproject.toml
plugins: anyio-4.10.0
collected 6 items                                               

tests/test_engine.py::test_kv_cache_resize PASSED         [ 16%]
tests/test_engine.py::test_multi_sample_first_token_diversity PASSED [ 33%]
tests/test_engine.py::test_seed_reproducibility PASSED    [ 50%]
tests/test_engine.py::test_temperature_zero_determinism PASSED [ 66%]
tests/test_engine.py::test_max_tokens_respected PASSED    [ 83%]
tests/test_engine.py::test_num_samples_count PASSED       [100%]

======================= 6 passed in 2.00s =======================

```

## Testing
Just to make sure. I ran the full test suite after changes and pasted the output below:
```bash
❯ PYTHONPATH=. uv run pytest

====================== test session starts ======================
platform darwin -- Python 3.10.18, pytest-8.4.2, pluggy-1.6.0
rootdir: ~/dev/github/barisozmen/nanochat
configfile: pyproject.toml
testpaths: tests
plugins: anyio-4.10.0
collected 11 items                                              

tests/test_engine.py ......                               [ 54%]
tests/test_rustbpe.py .....                               [100%]

====================== 11 passed in 9.82s =======================
```

## Further

As you may notice, the test fixtures could be deduplicated. I kept them as-is to match the existing test file structure, but I’m happy to refactor if you prefer.

More tests can be added quickly using a similar process in  "Appendix: AI usage" section (by picking critical, high value, and non-brittle ones). While it would make the codebase a bit bigger, the trade-off can be favorable: they will provide peace of mind during refactoring and might save time by catching bugs early. New tests didn't introduce any noticeable test runtime overhead.

## Appendix: AI usage

I used Claude Code (Opus 4.5) in planning mode first with the following prompt:

```
checkout at a new branch and then write tests for there to add in @tests/test_engine.py

tests should be minimal and very to do the point. the main purpose is making developers confident that when they refactor or code change, the code is not broken and work correctly. 

always evaluate new tests based on its value and cost of maintenance and brittleness. don't write any brittle tests that depend on implementation rather than interface. tests should test the interface and expected behavior

additionally, add PR description markdown.

First, read my previous PRs and their descriptions to see the format and get ideas:
https://github.com/karpathy/nanochat/pull/394
https://github.com/karpathy/nanochat/pull/383

ultrathink
```

After Claude Code presented its plan and proposed a set of high-value tests, I selected those marked as “Critical” and “High.” It then added these tests to the codebase. I then went through all tests and extended them with for loops and more assertions to cover more cases.

### DIFF
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -253,8 +253,4 @@ def test_different_seeds_introduce_variation_when_temperature_nonzero():
         outputs.add(tuple(results[0]))
 
     # Sanity check: sampling actually introduces variation
-    assert len(outputs) > 1, (
-        f"All seeds produced the same output: {outputs}"
-        f"with temperature > 0 and different seeds, this is statistically impossible."
-        f"this implies an issue within the engine."
-    )
+    assert len(outputs) > 1, "All seeds produced the same output which is statistically highly improbable."


tests/test_engine.py
```diff
@@ -185,3 +185,52 @@ def test_multi_sample_first_token_diversity():
...
+    """Temperature=0 is deterministic regardless of seed."""
+    model = MockModel()
+    engine = Engine(model, ByteTokenizer())
+    prompt = [261, 72, 101, 108, 108, 111]
+
+    for seed in [1, 42, 123, 999]:
+        r1, _ = engine.generate_batch(prompt, temperature=0.0, max_tokens=5, seed=seed)
+        r2, _ = engine.generate_batch(prompt, temperature=0.0, max_tokens=5, seed=seed)
+        r3, _ = engine.generate_batch(prompt, temperature=0.0, max_tokens=5, seed=seed)
+        assert r1 == r2 == r3, "Temperature=0 must result in the same output for the same prompt regardless of seed."
```
svlandeg: This doesn't test what it says it does: it's still keeping the seed for each comparison fixed. Instead, this should be something like:

```suggestion
    r1, _ = engine.generate_batch(prompt, temperature=0.0, max_tokens=5, seed=1)
    r2, _ = engine.generate_batch(prompt, temperature=0.0, max_tokens=5, seed=42)
    r3, _ = engine.generate_batch(prompt, temperature=0.0, max_tokens=5, seed=123)
    assert r1 == r2 == r3, "Temperature=0 must result in the same output for the same prompt regardless of seed."
```

svlandeg (commented):
Thanks! I think it makes sense to add some additional tests for the engine. One test was implemented wrongly and I've gone ahead and corrected it.

@karpathy: if we want this, maybe let's also merge https://github.com/karpathy/nanochat/pull/151 (I think it requires enabling GH actions on the repo first).

svlandeg (commented):
For completeness, we could also add a test that double checks that the results are different with different seeds 🤷‍♀️

tests/test_engine.py
```diff
@@ -206,11 +206,10 @@ def test_temperature_zero_determinism():
...
 
-    for seed in [1, 42, 123, 999]:
-        r1, _ = engine.generate_batch(prompt, temperature=0.0, max_tokens=5, seed=seed)
-        r2, _ = engine.generate_batch(prompt, temperature=0.0, max_tokens=5, seed=seed)
-        r3, _ = engine.generate_batch(prompt, temperature=0.0, max_tokens=5, seed=seed)
-        assert r1 == r2 == r3, "Temperature=0 must result in the same output for the same prompt regardless of seed."
+    r1, _ = engine.generate_batch(prompt, temperature=0.0, max_tokens=5, seed=1)
+    r2, _ = engine.generate_batch(prompt, temperature=0.0, max_tokens=5, seed=42)
+    r3, _ = engine.generate_batch(prompt, temperature=0.0, max_tokens=5, seed=123)
+    assert r1 == r2 == r3, "Temperature=0 must result in the same output for the same prompt regardless of seed."
```
barisozmen: thanks for correcting! I missed that

barisozmen:
> For completeness, we could also add a test that double checks that the results are different with different seeds 🤷‍♀️

I just added :) 

It's `test_different_seeds_introduce_variation_when_temperature_nonzero()`. It has a similar structure with the previous test `test_multi_sample_first_token_diversity()`. They both rely on extreme statistical unlikeliness of producing same output multiple times.

tests/test_engine.py
```diff
@@ -185,3 +185,76 @@ def test_multi_sample_first_token_diversity():
...
+            seed=seed,
+        )
+        outputs.add(tuple(results[0]))
+
+    # Sanity check: sampling actually introduces variation
+    assert len(outputs) > 1, (
+        f"All seeds produced the same output: {outputs}"
+        f"with temperature > 0 and different seeds, this is statistically impossible."
+        f"this implies an issue within the engine."
+    )
```
svlandeg: Let's just make the assert message shorter, this isn't user-facing anyway.

```suggestion
    assert len(outputs) > 1, "All seeds produced the same output which is statistically highly improbable."
```

PR #405, opened 2026-01-02
Batched generation from multiple prompts (with variable lengths)

Kripner:
Adds support for **batched generation from multiple prompts** (possibly of varying lengths).
Use cases:
- faster evals
- efficiently hosting a model for multiple users
- downstream applications that use the model in an outer loop such as tree-search (example: AlphaProof)

### Changes

`engine.py`:
- `Engine.generate()` now accepts either `list[int]` (single prompt) or `list[list[int]]` (batched prompts)
- Prompts are left-padded to the longest length, preserving RoPE relative positions. Token 0 is used for padding (doesn't matter, as it is masked)
- Attention mask prevents attending to pad tokens during both prefill and decode
- Prefill KV cache is replicated for `num_samples` per prompt
- Yields nested `list[list[int]]` of shape `(num_prompts, num_samples)` for batched input (and flat `list[int]` for single-prompt input)
- `generate_batch()` mirrors the changes: flattens internally, reshapes outputs back to (num_prompts, num_samples, seq_len) on return for batched input

`gpt.py`:
- Added optional `attention_mask` parameter to `GPT.forward()`, `Block.forward()`, and `CausalSelfAttention.forward()`
- When provided, the mask is passed directly to `F.scaled_dot_product_attention`

### Tests

Added tests to `test_engine.py`:
- `test_batched_generation_consistency` - verifies individual vs batched results match (with both randomly-initialized and pretrained models)
- `test_batched_generation_single_prompt` - single prompt in batched mode matches non-batched
- `test_batched_generation_stochastic` - verifies diverse outputs for temperature > 0

### DIFF
--- a/nanochat/engine.py
+++ b/nanochat/engine.py
@@ -359,7 +359,7 @@ class Engine:
                 )
             else:
                 logits = self.model.forward(ids, kv_cache=kv_cache_decode)
-            logits = logits[:, -1, :]
+            logits = logits[:, -1, :]  # (B, vocab_size)
 
     def generate_batch(self, tokens, num_samples=1, **kwargs):
         """


dustinwloring1988:
Did this make training faster since you are able to batch the benchmarks?

Kripner:
> It feels to me like this complicates the repo quite a bit, with the original aim being that `nanochat` remains quite minimal in itself. Especially with the input formats now varying between `list[int]` and `list[list[int]]`, I feel like the educational simplicity of the repo is getting a bit lost. Will leave the final decision on this with Andrej ofc.

It's easy to change it to always return `list[int]` (that actually simplifies the code a bit) - I just thought it would be more intuitive if the return shape mirrored the shape of input `tokens`.

I added batch generation because I needed it in my own project (nanoproof), and I suspect it might be useful elsewhere where the LLM is integrated in some larger pipeline. In nanochat, maybe a good use case would be using it to speed up `core_eval.py` (as @dustinwloring1988 pointed out). I will try that next weekend.

Kripner:
The only significant change is in `Engine.generate` (`gpt.py` just optionally forwards an attention mask to `torch.SDPA`). Another option is to keep `Engine.generate` as simple as possible and add `Engine.generate_batched` (and then probably rename `Engine.generate_batch`).

PR #412, opened 2026-01-04
fix typo in scripts/chat_rl.py

why2011btv:
typo in comments: change "GAPO" to "DAPO"

### DIFF
--- a/scripts/chat_rl.py
+++ b/scripts/chat_rl.py
@@ -6,7 +6,7 @@ simpler and more similar to just REINFORCE:
 
 1) Delete trust region, so there is no KL regularization to a reference model
 2) We are on policy, so there's no need for PPO ratio+clip.
-3) We use GAPO style normalization that is token-level, not sequence-level.
+3) We use DAPO style normalization that is token-level, not sequence-level.
 4) Instead of z-score normalization (r - mu)/sigma, only use (r - mu) as the advantage.
 
 1 GPU:


svlandeg (commented):
I don't think this is a typo, see also https://github.com/karpathy/nanochat/discussions/1 for more details.

why2011btv:
> I don't think this is a typo, see also #1 for more details.

Thanks for reviewing this Sofie! But I’m fairly certain "GAPO" is a typo for "DAPO", and I'd say the "GAPO" in #1 is a typo as well.

While there are several recent papers titled "GAPO," they generally focus on adversarial training ([ACL 2025](https://aclanthology.org/2025.acl-long.13.pdf), 2 citations for now) or group-aware optimization ([arXiv:2511.12596](https://arxiv.org/pdf/2511.12596), 1 citation for now) and do not address token-level normalization.

In contrast, DAPO ([arXiv:2503.14476](https://arxiv.org/pdf/2503.14476), 810 citations for now) **pioneered the use of token-level normalization** instead of sequence-level (see Section 3.3 of the paper). Considering DAPO's widespread adoption (800+ citations) and its integration into major frameworks like [verl](https://github.com/volcengine/verl), it aligns perfectly with the logic implemented here.

svlandeg:
Hi! Thanks for providing a reference paper. I was maybe too quick to assume that Andrej wouldn't make the same typo twice, and assumed GAPO was some method/variation I hadn't heard of yet. Either way 🥲  Will leave up to Andrej to make the final call, ofc :-)

PR #414, opened 2026-01-05
Update error message for invalid message role

why2011btv:
role only allows "user" / "assistant", not "system"


karpathy:
The tokenizer can actually render system, there's a chunk of code for it:

```python
        # sometimes the first message is a system message...
        # => just merge it with the second (user) message
        if conversation["messages"][0]["role"] == "system":
            # some conversation surgery is necessary here for now...
            conversation = copy.deepcopy(conversation) # avoid mutating the original
            messages = conversation["messages"]
            assert messages[1]["role"] == "user", "System message must be followed by a user message"
            messages[1]["content"] = messages[0]["content"] + "\n\n" + messages[1]["content"]
            messages = messages[1:]
        else:
            messages = conversation["messages"]
        assert len(messages) >= 1, f"Conversation has less than 1 message: {messages}"
```

svlandeg:
@karpathy:

This particular code path (where the edit of this PR is) is guarded by an `if message.role not in ["user", "assistant"]` though, so it doesn't look like `"system"` actually occurs at this point, because otherwise we would have seen crashes.

IIRC the "system" messages are being merged/overwritten? Right so in the code snippet you quote, the system message at index 0 is merged with the user message at index 1, and the role set at "user"? Or am I reading this wrong 🤔 

Either way, we should update either the message or the `if` statement, to be consistent.

PR #417, opened 2026-01-06
Fix undefined variable

adriablancafort:
I noticed the variable `device_batch_size` was undefined and corrected it to `args.device_batch_size`


svlandeg (approved):
Thanks! This is indeed a bug, a leftover from a recent refactor. 

I also confirm that the import is indeed unused and can be removed.

PR #425, opened 2026-01-09
Fix repackage_data_reference.py no output for small data

lenkog:
This fixes a bug in the example code for data repackaging where the last shard is not written if it isn't full.
If the dataset is small (e.g., specialized dataset) and doesn't fill up even one shard, no shards will be output at all.

The fix consists of forcing the current shard to be written out once the last document is processed.

### DIFF
--- a/dev/repackage_data_reference.py
+++ b/dev/repackage_data_reference.py
@@ -46,13 +46,14 @@ shard_characters = 0
 total_docs_processed = 0
 total_time_spent = 0
 t0 = time.time()
-for doc in ds:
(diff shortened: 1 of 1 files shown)


lenkog:
I used the script on a small dataset and nothing came out. I had to troubleshoot.
This PR implements a fix that could save someone some time.
If it doesn't matter much either way, I'm curious why your suggestion is *not* to take the fix?

svlandeg:
Because it might be intentional that you never get 'incomplete' shards (and instead you should change the parameters of the script). 

Either way, leaving the decision with Andrej.

PR #428, opened 2026-01-09
Add type annotations across the codebase for readability and maintainability

pkrefta:
This pull request adds Python type annotations to functions and methods across the nanochat codebase. Since nanochat is intentionally educational, I believe stronger typing meaningfully improves readability, navigation, and confidence when studying or modifying the code 😃 

I believe type hints will help readers understand data flow and expected shapes/structures faster and catch mistakes earlier via editors / static checkers etc.

The original code already contained a small number of annotations; I kept those mostly intact and expanded from there in a consistent style.

These annotations are not perfect and 100%-complete, but I did my best to balance correctness, clarity, and code size.

**Deliberate choices (I'm open for discussions)**
- Modern unions `int | None` instead of `Optional[int]` to avoid extra imports and keep syntax current.
- Built-in generics `list, dict, tuple` instead of `List, Dict, Tuple` again to avoid additional imports and keep the code compact
- I did not introduce helper types like TypeAlias, NamedTuple, OrderedDict, etc. This means some return types can be long, e.g. ```def get_example(self, index: int) -> dict[str, list[dict[str, str]] | str | list[str]]:``` That verbosity is a conscious trade-off to keep the repository lightweight and avoid adding too much “type infrastructure”.
- Where type declarations became lengthy, I split them across multiple lines to keep the code readable.

**Testing**

I ran the full speedrun.sh pipeline to verify everything still works as expected and to ensure there are no syntax errors introduced by the typing changes.

**LLM disclosure**

All annotations were added manually while I was reading and learning the code (I’m a senior Python engineer stepping into the AI/ML world). LLMs were helpful for explaining some trickier parts—especially torch-related pieces—and suggested types in a few places, but the majority (~80-90%) of the work and final decisions were mine.

### DIFF
--- a/scripts/chat_eval.py
+++ b/scripts/chat_eval.py
@@ -31,9 +31,16 @@ from tasks.spellingbee import SpellingBee
 # -----------------------------------------------------------------------------
 # Generative evaluation loop (we go one problem at a time, sample, evaluate)
 
-def run_generative_eval(task_object: Task, tokenizer: RustBPETokenizer, model: GPT, engine: Engine, num_samples: int,
-                        max_new_tokens: int, temperature: float, top_k: int, max_problems: int | None = None):
-
+def run_generative_eval(
+    task_object: Task,
+    tokenizer: RustBPETokenizer,
+    model: GPT, engine: Engine,
+    num_samples: int,
+    max_new_tokens: int,
+    temperature: float,
+    top_k: int,
+    max_problems: int | None = None
+) -> float:
     ddp, ddp_rank, ddp_local_rank, ddp_world_size = get_dist_info()
     device = model.get_device()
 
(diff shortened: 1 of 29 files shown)


scripts/chat_eval.py
```diff
@@ -4,8 +4,8 @@
 code lives in nanochat directory and is imported from here.
 
 Example runs:
-python -m scripts.chat_eval -a ARC-Easy
-torchrun --nproc_per_node=8 -m scripts.chat_eval -- -a ARC-Easy
+python -m scripts.chat_eval -i mid -a ARC-Easy
+torchrun --nproc_per_node=8 -m scripts.chat_eval -- -i mid -a ARC-Easy
```
svlandeg, line 8: `-i` is indeed a required argument for the `chat_eval` script, so whatever happens to this PR we should make sure to merge this edit (perhaps as a separate PR)

svlandeg: added this to https://github.com/karpathy/nanochat/pull/438

pkrefta:
Hi there!
Even though my PR was closed, I hope you don't mind me sharing a few thoughts.

Thanks @svlandeg , for taking the time to review my proposals 😃 

I totally agree that some of the annotations are quite lengthy - those were conscious choices on my part, but I'm definitely open to changes. We could always simplify things like changing `batch_size: int | None = None` to `batch_size: int = None`. It would be shorter - `mypy` might not be thrilled, but type annotations are often about finding the right balance. 
Similarly, we could change `torch.Tensor` to just `Tensor`, since there probably won't be other Tensor types in the code. If you'd like, I'd be happy to update my code to shorten things where needed. But if you feel that typing isn't necessary for this repository, that's totally fine too - I learned a lot just from adding these annotations 😃

svlandeg:
Hi! Really do appreciate the enthousiasm & effort 🙏 

If Andrej doesn't signal that he'd want some of these in shortened form, I would refrain from putting more effort into this 😉

PR #429, opened 2026-01-09
Download all required parquet shards before training the tokenizer

aioaneid:
In order for the tokenizer to be trained reproducibly, it needs to have all the required parquet shards available up front.

Besides the "train" shards, it also needs one extra "val" shard.

For completeness, I used the attached script to compute the number of shards to download:
[count_shards.py](https://github.com/user-attachments/files/24536336/count_shards.py)

For each modified script I verified that:
* One shard fewer is insufficient: 5 != 6, 10 != 11, 20 != 21
* One shard more is unnecessary: 6 = 7, 11 = 12, 21 = 22

Tokenizer training:
```
for i in 5 6 7 10 11 12 20 21 22; do
  d=.cache/nanochat.$(printf '%02d' "$i")
  mkdir -p "$d/base_data"

  for ((j=0; j<i; ++j)); do
    ln -s ~/.cache/nanochat/base_data/shard_$(printf '%05d' "$j").parquet \
          "$d/base_data/" 2>/dev/null
  done

  NANOCHAT_BASE_DIR="$d" \
    uv run -m scripts.tok_train --max_chars=$(( (i / 5) * 1000000000 ))
done

crc32 $(find .cache -name token_bytes.pt | sort)

echo

crc32 $(find .cache -name tokenizer.pkl | sort)
```

Final output:
```
05217488	.cache/nanochat.05/tokenizer/token_bytes.pt
247c4cd3	.cache/nanochat.06/tokenizer/token_bytes.pt
247c4cd3	.cache/nanochat.07/tokenizer/token_bytes.pt
f563ce74	.cache/nanochat.10/tokenizer/token_bytes.pt
7e580144	.cache/nanochat.11/tokenizer/token_bytes.pt
7e580144	.cache/nanochat.12/tokenizer/token_bytes.pt
5a9423db	.cache/nanochat.20/tokenizer/token_bytes.pt
b103bb1d	.cache/nanochat.21/tokenizer/token_bytes.pt
b103bb1d	.cache/nanochat.22/tokenizer/token_bytes.pt

d7589564	.cache/nanochat.05/tokenizer/tokenizer.pkl
4ba2aa8b	.cache/nanochat.06/tokenizer/tokenizer.pkl
4ba2aa8b	.cache/nanochat.07/tokenizer/tokenizer.pkl
38d4004c	.cache/nanochat.10/tokenizer/tokenizer.pkl
aa63c88b	.cache/nanochat.11/tokenizer/tokenizer.pkl
aa63c88b	.cache/nanochat.12/tokenizer/tokenizer.pkl
25345574	.cache/nanochat.20/tokenizer/tokenizer.pkl
a73bfcb3	.cache/nanochat.21/tokenizer/tokenizer.pkl
a73bfcb3	.cache/nanochat.22/tokenizer/tokenizer.pkl
```


svlandeg (commented):
Thanks for the PR!

So let me try to recap to see whether I understand the edits & reasoning here.

Basically the `speedrun.sh` scripts claims to train the tokenizer on ~2B characters. Originally, these numbers were divided by 250M as that's roughly the amount of chars in one data shard. But the `doc_cap` parameter in `tok_train.py` is set to 10K per document (by default), resulting in rougly 200M chars per shard instead (as can be seen with your script, cf appendix 👇 ). 

So to determine the number of shards we want, we should divide by 200M and not 250M. Makes sense to me, though results will ofc be different when users set `doc_cap` specifically.

Thinking further - doesn't this then also influence the calculation of required number of shards for pretraining? Right now, 240+8 shards are downloaded because 54MB / 250M = 216, but if we divide by 200M we'd have 270.

And then the same reasoning for the other two shell scripts `runcpu.sh` and `run1000.sh`, and also for the new `miniseries.sh` which currently downloads 240 shards as well.

## Appendix

Running your script with `doc_cap` set to 10K (default):

> Shard 0: 199493307 chars, cumulative: 199,493,307
Shard 1: 200952345 chars, cumulative: 400,445,652
Shard 2: 200448510 chars, cumulative: 600,894,162
Shard 3: 199391055 chars, cumulative: 800,285,217
Shard 4: 203048922 chars, cumulative: 1,003,334,139

--> ~200MB per shard


Running your script with `doc_cap` set to 100K (custom):

> Shard 0: 249787372 chars, cumulative: 249,787,372
Shard 1: 250358918 chars, cumulative: 500,146,290
Shard 2: 248320382 chars, cumulative: 748,466,672
Shard 3: 248726915 chars, cumulative: 997,193,587
Shard 4: 251334496 chars, cumulative: 1,248,528,083

--> ~250MB per shard

aioaneid:
It does make sense indeed to also correct the number of shards for pre-training, so I added a commit for that purpose. That could also be done separately though, focussing here only on making the tokenizer training more deterministic by making sure it has access to all the shards it needs. Please let me know what you prefer.

svlandeg:
Sorry to confuse things, actually it doesn't make sense to update the numbers for pretraining as the 250MB is still valid there, because `doc_cap` is only defined in `tok_train.py` (I had already crossed out those sentences but clearly too late...)

aioaneid:
I see. Removed the second commit.

PR #432, opened 2026-01-10
save checkpoint on ctrl+c during pretraining

suspicious-pineapple:
perhaps breaking the spirit of a minimal implementation, but it might save some unlucky fellow several hours of otherwise wasted training time someday (such as myself, yesterday)

todo: make it cleaner somehow, not putting a tiny QOL thing in the beginning of the file, and dont duplicate code

### DIFF
--- a/scripts/base_train.py
+++ b/scripts/base_train.py
@@ -12,6 +12,7 @@ python -m scripts.base_train --depth=4 --max_seq_len=512 --device_batch_size=1 -
 """
 
 import os
+import signal
 os.environ["PYTORCH_ALLOC_CONF"] = "expandable_segments:True"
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
If the loop is stopped somewhere halfway, how can we know we're saving a checkpoint in a consistent state?

suspicious-pineapple:
hmm fair
i suppose one could add it as an exit condition here?
image
issue would be if the loop itself was hanging, it wouldnt finish..

svlandeg:
As this PR has become a bit outdated and doesn't seem finished/polished, I'll go ahead and close it for now 🙏

PR #436, opened 2026-01-14
fix: Restore CPU/MPS support by adding FA3 fallback

hasso5703:
This PR addresses a regression introduced with the Flash Attention 3 integration. Running the model on non-CUDA devices (specifically macOS with MPS or CPU only) currently causes a crash because flash_attn is unavailable or None.

Changes:

- Added a conditional check in CausalSelfAttention.forward to fallback to F.scaled_dot_product_attention when flash_attn is not loaded.
- Implemented a manual KV-cache update loop in the inference path to handle the lack of flash_attn_with_kvcache on MPS/CPU.
- Added explicit dtype casting (.to(dtype=q.dtype)) during the manual cache update to resolve RuntimeError: Expected query, key, and value to have the same dtype (MPS strict typing issue between float32 query and bfloat16 cache).

Testing:

Verified on MacBook Air (Apple Silicon/MPS).
Confirmed that training runs successfully (loss decreases) via dev/runcpu.sh.
Confirmed that inference (text generation) runs without crashing after training.

### DIFF
--- a/nanochat/gpt.py
+++ b/nanochat/gpt.py
@@ -35,8 +35,12 @@ try:
     # Flash Attention 3 uses NVIDIA Hopper-specific features like TMA (Tensor Memory Accelerator).
     # These are only physically available on GPUs with Compute Capability >= 9.0 (e.g. H100).
     # We explicitly check for this to prevent "No kernel image available" crashes on Ampere/Ada GPUs (RTX 30xx/40xx) etc.
-    if torch.cuda.is_available() and torch.cuda.get_device_capability()[0] >= 9:
-        flash_attn = get_kernel('varunneal/flash-attention-3').flash_attn_interface
+    if torch.cuda.is_available():
+        if torch.cuda.get_device_capability()[0] >= 9:
+            flash_attn = get_kernel('varunneal/flash-attention-3').flash_attn_interface
+        else:
+            # If the kernel image is not available, try installing the wheel manually from https://windreamer.github.io/flash-attention3-wheels/
+            import flash_attn_interface as flash_attn
 except Exception:
     # Fallback to PyTorch SDPA on non-Hopper NVIDIA GPUs, Mac (MPS), or CPU.
     pass


nanochat/gpt.py
```diff
@@ -28,8 +28,8 @@
 os.environ["HF_HUB_DISABLE_PROGRESS_BARS"] = "1"
 # Official docs of FA3 label it as "beta" and want you to install FA3 from source, which is a pain.
 # Wishing for official FA3 wheels soon, for now this seems to be a fast way to get them (ty varunneal)
-from kernels import get_kernel
-flash_attn = get_kernel('varunneal/flash-attention-3').flash_attn_interface
+
+flash_attn = None
```
svlandeg, line 33: This simply removes FA3 for all setups? Instead, we could do something like a `try catch`.

hasso5703: You are absolutely right, thanks for catching this. Disabling it globally was an oversight on my part. I will wrap the kernel loading in a try-except block to preserve FA3 for supported systems while allowing fallback on others. Pushing the fix momentarily.

hasso5703:
Hardware Verification
I have personally verified this implementation across all three major backends to ensure no regression and correct fallback behavior:

- CUDA (Consumer): NVIDIA RTX 4070 Ti Super (Ada Lovelace, SM 8.9).
  - Result: Correctly detects `capability < 9.0`, skips FA3 loading, and falls back to SDPA. Training runs stable.
- MPS (Apple Silicon): Apple M4 Chip (macOS).
  - Result: Correctly handles the lack of CUDA, falls back to SDPA with manual cache updates.
- CPU: Intel i5 14600K.
  - Result: The explicit `(B, H, T, D)` transposition fixes the dimension mismatch crash during inference.

karpathy:
Thank you I was going to write basically the same fallback. I'll double check it tomorrow and plan to merge.

nanochat/gpt.py
```diff
@@ -29,7 +29,17 @@
...
 from kernels import get_kernel
-flash_attn = get_kernel('varunneal/flash-attention-3').flash_attn_interface
+
+flash_attn = None
+try:
+    # Flash Attention 3 uses NVIDIA Hopper-specific features like TMA (Tensor Memory Accelerator).
+    # These are only physically available on GPUs with Compute Capability >= 9.0 (e.g. H100).
+    # We explicitly check for this to prevent "No kernel image available" crashes on Ampere/Ada GPUs (RTX 30xx/40xx) etc.
+    if torch.cuda.is_available() and torch.cuda.get_device_capability()[0] >= 9:
+        flash_attn = get_kernel('varunneal/flash-attention-3').flash_attn_interface
```
svlandeg: 30xx/40xx series [could still benefit](https://github.com/Dao-AILab/flash-attention/issues/1049#issuecomment-2695283567) from FA3 if they install a prebuilt wheel from e.g.  https://windreamer.github.io/flash-attention3-wheels/, and then `import flash_attn_interface as flash_attn`. Tested on my 3090 (cuda 12.8) and works with this additional `else`.

```suggestion
    if torch.cuda.is_available():
        if torch.cuda.get_device_capability()[0] >= 9:
            flash_attn = get_kernel('varunneal/flash-attention-3').flash_attn_interface
        else:
            # If the kernel image is not available, try installing the wheel manually from https://windreamer.github.io/flash-attention3-wheels/
            import flash_attn_interface as flash_attn
```

karpathy:
I pushed my own version here: 

https://github.com/karpathy/nanochat/commit/8203efa9190e9d8da7419de29599e91a9242b968
https://github.com/karpathy/nanochat/commit/b62a5bc44aafef02eba6c39236180aa424a82674

(You have to be a lot more careful with the KV Cache handling and the sliding window handling)

Closing.

PR #437, opened 2026-01-14
When evaluating language_modeling tasks, ignore the case of the answer

askerlee:
I noticed on some tasks, the model I evaluated outputs a different case from the ground truth answer. For example, for many questions in bigbench_cs_algorithms, the model answers "valid" instead of "Valid" which is the ground truth and is regarded as being wrong. In this case, I guess it's fair to ignore the case. 

The change is a simple one: just decode the predicted tokens and the actual (ground truth) tokens, and match their lowercase forms.

(There are 4 commits shown below, but 3 of them are on a different branch. Only the last one is for this pull request.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ eval_bundle/
 .claude
 CLAUDE.md
 wandb/
+*.egg-info/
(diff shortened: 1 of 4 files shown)


svlandeg:
@askerlee: can you address/resolve the Copilot review comments?

PR #438, opened 2026-01-14
Fix args in readme

svlandeg:
Update the `device_batch_size` arg to `device-batch-size` according to [this latest refactor](https://github.com/karpathy/nanochat/commit/7312ec98985a8b478fc98ef54b30bdc0baed1989#diff-f01ca501612c5f260e12ac1171d6705e6887825cba06360c99429531072ef130), as otherwise we'd get

```
base_train.py: error: unrecognized arguments: --device_batch_size=16
```

 & add the required `-i` to example `chat_eval` runs in the inline documentation. This edit was originally by @pkrefta from https://github.com/karpathy/nanochat/pull/428

 & small typo fix

### DIFF
--- a/scripts/chat_eval.py
+++ b/scripts/chat_eval.py
@@ -4,8 +4,8 @@ All the generic code lives here, and all the evaluation-specific
 code lives in nanochat directory and is imported from here.
 
 Example runs:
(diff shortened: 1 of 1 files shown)


karpathy:
ah yes ,ty

Issue #427, opened 2026-01-09
base_eval.py: hellaswag gets progressively slower and leaks memory on small models (Mac Studio)


pkrefta:
I’m running `base_eval.py` on a very small model trained with `base_train.py`.  Evaluation starts normally, but on the **hellaswag** task each example becomes slower and slower, memory usage grows continuously, and the script will take massive amount time to finish. Other tasks finish in minutes, but hellaswag runs for hours.

Environment - Mac Studio, Python 3.10.19

I trained "much, much smaller" model  using a command from comment in `base_train.py`

```bash
python -m scripts.base_train --depth=4 --max_seq_len=512 --device_batch_size=1 --eval_tokens=512 --core_metric_every=-1 --total_batch_size=512 --num_iterations=20
```

I gave script few hours to run but it didn't finish. When interrupted with Ctrl+C, it always stops here:

```text
Traceback (most recent call last):
  File "/Users/user/.pyenv/versions/3.10.19/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/user/.pyenv/versions/3.10.19/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/Users/user/Projects/nanochat/scripts/base_eval.py", line 219, in <module>
    main()
  File "/Users/user/Projects/nanochat/scripts/base_eval.py", line 182, in main
    out = evaluate_model(model, tokenizer, device, max_per_task=args.max_per_task)
  File "/Users/user/Projects/nanochat/scripts/base_eval.py", line 107, in evaluate_model
    accuracy = evaluate_task(model, tokenizer, data, device, task_meta)
  File "/Users/user/Projects/nanochat/nanochat/core_eval.py", line 254, in evaluate_task
    is_correct = evaluate_example(idx, model, tokenizer, data, device, task_meta)
  File "/Users/user/Projects/nanochat/.venv/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
  File "/Users/user/Projects/nanochat/nanochat/core_eval.py", line 234, in evaluate_example
    mean_losses = [losses[i, si-1:ei-1].mean().item()
  File "/Users/user/Projects/nanochat/nanochat/core_eval.py", line 234, in <listcomp>
    mean_losses = [losses[i, si-1:ei-1].mean().item()
KeyboardInterrupt
[...]

PR #447, opened 2026-01-18
moved `save_checkpoint()` above `evaluate_model()`

### LINKED ISSUE #446: Move `save_checkpoint()` before `evaluate_model()`
I've just started using nanochat on an iMac M4 32GB and I'm having the same problem as issue #427 where evals are taking too much memory, but in the process of doing this and killing the evals, it doesn't save the checkpoints even though it has finished training.

If I move the `save_checkpoint()` if block above the `evaluate_model()` if block in `base_train.py` it does save the checkpoint before performing the evals.

I'm not familiar enough with the code base to know whether I should be running it with different parameters or what other effects this move might have.

The `save_checkpoint()` block:

https://github.com/karpathy/nanochat/blob/d58fcd9d7331efba0224d59e026833738e7547a6/scripts/base_train.py#L338-L360

Needs to be before the `evaluate_model()` if block:

https://github.com/karpathy/nanochat/blob/d58fcd9d7331efba0224d59e026833738e7547a6/scripts/base_train.py#L301-L315

jolonf:
This is a very simple change: the if block which does the `save_checkpoint()` is moved above the if block that does the `evaluate_model()` in base_train.py.

Prior to this change the saving of the checkpoint needed to wait until all of the evals were run, even though the model was trained.


svlandeg (commented):
In a sense this change is fine: the CORE metric evaluation loop and the model sampling loop don't set any parameters that are saved to disk, as far as I can see.

That said, if the evaluation runs out of memory that is a different problem (#427) that should be addressed, and in that sense this is really just a hacky workaround that doesn't actually fix that particular problem. But I can see how it would be nice to make sure the model gets saved if you do have the interrupt the eval process.

One more note: if this doesn't get merged right away, we'll have to be carefull that the code block L338-L360 hasn't diverged on `master` before we merge this in.

svlandeg (approved):

svlandeg (commented):
I updated the PR to fix merge conflicts with `master` and to ensure that the moved block is exactly what we have on `master` right now.

PR #451, opened 2026-01-19
fix: pass device_type to compute_init in engine.__main__

xiayan0118:
When running engine.py directly on non-GPU devices (CPU, MPS), compute_init() needs the device_type parameter to initialize correctly. This fixes failures on machines without CUDA support.


karpathy:
default values are terrible. always some issue comes up.

PR #454, opened 2026-01-20
Fix escape character in README bibtex entry

BlackSamorez:
Unescaped $ sign (as in `100$`) is a LaTeX error. The citation becomes misformed. This PR fixes it.


svlandeg (approved):
Good catch, thanks!

BlackSamorez:
For reference, some pre-prints citing nanochat apply exactly the same fix:
 - https://arxiv.org/abs/2511.13761
 - https://arxiv.org/abs/2511.13319

PR #456, opened 2026-01-22
Fix SDPA KV-cache decode to respect sliding window

### LINKED ISSUE #452: Flash Attention Fallback - Missing sliding window in `flash_attn_with_kvcache`?
In the newly imlemented flash attention module, there is a if logic
```
# Single token generation
    if Tq == 1:
        return F.scaled_dot_product_attention(q, k, v, is_causal=False, enable_gqa=enable_gqa)
```

For cases where kv_cache is not None, seems like it forgets to consider sliding window?

KartikVashishta:
Fix #452 : SDPA fallback now respects sliding window during single-token KV-cache decode by slicing K/V to the last window + 1 tokens.

Note: Keep is_causal=False for SDPA Tq == 1 due to PyTorch SDPA KV-cache causal behavior: https://github.com/pytorch/pytorch/issues/144858

### DIFF
--- a/nanochat/flash_attention.py
+++ b/nanochat/flash_attention.py
@@ -69,28 +69,28 @@ def _sdpa_attention(q, k, v, window_size, enable_gqa):
     if (window < 0 or window >= Tq) and Tq == Tk:
         return F.scaled_dot_product_attention(q, k, v, is_causal=True, enable_gqa=enable_gqa)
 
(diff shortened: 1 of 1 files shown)


karpathy:
Merged via 3ba42e8 with an additional test case. Thanks @KartikVashishta!

PR #464, opened 2026-01-26
Combined Optimizer

chrisjmccormick:
Hi there! 
I contributed many of the recent optimizer… optimizations to modded, such as replicating small parameters, fusing all of the adam and muon update math (piping in hyperparameters via 0-D cpu tensors was the secret sauce that unlocked all that!) and reworking the variance reduction algebra to be more memory efficient. 

It was awesome to see those changes pulled into this repo, and the consolidated fused Muon update is very elegant. I’ve written a similar function now to incorporate into modded. Also, the way you consolidated the gradient stacking ops is a big improvement; I had issues getting that to work and clearly gave up on it too quickly.

I’m planning to publish the optimized optimizer in its own repo, with a more thorough write-up on the improvements, and figured this could be a great place to discuss it and get feedback on what the broader community might find most helpful (or troublesome).

I think the version here in nanochat is a better starting point, since we’ve started making some pretty exotic changes to the one in modded (probably with more on the way). 

This pull request includes some potential changes I have in mind for the ‘published’ version that I thought make sense here as well.

### Combined Optimizers

Similar to the published NorMuon repo, I refactored nanochat’s optimizer in this PR to be a combined “MuonWithAdam”.

**Conceptual Value**

In terms of pedagogical value and cognitive load, honestly, I could see it both ways (merged or separate). You can see in the PR it simplifies the code in some places by eliminating the loops over multiple optimizers.

The conceptual framing would be “there’s still just one optimizer, but we handle the update math differently depending on the parameter”. Since you’ve already moved to per-parameter betas and lr, I think that’s not too big of a leap to specify the per-parameter optimizer as well.

**Communication**

I think where the merge makes the most sense is in the distributed implementation. It feels more elegant, I think, to have unified logic around the 3-part process of “scatter gradients, work on gradients, reduce updates”.

When you look at the traces, you can also see how dividing the ‘step’ functions creates an unnecessary synchronization barrier, where Adam must complete its three phases before Muon can start any comms. 

image

It’s not much of a problem, since optimization is so comms-dominated, and probably won't register as an improvement in these larger runs. I don’t think merging the optimizers adds complexity, though, so as long as it doesn't make anything slower, it seems worth considering.

Here's what it looks like with this PR. I've learned you can't read too much into what the comms look like in these trace files, because it can vary significantly. But it at least conveys the flow and the potential improvement. 

image

Another small thing is that merging the optimizers allows for explicit control over the ordering of communication and work. I don’t have a good explanation for why (I think I need better visibility into what’s actually going on in comms across the gpus), but we’ve seen speed improvements in modded from changing the order around.  

### Non-Distributed Adam

I noticed that there was a non-distributed version of Muon implemented–that’s a nice inclusion for anyone reading the code to learn about the math, since it eliminates the complexity of the communication logic. 

As part of merging the optimizers, I implemented a non-distributed AdamW to merge with Muon. 

If you do decide to keep the optimizers separate, a small note on the DistAdamW implementation is that it doesn’t currently gate the comms for the single gpu case, so there’s some unnecessary overhead on single GPUs occurring because it’s still launching those scatters.

### Adam Padding, VE Gates

I noticed that you currently have the VE Gates optimized by Muon. We moved all of our gates to Adam, primarily because fused Adam is ridiculously fast compared to Muon, but I think it’s also interesting to consider whether it’s beneficial to require the model to use orthogonal directions of the residual stream to gate the different heads, vs. allowing some dependance between them. 

I was curious if you had tested this. I think @classiclarry did find that moving the gates to adam had a slight penalty to loss. Thinking about it now, I’m wondering if it has more to do with the normalization and variance reduction of the gate weights than the orthogonalization. I’ll have to play with that!

Moving them to Adam would add some communication overhead, though, because it means that each layer’s gates would be communicated separately, whereas Muon stacks them into a single op. 

I’m contributing a change to modded that I think could make a lot of sense in general. Similar to how Adam currently checks the parameter sizes and uses all_reduce for the small ones, I’m planning to have the optimizer pack them all into a single comms op. This allows them to remain independent and have their own hyperparameters, while avoiding the penalty of all of the tiny comms ops. The gates are small enough to be included in this packing op.

### Reshaping for Muon

Muon is very shape sensitive. Unlike AdamW, which updates parameters independently, the Muon update math breaks if you slice up weight matrices, or if you concatenate matrices together (such as fusing QKV).

This is an interesting problem. It’s currently avoided in nanochat because Q, K, V are kept separate. However, it seems to be standard practice to fuse matrix operations wherever possible. This creates a footgun for anyone who is using nanochat as the basis for their research and decides to fuse a matrix op. If they’re not aware of this property of Muon, their validation loss will suffer.

The solution I’m hoping to go with, and that we currently have in modded, is to require that any parameter optimized by Muon needs to explicitly state how it should be viewed/reshaped for the orthogonalization step. This serves both to make the user aware of this important consideration, and to add support for fusing matrices like QKV. 

I haven’t added that to this PR yet, though.

Finally–thank you so much for nanoGPT and nanochat, they’re (obviously) a huge contribution to our community. 

Look forward to hearing any feedback you have on the optimizers.

### Smaller Details
* Added support for single GPU with torchrun, e.g., `torchrun --nproc_per_node=1 scripts.base_train.py` by gating on both ddp and world size: 
```python
OptimizerClass = DistMuonAdamW if (ddp and world_size > 1) else MuonAdamW
```


speedrun.sh
```diff
@@ -0,0 +1,153 @@
+#!/bin/bash
```
svlandeg: Did you mean to add this one as a separate new file? On current `master`, this script has been moved to the `runs` subfolder so if you needed to make any changes to it, it'd be best to do it there :-)

chrisjmccormick: Oops! No, this doesn't belong in the commit at all, I'll remove it.

chrisjmccormick:
One other thing I found to maybe incorporate. The Nesterov momentum step for Muon uses a slower 'transpose' kernel for the tall MLP input matrix:

image

The current Polar express code checks for tall matrices and just transposes them:

```python
# Polar express
X = g.bfloat16()
if g.size(-2) > g.size(-1):
    X = X.mT
X = X / (X.norm(dim=(-2, -1), keepdim=True) * 1.02 + 1e-6)
for a, b, c in polar_express_coeffs[:ns_steps]:
    A = X @ X.mT
    B = b * A + c * (A @ A)
    X = a * X + B @ X
if g.size(-2) > g.size(-1):
    X = X.mT
g = X
```

This is often fine because CUDA is good at hiding any memory layout issues for matrix multiplication, but it can't hide layout issues for element-wise operations like applying the momentum buffer. 

There's an easy fix, though--we can flip the algebra around for tall matrices, and do:

```python
# Polar express
X = g.bfloat16()

X = X / (X.norm(dim=(-2, -1), keepdim=True) * 1.02 + 1e-6)

if g.size(-2) > g.size(-1): # Tall matrix
    for a, b, c in polar_express_coeffs[:ns_steps]:
        A = X.mT @ X
        B = b * A + c * (A @ A)
        X = a * X + X @ B

else: # Wide matrix (original math)
    for a, b, c in polar_express_coeffs[:ns_steps]:
        A = X @ X.mT
        B = b * A + c * (A @ A)
        X = a * X + B @ X

return X
```

The first two lines, calculating `A` and `B`, are identical between the two shapes. 
The third one holds because BX = X'B

Small wins. :)

A related but more nuanced topic is that it's possible to store both weight matrices as tall, and then they get stacked and operated on in a single Muon pass instead of two. It works out nicely for `d20`, too, since it becomes a stack of 40, which evenly spreads across 8 gpus.

The forward and backward kernels are just as fast, but there are some interesting considerations around some shape-based heuristics built into NorMuon. Things improved in modded when we switched to the tall orientation on both matrices. I can elaborate more on that when I get a chance.

chrisjmccormick:
I tried running some experiments to measure any improvement from merging the optimizers. I haven't been able to determine that yet, but here's what I've found so far. 

In the below plot:

* The cool colored runs are the current repo (which includes the bigram table) and the warm colored runs are this PR. 
    * So the difference here is probably much more to do with the bigram table than the optimizers.
* I clamped the step times to 1.07s because otherwise the garbage collection steps make the plots hard to read. (The weird logic in the plot title is the clamping)
* These are d20 on an 8xA100, with FA2 (without FA2 the step time was ~1.6s). 
* I tried d25 but hit OOM.

image

I tried merging in the bigram tables into this PR to get some data that included those, but I hit OOM on the 8xA100 instance. I'd probably need to roll back to the previous commit to make a comparison here? Or get my hands on an 8xH100 instance, which I haven't been able to do for a while.

Or it sounds like we may be rolling back the bigrams anyway?

The other noteworthy thing in the plots is that there are definitely outlier runs. The second run of the baseline, and the 1st run of this PR were both a good deal lower than their respective means.

**Recompiles**

Below are a couple log files which include the recompiles printed out. It looks the bigram table introduced some more recompiles.

[d20-8xA100-FA2-merged-optims-2.log](https://github.com/user-attachments/files/24920477/d20-8xA100-FA2-merged-optims-2.log)
[d20-8xA100-FA2-baseline-2.log](https://github.com/user-attachments/files/24920479/d20-8xA100-FA2-baseline-2.log)

**Next**

So no answer yet to whether the merged optimizer is faster or slower. I wouldn't expect it to make much of a difference, so it might be more of a philosophical choice to include it or not. I'll have to keep testing, but wanted to give an update and get any thoughts on what to do about the bigram table. Thanks!

chrisjmccormick:
Here's the actual run data if you want to poke at it:

https://wandb.ai/chrismccormick/nanochat-merged-optims?nw=nwuserchrismccormick

svlandeg:
With this mostly merged [here](https://github.com/karpathy/nanochat/commit/41bb2eac320b927ec093d31db63d8b8a4aac6b2c), should we close this one?

PR #467, opened 2026-01-28
Fix generate() crash when top_k=0

### LINKED ISSUE #466: generate() crashes when top_k=0
### Problem

The `generate()` method crashes when it is called with `top_k=0`.

Although `top_k=0` is treated as a valid value elsewhere in the project (used to indicate that top-k filtering is disabled), `generate()` does not handle this case correctly. When `top_k` is set to zero, the function still attempts to apply top-k filtering to the logits, which results in invalid tensor operations and causes generation to fail.

This leads to inconsistent behavior between `generate()` and other sampling logic in the codebase, and prevents `generate()` from being used safely with `top_k=0`.

gupta-8:
Fixes a crash in `generate()` when `top_k` is set to 0 by treating it as disabling top-k filtering, keeping generation behavior consistent with the rest of the code. Fixes #466

### DIFF
--- a/nanochat/gpt.py
+++ b/nanochat/gpt.py
@@ -489,7 +489,7 @@ class GPT(nn.Module):
         for _ in range(max_tokens):
             logits = self.forward(ids) # (B, T, vocab_size)
             logits = logits[:, -1, :] # (B, vocab_size)
-            if top_k is not None:
+            if top_k is not None and top_k > 0:
                 v, _ = torch.topk(logits, min(top_k, logits.size(-1)))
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
Thanks for the PR @gupta-8, makes sense.

I can replicate the crash by executing the quick script in `engine.py` and setting
```
kwargs = dict(max_tokens=64, temperature=0.0, top_k=0)
```
The crash is fixed with this PR, and the edit you have here is in line with `sample_next_token` from `engine.py` which also checks
```
if top_k is not None and top_k > 0:
```

PR #470, opened 2026-01-28
Fix link to DeepSeek Engram paper

svlandeg:
Link to https://arxiv.org/abs/2601.07372 instead


karpathy:
wow, this was totally hallucinated by an LLM lol!

PR #475, opened 2026-01-30
Add Blackwell (SM100) GPU support via SDPA fallback

francip:
This PR was created using Claude Code and reviewed and tested by @francip on an RTX Pro 6000 Blackwell GPU.

**Problem**

Flash Attention 3 kernels are compiled specifically for Hopper architecture (SM90). The previous detection logic checked if major < 9, which meant any GPU with compute capability 9.0+ would attempt to load FA3 kernels. This caused a crash on Blackwell GPUs (SM100):

CUDA error (/root/flash-attention/hopper/flash_fwd_launch_template.h:164):
no kernel image is available for execution on the device

**Solution**

Changed the FA3 detection to check for exactly SM90 (Hopper) rather than SM90+:

# Before

```
if major < 9:  # Hopper is sm90
    return None
```

# After

```
if major != 9:  # FA3 kernels compiled for Hopper (sm90) only
    return None
```

This ensures:
- Hopper (SM90): Uses Flash Attention 3 ✓
- Blackwell (SM100): Falls back to PyTorch SDPA ✓
- Ada (SM89) and older: Falls back to PyTorch SDPA ✓

Testing

Verified on RTX Pro 6000 (Blackwell/SM100):
- Inference works correctly with SDPA fallback
- Training scripts use the same unified flash_attn interface and will work correctly

**Future Work**

When FA3 kernels are compiled for Blackwell, the check can be updated to:

```
if major not in (9, 10):  # Hopper (sm90) and Blackwell (sm100)
```

### DIFF
--- a/nanochat/flash_attention.py
+++ b/nanochat/flash_attention.py
@@ -25,7 +25,7 @@ def _load_flash_attention_3():
     if not torch.cuda.is_available():
         return None
     try:
-        major, minor = torch.cuda.get_device_capability()
+        major, _ = torch.cuda.get_device_capability()
         # FA3 kernels are compiled for Hopper (sm90) only
         # Ada (sm89), Blackwell (sm100) need SDPA fallback until FA3 is recompiled
         if major != 9:


nanochat/flash_attention.py
```diff
@@ -21,12 +21,14 @@
...
 # =============================================================================
 def _load_flash_attention_3():
-    """Try to load Flash Attention 3 (requires Hopper+ GPU)."""
+    """Try to load Flash Attention 3 (requires Hopper GPU, sm90)."""
     if not torch.cuda.is_available():
         return None
     try:
-        major, _ = torch.cuda.get_device_capability()
-        if major < 9:  # Hopper is sm90
+        major, minor = torch.cuda.get_device_capability()
```
svlandeg: (avoiding unnecessary edits)
```suggestion
        major, _ = torch.cuda.get_device_capability()
```

svlandeg (approved):
Hi @francip, thanks for the PR! Could you please tick the box that says something like "Maintainers are allowed to edit this pull request."? It's easier to get it through the review & merge process if we can edit 🙂 

Anyway, fix looks good to me! 🙏 

Did you test yourself that this PR now runs correctly on your Blackwell setup?

francip:
Yes, I tested it both for inference and with a d20 SFT training run on 900 samples. Will check the checkbox once I get to laptop, sorry.

francip:
Can't find the dang checkbox on mobile site, merged your suggestion and verified it runs. (Termius + Tailscale FTW)

francip:
Lol, I am probably going crazy, but I swear that checkbox has been disappeared. I can't find it on the web either, and I am not sure why. Anyway, let me know if you need me to change anything else before merging.

svlandeg:
> Lol, I am probably going crazy, but I swear that checkbox has been disappeared. I can't find it on the web either, and I am not sure why. Anyway, let me know if you need me to change anything else before merging.

Ah, not sure, could be that you only see the checkbox when creating the PR 🤷‍♀️

PR #482, opened 2026-01-31
Changyao/pre train

changyaochen:


svlandeg (commented):
Hi! It would be appropriate to leave some kind of description in the PR's body to explain the changes; what they are and why. 

Anyway looks like formatting, typing and docstring changes. Judging from similar PRs in the past,  I doubt we'll want to merge this...

PR #483, opened 2026-02-01
docs: clarify LR warmdown semantics in base_train

dipeshbabu:
This PR updates the LR schedule comments and semantics description for `base_train.py` to better reflect intended behavior and edge cases.

- The warmup and warmdown regions:
- Warmup runs over the first `warmup_iters` optimizer steps.
- Warmdown runs over the last `warmdown_iters` optimizer steps.
- The schedule is robust to `warmup_iters==0`, `warmdown_iters==0`, and warmup/warmdown overlap.

Note: This is primarily a clarification/cleanup PR (not a behavioral bug fix).

### DIFF
--- a/scripts/base_train.py
+++ b/scripts/base_train.py
@@ -347,27 +347,33 @@ print0(f"Tokens : Scaling params ratio: {total_batch_size * num_iterations / num
 print0(f"Total training FLOPs estimate: {num_flops_per_token * total_tokens:e}")
 
 # Learning rate schedule (linear warmup, constant, linear warmdown)
-def get_lr_multiplier(it):
-    # Note: optimizer steps run for it in [0, num_iterations-1]
+def get_lr_multiplier(it: int) -> float:
+    # Note: optimizer steps run for it in [0, num_iterations - 1]
     warmup_iters = round(args.warmup_ratio * num_iterations)
     warmdown_iters = round(args.warmdown_ratio * num_iterations)
 
-    # Warmup (avoid division by zero when warmup_iters == 0)
-    if warmup_iters > 0 and it < warmup_iters:
+    # Clamp to sane ranges
+    warmup_iters = max(0, min(warmup_iters, num_iterations))
+    warmdown_iters = max(0, min(warmdown_iters, num_iterations))
+
+    # Warmup: linear ramp from (1/warmup_iters) .. 1.0 over warmup_iters steps
+    if it < warmup_iters:
+        # safe: if warmup_iters == 0 this branch is unreachable
         return (it + 1) / warmup_iters
 
-    # Warmdown should cover the last `warmdown_iters` optimizer steps:
+    # Warmdown: apply over the last warmdown_iters optimizer steps:
     # it in [num_iterations - warmdown_iters, num_iterations - 1]
     if warmdown_iters > 0:
         warmdown_start = num_iterations - warmdown_iters
-        # Ensure warmdown doesn't start before warmup ends (prevents overlap weirdness)
+
+        # If warmup overlaps warmdown, start warmdown only after warmup is done
         warmdown_start = max(warmdown_start, warmup_iters)
 
         if it >= warmdown_start:
-            # progress: 1.0 at warmdown_start, 0.0 at last optimizer step (num_iterations - 1)
-            span = max(1, (num_iterations - 1) - warmdown_start)  # denom >= 1
-            progress = (num_iterations - 1 - it) / span
-            return progress * 1.0 + (1.0 - progress) * args.final_lr_frac
+            # progress goes 1.0 -> 1.0/warmdown_iters (NOT to 0 by default)
+            # This matches the original behavior where final step doesn't hit 0
+            progress = (num_iterations - it) / warmdown_iters
+            return args.final_lr_frac + (1.0 - args.final_lr_frac) * progress
 
     return 1.0
 


scripts/base_train.py
```diff
@@ -348,15 +348,28 @@ def get_scaling_params(m):
 
 # Learning rate schedule (linear warmup, constant, linear warmdown)
 def get_lr_multiplier(it):
-    warmup_iters = round(args.warmup_ratio * num_iterations)
-    warmdown_iters = round(args.warmdown_ratio * num_iterations)
-    if it < warmup_iters:
+    # Note: optimizer steps run for it in [0, num_iterations-1]
+    warmup_iters = int(round(args.warmup_ratio * num_iterations))
+    warmdown_iters = int(round(args.warmdown_ratio * num_iterations))
```
svlandeg: We don't need `int()` here, as `round` without a second argument `ndigits` will in fact produce an `int`:
```
x = round(3.42)
print(x, type(x))

# 3 <class 'int'>
```

scripts/base_train.py
```diff
@@ -348,15 +348,28 @@ def get_scaling_params(m):
 
 # Learning rate schedule (linear warmup, constant, linear warmdown)
 def get_lr_multiplier(it):
+    # Note: optimizer steps run for it in [0, num_iterations-1]
     warmup_iters = round(args.warmup_ratio * num_iterations)
     warmdown_iters = round(args.warmdown_ratio * num_iterations)
-    if it < warmup_iters:
+
+    # Warmup (avoid division by zero when warmup_iters == 0)
+    if warmup_iters > 0 and it < warmup_iters:
```
svlandeg: There's no need for extra code to avoid this zero division. Because if `warmup_iters` is 0, then `it` will never be smaller than it. One thing you could do is `assert it >= 0` at the start of the function.

dipeshbabu: Good point, agreed. If warmup_iters == 0, it < warmup_iters is never true, so the division is unreachable. I’ll simplify the warmup block accordingly.

dipeshbabu:
> Hi, thanks for the PR!
> 
> When I run a toy `base_train` run on `master` with `--num-iterations=20 --warmup-ratio=0.25 --warmdown-ratio=0.25`, the `iters` should be 5 (20/4) for both warmup and warmdown, and I get:
> 
> ```
> step 00000/00020 (0.00%) | loss: 11.090847 | lrm: 0.20
> step 00001/00020 (5.00%) | loss: 11.084596 | lrm: 0.40
> step 00002/00020 (10.00%) | loss: 11.071568 | lrm: 0.60
> step 00003/00020 (15.00%) | loss: 11.050060 | lrm: 0.80
> step 00004/00020 (20.00%) | loss: 11.004250 | lrm: 1.00
> step 00005/00020 (25.00%) | loss: 10.904243 | lrm: 1.00
> step 00006/00020 (30.00%) | loss: 10.664922 | lrm: 1.00
> step 00007/00020 (35.00%) | loss: 10.513832 | lrm: 1.00
> step 00008/00020 (40.00%) | loss: 10.258086 | lrm: 1.00
> step 00009/00020 (45.00%) | loss: 10.024872 | lrm: 1.00
> step 00010/00020 (50.00%) | loss: 9.811623 | lrm: 1.00
> step 00011/00020 (55.00%) | loss: 9.613647 | lrm: 1.00
> step 00012/00020 (60.00%) | loss: 9.400777 | lrm: 1.00
> step 00013/00020 (65.00%) | loss: 9.212350 | lrm: 1.00
> step 00014/00020 (70.00%) | loss: 9.029242 | lrm: 1.00
> step 00015/00020 (75.00%) | loss: 8.885534 | lrm: 1.00
> step 00016/00020 (80.00%) | loss: 8.747950 | lrm: 0.80
> step 00017/00020 (85.00%) | loss: 8.606284 | lrm: 0.60
> step 00018/00020 (90.00%) | loss: 8.468077 | lrm: 0.40
> step 00019/00020 (95.00%) | loss: 8.347053 | lrm: 0.20
> ```
> 
> so step 0-4 has the warmup with `lrm` 0.2, 0.4, 0.6, 0.8 and then 1. The warmdown steps are in steps 16-19 with 0.8, 0.6, 0.4 and 0.2. Or you could interpret this as step 15 being the first warmdown step at `lrm` 1.0 (which is not exactly what the code does, but it boils down to the same thing in practice).
> 
> With this PR, I'd get:
> 
> ```
> step 00000/00020 (0.00%) | loss: 11.090847 | lrm: 0.20
> step 00001/00020 (5.00%) | loss: 11.084596 | lrm: 0.40
> step 00002/00020 (10.00%) | loss: 11.071568 | lrm: 0.60
> step 00003/00020 (15.00%) | loss: 11.050060 | lrm: 0.80
> step 00004/00020 (20.00%) | loss: 11.004249 | lrm: 1.00
> step 00005/00020 (25.00%) | loss: 10.904243 | lrm: 1.00
> step 00006/00020 (30.00%) | loss: 10.664915 | lrm: 1.00
> step 00007/00020 (35.00%) | loss: 10.513753 | lrm: 1.00
> step 00008/00020 (40.00%) | loss: 10.258013 | lrm: 1.00
> step 00009/00020 (45.00%) | loss: 10.024803 | lrm: 1.00
> step 00010/00020 (50.00%) | loss: 9.811552 | lrm: 1.00
> step 00011/00020 (55.00%) | loss: 9.613564 | lrm: 1.00
> step 00012/00020 (60.00%) | loss: 9.400695 | lrm: 1.00
> step 00013/00020 (65.00%) | loss: 9.212276 | lrm: 1.00
> step 00014/00020 (70.00%) | loss: 9.029181 | lrm: 1.00
> step 00015/00020 (75.00%) | loss: 8.885474 | lrm: 1.00
> step 00016/00020 (80.00%) | loss: 8.747892 | lrm: 0.75
> step 00017/00020 (85.00%) | loss: 8.606513 | lrm: 0.50
> step 00018/00020 (90.00%) | loss: 8.469306 | lrm: 0.25
> step 00019/00020 (95.00%) | loss: 8.350844 | lrm: 0.00
> ```
> 
> -> the warmup is still the same, and the amount of warmdown steps (in practice) is the same as well, but now the multiplier goes down to 0 instead of to 0.2.
> 
> The PR description says that "warmdown began one step too late" but in fact the warmdown steps with the implementation of this PR haven't changed, it's the actual `lrm` value that now falls in `[0,1]` instead of stopping at `0.2` previously. The first warmdown step now drops even sharper.
> 
> Anyway I don't think this is an actual bug, but rather intended behaviour, that `lrm` doesn't go down to `0.0` in the final learning step.

Thanks, this isn’t really a bug; it’s schedule semantics. I’m updating this PR to avoid claiming an off-by-one fix, keep the LR behavior consistent with master, and limit changes to clarification and minor cleanup (e.g. removing redundant `int(round(...))`). Force-pushed update incoming.

karpathy:
I don't fully understand what this PR is trying to fix but I think it's something very minor and it adds a lot of lines.

PR #486, opened 2026-02-01
fix SFT row packing to avoid fully padded rows

chenzl25:
~Ensure SFT conversation rendering respects row_capacity by passing it as max_tokens.~
- I ran `runcpu.sh` and found that max-seq-len is 512 by default. During SFT the model loss would go down almost 0 resulting a really bad overfitting (generate `BOS` always for any inputs).
 - Update the docstring to reflect padding behavior and empty-row truncation.
 - When no conversation fits and the row is empty, truncate the shortest conversation to avoid fully padded rows.

### DIFF
--- a/scripts/chat_sft.py
+++ b/scripts/chat_sft.py
@@ -152,7 +152,7 @@ def sft_data_generator_bos_bestfit(split, buffer_size=100):
         nonlocal cursor, epoch
         while len(conv_buffer) < buffer_size:
             conversation = dataset[cursor]
-            ids, _ = tokenizer.render_conversation(conversation)
+            ids, _ = tokenizer.render_conversation(conversation, max_tokens=row_capacity)
             conv_buffer.append(ids)
             cursor += ddp_world_size
             if cursor >= dataset_size:
--- a/scripts/chat_sft.py
+++ b/scripts/chat_sft.py
@@ -130,7 +130,8 @@ def sft_data_generator_bos_bestfit(split, buffer_size=100):
 
     Each row in the batch starts with BOS (beginning of a conversation).
     Conversations are packed using best-fit algorithm. When no conversation fits,
-    the row is padded (instead of cropping) to ensure no tokens are ever discarded.
+    the row is padded. If the row is still empty, we fall back to truncating one
+    conversation to avoid fully padded rows.
     Padding positions have targets masked with -1 (ignore_index for cross-entropy).
     """
     global last_step, approx_progress, current_epoch
@@ -152,7 +153,7 @@ def sft_data_generator_bos_bestfit(split, buffer_size=100):
         nonlocal cursor, epoch
         while len(conv_buffer) < buffer_size:
             conversation = dataset[cursor]
-            ids, _ = tokenizer.render_conversation(conversation, max_tokens=row_capacity)
+            ids, _ = tokenizer.render_conversation(conversation)
             conv_buffer.append(ids)
             cursor += ddp_world_size
             if cursor >= dataset_size:
@@ -188,12 +189,25 @@ def sft_data_generator_bos_bestfit(split, buffer_size=100):
                     row.extend(conv)
                     consumed += ddp_world_size  # Track actual consumption
                 else:
-                    # No conversation fits - pad the remainder instead of cropping
-                    # This ensures we never discard any tokens
-                    content_len = len(row)
-                    row.extend([bos_token] * remaining)  # Pad with BOS tokens
-                    padded = True
-                    break  # Row is now full (with padding)
+                    if len(row) == 0:
+                        # No conversation fits even on an empty row.
+                        # Truncate the shortest one to avoid fully padded rows.
+                        trunc_idx = -1
+                        trunc_len = None
+                        for i, conv in enumerate(conv_buffer):
+                            conv_len = len(conv)
+                            if conv_len > remaining and (trunc_len is None or conv_len < trunc_len):
+                                trunc_idx = i
+                                trunc_len = conv_len
+                        conv = conv_buffer.pop(trunc_idx)
+                        row.extend(conv[:row_capacity])
+                        consumed += ddp_world_size
+                    else:
+                        # No conversation fits - pad the remainder
+                        content_len = len(row)
+                        row.extend([bos_token] * remaining)  # Pad with BOS tokens
+                        padded = True
+                    break  # Row is now full (with padding or truncation)
 
             # Track content length: full row if no padding, otherwise the length before padding
             if padded:


svlandeg (commented):
Hm, I worry about the effect this will have on "normal" runs where `max-seq-len` is not overly small. You'll end up with many truncated conversations, right? That doesn't seem ideal in the SFT stage.

chenzl25:
> Hm, I worry about the effect this will have on "normal" runs where max-seq-len is not overly small. You'll end up with many truncated conversations, right? That doesn't seem ideal in the SFT stage.
 
Good point. When you say “max‑seq‑len is not overly small,” do you mean something like 512–2048?

 `render_conversation` already truncates to max_tokens (default 2048), so we’re already truncating any conversation longer than that. This change just aligns the truncation cap with the actual row_capacity we’re
  packing into, which avoids the “never fits → pad forever” failure mode.

 So if row_capacity = 4096, capping at 4096 is actually less truncation than the current default 2048, and it lets us preserve more of the original conversation.

 For the runcpu case, I think truncation is an acceptable trade‑off: the alternative is silently dropping long conversations and potentially producing padded‑only batches, which is worse.

svlandeg:
> For the runcpu case, I think truncation is an acceptable trade‑off

The truncation completely destroys the point of this "best fit" dataloader though. I had a look into the actual data: the first 300 examples of the `train` split contain encoded conversations of length 0-1500. When the algorithm kicks of with the normal code from `master` and `max-seq-len=512`, it will first select an encoded conversation of length 509. This is nice and the point of the best fit loader: the whole thing fits. The 4 remaining places are padded. Next, a conversation of length 500 is added, then another of len 498, then one of len 487 which gets combined with a small conversation of len 26. etc.

But with the edit on this PR, all the truncated conversation of length > 512 will take precedence, because they seem to fit more "perfectly" than the 509 one for instance. So the whole row gets filled with truncated conversations, and the actual best fitting untruncated ones don't get a chance to get selected.

> This change just aligns the truncation cap with the actual row_capacity we’re packing into, which avoids the “never fits → pad forever” failure mode.

> the alternative is silently dropping long conversations and potentially producing padded‑only batches, which is worse.

I think the alternative better fix should be to first add the best-fitting ones; there are plenty of conversations that fit, untruncated, into the rows even with `max-seq-len=512`. However, at the point in the loop where there is no data left in the buffer that fits, even when `remaining == row_capacity`, perhaps at that point we should consider truncating the remaining conversations and using those instead of producing fully padded rows.

svlandeg:
> When you say “max‑seq‑len is not overly small,” do you mean something like 512–2048?

Yes, exactly.

> `render_conversation` already truncates to max_tokens (default 2048), so we’re already truncating any conversation longer than that. (...) So if row_capacity = 4096, capping at 4096 is actually less truncation than the current default 2048, and it lets us preserve more of the original conversation.

I don't think I've seen runs using `max_seq_len` larger than 2048, as that is the default that is used in `chat_sft.py`. And from looking at the data it seems like truncating at 2048 will hardly truncate anything, most of the encoded conversations are smaller.

Still, IMO we should think about whether we want this default truncation at 2048 on `render_conversation` - not capping (or capping slightly too large) might actually be better and more expected behaviour in combination with the best fit algorithm. This would be the opposite solution as to what this PR is proposing though.

chenzl25:
> Still, IMO we should think about whether we want this default truncation at 2048 on render_conversation - not capping (or capping slightly too large) might actually be better and more expected behaviour in combination with the best fit algorithm. This would be the opposite solution as to what this PR is proposing though.

You are right. That make sense. I follow your idea to keep the best-fit algorithm work as expected and do the following changes: when we find that the row is empty after the best-fit algorithm, we can fallback to conversation truncation, so there is no fully padded rows in the worst case.

chenzl25:
I run the sft on the `runcpu` model and the result looks good for the typical question.
> "What is the capital of France?". 

Now it can answer 

> "Assistant: The capital of France is Paris. It is a city of Paris, located in the heart of the city and is located in the heart of France.<|assistant_end|>"

marcinbogdanski:
Please ignore my PR #817 above. It proposes the same solution that was initially discussed above (adding `max_tokens=row_capacity`, [d4db003](https://github.com/karpathy/nanochat/pull/486/commits/d4db00366119430fad124a59ff1f31ec855ca6ee))

I can confirm this issue is still live on current master `92d63d4e`. The current symptom: a run with `--max-seq-len=512` generates NaNs due to fully-padded batches.

I ran a quick test on both `master` and this PR merged.
```bash
git checkout master

# Get pre-trained checkpoint

# Test current master - NaN on step 00004
uv run python -m scripts.chat_sft --model-tag=d4-sft-bug --max-seq-len=512 --total-batch-size=16384 --eval-every=-1 --chatcore-every=-1
...
step 00001 (0.01%) | loss: 5.465866 | lrm: 1.00 | dt: 2853.09ms | tok/sec: 5,742 | mfu: 0.65 | epoch: 1 | total time: 0.00m
step 00002 (0.02%) | loss: 7.122874 | lrm: 1.00 | dt: 39.83ms | tok/sec: 411,353 | mfu: 46.48 | epoch: 1 | total time: 0.00m
step 00003 (0.02%) | loss: 8.389245 | lrm: 1.00 | dt: 39.67ms | tok/sec: 413,047 | mfu: 46.67 | epoch: 1 | total time: 0.00m
step 00004 (0.02%) | loss: nan | lrm: 1.00 | dt: 40.39ms | tok/sec: 405,666 | mfu: 45.83 | epoch: 1 | total time: 0.00m

# Merge this PR locally for testing
git checkout -b test-486-merge
git pull --no-edit --no-rebase origin refs/pull/486/head

# Test with PR merged - full epoch
uv run python -m scripts.chat_sft --model-tag=d4-sft-bug --max-seq-len=512 --total-batch-size=16384 --eval-every=-1 --chatcore-every=-1
...
step 00001 (0.01%) | loss: 5.465866 | lrm: 1.00 | dt: 2912.09ms | tok/sec: 5,626 | mfu: 0.64 | epoch: 1 | total time: 0.00m
step 00002 (0.02%) | loss: 7.122874 | lrm: 1.00 | dt: 40.39ms | tok/sec: 405,635 | mfu: 45.83 | epoch: 1 | total time: 0.00m
step 00003 (0.02%) | loss: 8.390747 | lrm: 1.00 | dt: 39.81ms | tok/sec: 411,530 | mfu: 46.50 | epoch: 1 | total time: 0.00m
...
step 22738 (99.99%) | loss: 2.548774 | lrm: 0.00 | dt: 38.97ms | tok/sec: 420,410 | mfu: 47.50 | epoch: 1 | total time: 14.83m
step 22739 (99.99%) | loss: 2.529194 | lrm: 0.00 | dt: 38.97ms | tok/sec: 420,428 | mfu: 47.50 | epoch: 2 | total time: 14.83m
step 22740 (100.00%) | loss: 2.537086 | lrm: 0.00 | dt: 38.99ms | tok/sec: 420,189 | mfu: 47.47 | epoch: 2 | total time: 14.83m
step 22741 (100.00%) | loss: 2.536981 | lrm: 0.00 | dt: 39.94ms | tok/sec: 410,209 | mfu: 46.35 | epoch: 2 | total time: 14.83m
step 22742 (100.00%) | loss: 2.526981 | lrm: -0.00 | dt: 38.93ms | tok/sec: 420,884 | mfu: 47.55 | epoch: 2 | total time: 14.83m
Step 22742 | Validation bpb: 0.7822
```

- Note 1: lrm `-0.00` on the final step is due to a separate bug, described in #816
- Note 2: `--total-batch-size=16384` to pin `grad_accum=1`, because SFT is currently broken for `grad_accum > 1` (see #816)


It seems this PR fixes the NaNs, but I have not investigated beyond a quick sanity check, and I don't endorse it either way. I just happened to be mentally in context.

Happy to run more tests if helpful.

marcinbogdanski:
I would like to propose extending this PR with an additional step 1). On the 512 example:

1) In `refill_buffer()` drop conversations that are fully masked in the first 513 tokens - they provide no training signal in any case.
    - this fixes fully-masked 513-length MMLU items being picked as a "perfect fit" ahead of shorter, untrimmed conversations (table below `MMLU, C)`)
    - dropped fully masked conversations should still count towards `consumed` so progress tracking/LR-scaling works as before
    - as a side note, even in normal `max_seq_len=2048` runs, about 0.5% of conversations are fully masked and are minor dead weight
2) When packing `rows`, first select the longest conversation that fits "naturally" in 513.
    - as mentioned above, this biases the early epoch towards shorter conversations, as they exhaust it biases towards longer, trimmed conversations.
3) When no "natural" conversations are left, pick and trim a longer one - this means all conversations in `conv_buffer` can be consumed, and no clogging will occur

I did a bit of data exploration, here is a table at `max_seq_len=512`, showing % of supervised training tokens:

| task | A) "natural" | B) long conv, preserved | C) long conv, lost | D) total |
|---|---|---|---|---|
| SmolTalk | 24M (8.2%) | 117M (39.9%) | 152M (51.9%) | 293M |
| MMLU | 0.2M (84.0%) | 0.0M (0.1%) | 0.04M (15.9%) | 0.2M |
| GSM8K | 0.8M (100%) | 0 | 0 | 0.8M |
| weighted mixture | 28M (9.3%) | 117M (39.4%) | 152M (51.3%) | 297M |

(Supervised tokens; mixture includes 3x/4x repeats; A+B+C=D)

The problem is SmolTalk. Only 8.16% of supervised training tokens appear in conversations that are shorter than 513 and fit "naturally". 39.90% of training tokens appear in longer conversations before the cut `conv[:513]`. The remaining 51.93% are lost because they appear after the cut `conv[513:]`.

Hard-cutting conversations at 513 has an issue of cutting through an assistant turn (see image attached). Out of "long" SmolTalk conversations, ~11% hit a case where the cut appears in the first assistant turn, so there is no `<|assistant_end|>`, which is not perfect.

I have also considered an alternative cutting mechanism: "keep as many user-assistant pairs as possible w/o crossing the 513 limit" - but it reduces SmolTalk supervised training tokens from above B) 117M to B') 77M (reducing total 24M+117M->24M+77M, ~29%)

Attachment:

image

FumingChang:
I hit the fully masked batch issue with the `runcpu.sh` setup (`max_seq_len=512`, d6, MPS). A guard showed zero valid targets at SFT step 3.

As a local experiment, I changed `refill_buffer()` to skip conversations with `len(ids) > row_capacity` or no supervised targets (`not any(mask[1:])`). This let my configured 1,500-iteration SFT run finish without NaNs. Evaluating checkpoint step 1499 on 100 questions per task gave ARC-Easy 33/100, ARC-Challenge 24/100, and MMLU 30/100. The model also did not reliably answer “What is the capital of France?” in my informal chat test.

I see that skipping fully masked conversations has already been proposed above. My additional experiment skips oversized conversations rather than truncating them. **Given the analysis that only about 8.2% of SmolTalk supervised tokens occur in conversations that naturally fit within 513 tokens, is losing the longer conversations the main reason to prefer fallback truncation?** I cannot attribute my model’s weak answers to this filtering without a controlled comparison against the same base checkpoint.

PR #487, opened 2026-02-01
Support nanochat as a dependency for other projects

francip:
# Summary

Enable nanochat to be used as a Python dependency in external projects.

# Problem

The `pyproject.toml` was missing the `[build-system] `section, which is required for pip to recognize and install the package. Without it, running pip install on nanochat (or adding it as a dependency) would fail because pip didn't know how to build the package.

# Changes

- Added [`build-system]` and [tool.setuptools] sections to pyproject.toml
- Added `*.egg-info/` to `.gitignore`

# Test Plan

Verified nanochat works as a dependency by adding the following to another project:

```
dependencies = [
    "nanochat",
]

[tool.uv.sources]
nanochat = { path = "../nanochat", editable = true }
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ __pycache__/
 dev-ignore/
 report.md
 eval_bundle/
+*.egg-info/
 
 # Secrets
 .env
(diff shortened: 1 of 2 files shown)


francip:
Ugh, wrong branch.

PR #489, opened 2026-02-01
Fix rotation calculations in apply_rotary_emb function

qdrk:
Reading some blogs on RoPE, it feels like the current implementation is a little off? Or am I missing something?

### DIFF
--- a/nanochat/gpt.py
+++ b/nanochat/gpt.py
@@ -52,8 +52,8 @@ def apply_rotary_emb(x, cos, sin):
     assert x.ndim == 4  # multihead attention
     d = x.shape[3] // 2
     x1, x2 = x[..., :d], x[..., d:] # split up last dim into two halves
-    y1 = x1 * cos + x2 * sin # rotate pairs of dims
-    y2 = x1 * (-sin) + x2 * cos
+    y1 = x1 * cos - x2 * sin # rotate pairs of dims
+    y2 = x1 * sin + x2 * cos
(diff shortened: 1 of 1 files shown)


DaliVana:
There is not much of a difference. Original will rotate clockwise and changed one rotate counter-clockwise.

svlandeg (approved):
Thanks for the PR @qdrk! I do think you're right that usually a slightly different formula is used, as the standard rotation matrix would be
```
[cos(θ)   -sin(θ)]
[sin(θ)    cos(θ)]
```
i.e. corresponding to the edits you made.

It would probably be nice to stick to the most frequent implementation/explanation of this, especially for education purposes.

qdrk:
@karpathy let me know if you have any feedback, otherwise I'd be pretty excited to get this merged :D

karpathy:
Thanks — you're right that this is the transpose of the textbook rotation matrix, but the two are exactly equivalent conventions rather than a bug: attention only sees the *relative* rotation R(θn−θm), and negating it is a sign flip on half the head channels that gets absorbed into the learned W_q/W_k. So there's no functional gain, while merging would silently break every existing trained checkpoint (the relative rotation they were trained under would be reversed at load time).

Given that trade, I landed a two-line comment instead (3616a02) noting the convention and why it stays, so future readers don't trip on the sign. Closing, but thanks for flagging it — the confusion was real and is now documented.

PR #496, opened 2026-02-04
Docs: add NorMuon paper link near second-order momentum implementation

zichongli5:
Hi! Thanks for the awesome work — it’s been super helpful to read and experiment with.

This PR adds a reference to the NorMuon paper at the spot where the second-order momentum is introduced on top of Muon. Since this is a specific design choice described in the paper, having the link there may help readers quickly find the original motivation and details.

If you’d prefer a different location or wording, I’m happy to adjust!

### DIFF
--- a/nanochat/optim.py
+++ b/nanochat/optim.py
@@ -243,6 +243,7 @@ class MuonAdamW(torch.optim.Optimizer):
         momentum_buffer = state["momentum_buffer"]
 
         # Second momentum buffer is factored, either per-row or per-column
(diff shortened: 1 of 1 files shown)


svlandeg:
Hi! Thanks for the note. The paper is now mentioned and linked in the dev log & `nanochat/optim.py` (cf [this edit](https://github.com/karpathy/nanochat/commit/718e5e9d67be7e92bd158f5471e1ce53de2e6a64)) so I think we can close this one 🙂

Issue #435, opened 2026-01-12
Proposal: Merge training shell scripts


jurca:
The repository currently contains two training shell scripts - `speedrun.sh` and `run1000.sh`. The logic is pretty much the same, with `speedrun.sh` containing more detailed comments and one less bug (see below), and the `run1000.sh` starting the web UI at the end.

I propose merging these two scripts into a single script configurable via environment variables or options with sensible defaults and a `--speedrun` option (alias to `--depth=20`) to configure those for the, well, speedrun configuration.

Thus, we will have a single well-documented script and the absence of logic duplication would make mistakes like [`run1000.sh` never waiting for the pretraining dataset being downloaded](https://github.com/karpathy/nanochat/blob/master/run1000.sh#L24) harder to make.

Am am willing to do the work and submit a PR - if there is interest, of course, so I'm asking first.

While at it, I have a few ideas for what to do next (in a separate PR(s)?):
- add opt-out for dependency installation
- add opt-out for dataset download
- add opt-out for training the tokenizer (or if there already is a trained tokenizer)
- add `set -eu -o pipefail` to the beginning of the script to terminate the script if any command or command in a pipeline fails or an unset variable is used (for improved DX)

I find the first three additions listed above useful when iterating in the same environment and training multiple models.

Would you be interested, please?

jayvenn21:
I’d be interested in helping here. However, I’d aim for a very minimal change.

Specifically, I’m thinking of eliminating logic duplication by making
`run1000.sh` a thin wrapper around `speedrun.sh`, keeping `speedrun.sh`
as the canonical reference and preserving existing behavior.

If that direction sounds reasonable, I can put together a small PR.

svlandeg:
Hi! `run1000.sh` has been removed, and `speedrun.sh` should now be the single source of truth to replicate the latest best entry on the [leaderboard](https://github.com/karpathy/nanochat?#leaderboard). So I'll go ahead and close this one 🙏

jurca:
Hello, perfectly understood 👍!

However, I would like to point out that the `runs/speedrun.sh` and `runs/runcpu.sh`  were still very similar, so I went ahead a submitted a PR merging the logic of these two, starting afresh from the current codebase: #508 **Unify & generalize training script logic**. 🙏

PR #508, opened 2026-02-06
Unify & generalize training script logic

jurca:
Merge the logic shared by the `runs/speedrun.sh` and `runs/runcpu.sh` scripts into a shared configurable and extensible common script `runs/train.sh`, which can also be used by itself and is configurable via environment variables.

The `train.sh` script provides reasonable defaults (almost exactly matching the speedrun script) and places all of configuration variables in one place at the top of the file for ease of editing/reference/overriding, along with explanatory comments.

This unification reduces the risk of diverging code paths with future modifications.

The code was tested on a single-node 8×H100 (PCIe) and seems to work correctly.

AI disclosure: All edits were done by hand and Github Copilot inline suggestions. Copilot Agent was used for code review before running the code.

See also: #435 Proposal: Merge training shell scripts (original proposal, code has shifted since the original idea).

### DIFF
--- a/runs/runcpu.sh
+++ b/runs/runcpu.sh
@@ -10,50 +10,22 @@
 # Think of this run as educational/fun demo, not something you should expect to work well.
 # You may also want to run this script manually and one by one, copy pasting commands into your terminal.
 
-# all the setup stuff
-export NANOCHAT_BASE_DIR="$HOME/.cache/nanochat"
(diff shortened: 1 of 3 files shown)


svlandeg:
Hi! I double checked with Andrej and he indeed prefers not to have a monolithic train script for now, but rather the ones we have right now that allow for easy/quick copy pasting of individual statements. So I'll go ahead and close this one.

PR #510, opened 2026-02-06
Fix comment in speedrun's base_train to match the actual code

svlandeg:
Follow-up to [this edit](https://github.com/karpathy/nanochat/commit/542beb0c8c175af2d52ec7065345dcd8f0162368) which didn't update the comment 😉


karpathy:
oops i just fixed this and only now found this in PRs

PR #515, opened 2026-02-08
fix/train-loss-grad-accum

Maxkrvo:
Previously train_loss was overwriting each micro step, so only the last micro batch's loss was logged. With grad_accum_steps=4 (default speedrun), this discarded 3/4 of the losses and made the EMA smoothed training curve noisier than necessary.

### DIFF
--- a/scripts/base_train.py
+++ b/scripts/base_train.py
@@ -485,13 +485,15 @@ while True:
     # evaluate the gradient
     synchronize()
     t0 = time.time()
+    train_loss = 0.0
     for micro_step in range(grad_accum_steps):
         with autocast_ctx:
             loss = model(x, y)
-        train_loss = loss.detach() # for logging
+        train_loss += loss.detach() # accumulate for logging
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
Hi, thanks for the PR! Even when averaged we'd still only be seeing the loss from a single rank. Judging from [this comment](https://github.com/karpathy/nanochat/pull/52#issuecomment-3403872860), I expect overwriting & only recording the final micro loss is actually a deliberate choice in `base_train.py` as well, to avoid bloat just for logging purposes.

Maxkrvo:
> Hi, thanks for the PR! Even when averaged we'd still only be seeing the loss from a single rank. Judging from [this comment](https://github.com/karpathy/nanochat/pull/52#issuecomment-3403872860), I expect overwriting & only recording the final micro loss is actually a deliberate choice in `base_train.py` as well, to avoid bloat just for logging purposes.

Hey svnlandeg thanks for the response, I wasn't aware this was the case. I appreciate the feedback!

PR #516, opened 2026-02-08
Speedrun Leaderboard: d26 0.5M batch, ratio 7.25 — 2.49h (9.6% faster than Run 3)

imxj:
### Summary

This run reverts the total batch size for d26 from 1M (introduced in Run 3) back to 0.5M (524,288 tokens), while reducing the param-data ratio from 8.25 to 7.25. The result is a **9.6% speedup** over Run 3, achieving GPT-2 capability in **2.49 hours** (8967s) with a CORE score of 0.2626.

### Launch Command

```bash
OMP_NUM_THREADS=1 torchrun --standalone --nproc_per_node=8 -m scripts.base_train -- \
    --depth=26 \
    --run="d26_r725_05M" \
    --model-tag="d26_r725_05M" \
    --device-batch-size=16 \
    --total-batch-size=524288 \
    --sample-every=-1 \
    --save-every=-1 \
    --core-metric-every=999999 \
    --target-param-data-ratio=7.25 \
    --fp8
```

### Result

```
core_metric 0.2626
step 12700
total_training_time 8967
Minimum validation bpb: 0.750008
```

**Reproducibility**: ran this configuration twice on an 8×H100 node:
- Run 1: CORE 0.2729, time 8985s, val_bpb 0.750751 ✅
- Run 2: CORE 0.2626, time 8967s, val_bpb 0.750008 ✅

Both comfortably clear the GPT-2 CORE threshold of 0.256525 (margin of +0.006 to +0.016).

### What Changed and Why

**The key insight: Run 3's auto-computed 1M batch size for d26 is suboptimal for the speedrun objective.**

Run 3 introduced auto batch size scaling that computes 1M as optimal for d26. This is correct for *compute-optimal* training, but "compute-optimal" and "speedrun-optimal" are different objectives:

1. **More steps > more tokens in the undertraining regime**: At 1M batch, d26 trains for only 7,226 steps. At 0.5M with ratio 7.25, we get 12,700 steps — fewer total tokens but more optimization steps. The model benefits more from additional gradient updates than from seeing more data when we're already undertraining.

2. **Direct evidence**: At ratio 7.5, the 1M batch fails CORE (0.2534) while 0.5M passes (0.2577), despite training on the same total tokens. This directly demonstrates the advantage of smaller batch size at these training horizons.

3. **Wall-clock savings**: 0.5M batch needs only 2× gradient accumulation vs 4× for 1M, and we train on 40% fewer total tokens (ratio 7.25 vs 8.25) while achieving a stronger CORE score.

### Experimental Evidence

Full ratio sweep at 0.5M batch:

| Ratio | Time (s) | CORE | Pass? | Notes |
|-------|----------|------|-------|-------|
| 7.0 | 8709 | 0.2348 | ❌ | Too undertrained |
| 7.125 | 8804-8875 | 0.2406-0.2571 | ❌ | 1/3 runs pass — unreliable |
| 7.15 | 8895 | 0.2594 | ✅ | Single run — needs confirmation |
| 7.2 | 8907 | 0.2537 | ❌ | |
| **7.25** | **8967-8985** | **0.2626-0.2729** | **✅** | **2/2 runs — reliable** |
| 7.5 | 9270-9275 | 0.2577-0.2595 | ✅ | Safe but slower |

For comparison, Run 3 at 1M batch: ratio 8.25 → CORE 0.2602, time 9922s.

**A note on CORE variance**: CORE scores exhibit run-to-run variance of approximately ±0.01-0.02 at this scale. For example, r7.125 produced CORE values of 0.2571, 0.2472, and 0.2406 across three runs with identical configuration. Similarly, r7.15 passes (0.2594) while r7.2 fails (0.2537), which is non-monotonic and attributable to evaluation noise. As noted in the LEADERBOARD docs, "the CORE metric can be a little bit noisy." We chose r7.25 for its reproducibility — it clears the threshold with enough margin to be robust against this variance.

### Code Changes

Minimal changes (3 files):
- `README.md`: Add Run 4 to leaderboard table
- `runs/speedrun.sh`: Update launch command (add `--total-batch-size=524288`, change ratio to 7.25)
- `dev/LEADERBOARD.md`: Add Run 4 documentation

### AI Disclosure

Experimental design and hyperparameter search were conducted using Claude Code. The author understands and takes responsibility for all claims and results.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Presently, the main focus of development is on tuning the pretraining stage, whi
 | 1 | 3.04 | 0.74833 | 0.2585 | d24 baseline, slightly overtrained | Jan 29 2026 | 348fbb3 | @karpathy |
 | 2 | 2.91 | 0.74504 | 0.2578 | d26 slightly undertrained **+fp8** | Feb 2 2026 | a67eba3 | @karpathy |
 | 3 | 2.76 | 0.74645 | 0.2602 | bump total batch size to 1M tokens | Feb 5 2026 | 2c062aa | @karpathy |
+| 4 | 2.49 | 0.75001 | 0.2626 | d26 0.5M batch, ratio 7.25 | Feb 8 2026 | TBD | @imxj |
 
 The primary metric we care about is "time to GPT-2" - the wall clock time needed to outperform the GPT-2 (1.6B) CORE metric on an 8XH100 GPU node. The GPT-2 CORE score is 0.256525. In 2019, the training of GPT-2 cost approximately $43,000 so it is incredible that due to many advances over 7 years across the stack, we can now do so much faster and for well below $100 (e.g. at the current ~$3/GPU/hr, an 8XH100 node is ~$24/hr, so 3 hours is ~$72).
 
--- a/dev/LEADERBOARD.md
+++ b/dev/LEADERBOARD.md
@@ -147,3 +147,41 @@ Minimum validation bpb: 0.74645
 ```
 
 The big change here is that the batch size was doubled from 0.5M to 1M, which works better for a d26 model and allowed me to decrease the number of optimization steps a bit via `--target-param-data-ratio` from 8.5 to 8.25. The TLDR is that the original batch size of 0.5M was tuned for d12, but bigger models (e.g. d26) prefer larger total batch size. I determined in experiments that d26 prefers 1M. Then I implemented and merged a principled way to calculate the optimal batch size given depth so that all nanochat models of all depths benefit. See [dev/LOG.md](dev/LOG.md) entry "2026-02-05: Auto Batch Size Scaling" for more detail.
+
+## Run 4
+
+Achieved Feb 8 2026. Launch command:
+
+```
+OMP_NUM_THREADS=1 torchrun --standalone --nproc_per_node=8 -m scripts.base_train -- \
+    --depth=26 \
+    --run="d26_r725_05M" \
+    --model-tag="d26_r725_05M" \
+    --device-batch-size=16 \
+    --total-batch-size=524288 \
+    --sample-every=-1 \
+    --save-every=-1 \
+    --core-metric-max-per-task=-1 \
+    --core-metric-every=999999 \
+    --target-param-data-ratio=7.25 \
+    --fp8
+```
+
+Result:
+
+```
+core_metric 0.2626
+step 12700
+total_training_time 8967
+Minimum validation bpb: 0.750008
+```
+
+Reproduced twice: Run 1 CORE 0.2729 (8985s), Run 2 CORE 0.2626 (8967s). Both comfortably clear the 0.256525 threshold.
+
+The key change is reverting the total batch size for d26 from 1M back to 0.5M (524,288 tokens) and lowering the param-data ratio from 8.25 to 7.25. While Run 3 showed that the auto-computed 1M batch is optimal for *compute-optimal* training of d26, "compute-optimal" and "speedrun-optimal" are different objectives. In the speedrun's undertraining regime, the model benefits more from additional optimization steps at smaller batch size (12,700 steps at 0.5M) than from seeing more data at larger batch size (7,226 steps at 1M). Direct evidence: at ratio 7.5, the 1M batch fails CORE (0.2534) while 0.5M passes (0.2577), despite training on the same total tokens.
+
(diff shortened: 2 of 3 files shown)


karpathy:
sounded a bit like free lunch so i tried to reproduce (on master, identical launch), and I got:

```
wandb: Run summary:
wandb:          core_metric 0.23613
wandb:                 step 12700
...
Minimum validation bpb: 0.750413
```

i.e.  the core metric is quite a bit lower than needed. The validation loss is about similar to yours. It's possible that CORE is that noisy? :\

imxj:
Thanks for trying it out.

My earlier results were from a single VM (vm1, 8H100). I’ve just rerun the same repro on both vm1 and a new VM (vm2, also 8H100). Here are the full r7.25 results:

Run | Machine | CORE | Val BPB | Time (s) | Pass?
-- | -- | -- | -- | -- | --
1 | VM1 | 0.2729 | 0.7501 | 8985 | ✓
2 | VM1 | 0.2626 | 0.7500 | 8967 | ✓
3 | VM1 | 0.2577 | 0.7500 | 8965 | ✓
4 | VM2 | 0.2395 | 0.7504 | 9032 | ✗
5 | VM2 | 0.2614 | 0.7502 | 9031 | ✓ 

Your result (0.2361) is very close to Run 4 (0.2395), which suggests there may be a machine dependent factor (such as different H100 binning?). I have another run going on VM2 to see if it consistently scores lower. Will update with that result.

Meanwhile val_BPB is basically identical across all runs (~0.750), so the model training quality seems stable and it’s the CORE eval that is fairly noisy. 

One observation: VM1 has newer VBIOS Version than VM2: 96.00.A5.00.01 (VM1) vs 96.00.89.00.01 (VM2). 

Update (02/10): run 5 is completed. This time it passed the core threshold with 0.2614. It seems the variance is non-negligible.

chrisjmccormick:
That's super helpful data!! I did some similar analysis down at `d12` on an 8xA100 instance:

|  Run |  CORE | vs. mean |    BPB    | vs. mean   | Training Time | vs. mean |
|:----:|:-----:|----------|:---------:|------------|:-------------:|----------|
|    1 | 11.88 |    -0.15 | 0.917 768 |  0.000 067 |        567.80 |   -0.470 |
|    2 | 12.06 |     0.03 | 0.917 703 |  0.000 001 |        568.09 |   -0.185 |
|    3 | 11.15 |    -0.88 | 0.917 746 |  0.000 044 |        568.26 |   -0.011 |
|    4 | 12.45 |     0.42 | 0.917 566 | -0.000 135 |        568.59 |    0.316 |
|    5 | 12.27 |     0.24 | 0.917 840 |  0.000 138 |        568.25 |   -0.021 |
|    6 | 11.45 |    -0.57 | 0.917 739 |  0.000 038 |        568.16 |   -0.110 |
|    7 | 12.59 |     0.57 | 0.917 625 | -0.000 076 |        568.41 |    0.135 |
|    8 | 12.35 |     0.32 | 0.917 624 | -0.000 078 |        568.62 |    0.348 |
| Mean | 12.03 |          | 0.917 701 |            |       568.271 |          |
|  Std |  0.50 |          | 0.000 090 |            |         0.269 |          |

It's nice to see how stable the validation loss is. 
Since it's been well-proven that a model with `0.750` bpb can outperform GPT-2 on CORE (with the right seed), maybe the validation loss could be the target instead?

chrisjmccormick:
I noticed today that the validation set is dependent on the number of shards you download--the script chooses whatever the last shard is to be the validation set.

I don't think the bpb is nearly as stable across shards, so that could potentially explain the difference in loss that the two of you are finding, if either of you have something other than 370 shards downloaded?

imxj:
Thanks, good catch up. I checked that both my VMs have exactly 370 shards downloaded (shard_00000 through shard_00369), and I verified the md5 checksums match across all of them. So the validation shard should be consistent.

karpathy:
> I noticed today that the validation set is dependent on the number of shards you download--the script chooses whatever the last shard is to be the validation set.
> 
> I don't think the bpb is nearly as stable across shards, so that could potentially explain the difference in loss that the two of you are finding, if either of you have something other than 370 shards downloaded?

Ahhhh, good point. Damn.

karpathy:
I'm going to close this PR because I don't think that just reducing iterations is the right way to push the performance. It's possible to luck out with the CORE score in this way sadly. It's possible we'd want to use val_bpb instead, but that requires fixing the bug (make val loss stable) and in addition, doing so will mean we can't compare models with different datasets or exotic tokenizations (but that's ok, these are rare and we can fall back to CORE for that).

Issue #514, opened 2026-02-08
[BUG] AssertionError in gpt.py: Evaluation prompts exceed static RoPE cache (seq_len=256)


suraj-self:
### Summary
The evaluation pipeline crashes with an AssertionError when the model is trained with a small context window (e.g., --max-seq-len=256). This is caused by the static initialization of the Rotary Positional Embedding (RoPE) cache, which does not account for the significantly longer prompts generated by multi-shot (few-shot) evaluation tasks.

### Context
In nanochat/gpt.py, the RoPE cache is initialized using a fixed $10\times$ multiplier
```
# nanochat/gpt.py line 182
self.rotary_seq_len = config.sequence_len * 10 # 10X over-compute should be enough, TODO make nicer?
```
When ```max_seq_len``` is **256**, the cache is capped at **2560**. However, evaluation tasks (like ARC, HellaSwag, or BIG-bench) often use 5-10 shot prompts. These prompts frequently exceed 2560 tokens, triggering the safety assertion in the ```forward``` pass

### Error Log
```
step 00460/00462 (99.57%) | loss: 3.949577 | lrm: 0.01 | dt: 2587.06ms | tok/sec: 101,328 | mfu: 0.00 | epoch: 1 | total time: 27.67m | eta: 0.1m
step 00461/00462 (99.78%) | loss: 3.972177 | lrm: 0.00 | dt: 2579.98ms | tok/sec: 101,606 | mfu: 0.00 | epoch: 1 | total time: 27.71m | eta: 0.1m
Step 00462 | Validation bpb: 1.210319
Evaluating: hellaswag_zeroshot (0-shot, type: multiple_choice)... accuracy: 0.2900 | centered: 0.0533 | time: 4.97s
Evaluating: jeopardy (10-shot, type: language_modeling)... accuracy: 0.0000 | centered: 0.0000 | time: 5.10s
Evaluating: bigbench_qa_wikidata (10-shot, type: language_modeling)... accuracy: 0.0000 | centered: 0.0000 | time: 3.85s
Evaluating: arc_easy (10-shot, type: multiple_choice)... accuracy: 0.3480 | centered: 0.1307 | time: 8.14s
Evaluating: arc_challenge (10-shot, type: multiple_choice)... accuracy: 0.2360 | centered: -0.0187 | time: 8.85s
Evaluating: copa (0-shot, type: multiple_choice)... accuracy: 0.4800 | centered: -0.0400 | time: 0.66s
[...]

PR #520, opened 2026-02-10
Fix RoPE cache overflow for long prompts with KV cache

dipeshbabu:
When using KV cache during generation/inference, RoPE cos/sin buffers are sliced with an absolute offset (`T0 = kv_cache.get_pos()`), but the code only validated `T <= cache_len`. This can crash once `T0 + T` exceeds the cached RoPE length (even when `T` is small), matching the failure reported in #514.

Root cause:
RoPE cache bounds were validated against `T` instead of `T0 + T`, so long contexts / long generation runs can hit an out-of-bounds slice:
`cos[:, T0:T0+T]`, `sin[:, T0:T0+T]`.

Change:
- Ensure the RoPE cache covers `T0 + T` before slicing.
- Grow the cache to the next power of two for amortized behavior.
- Overwrite existing buffers instead of re-registering them.

### DIFF
--- a/nanochat/gpt.py
+++ b/nanochat/gpt.py
@@ -179,8 +179,7 @@ class GPT(nn.Module):
         # Precompute a reasonably large RoPE cache up front (cheap relative to model weights).
         # The cache may also grow lazily in forward() if generation exceeds this length.
         self.rotary_seq_len = config.sequence_len * 10
-        # Bound lazy growth to avoid unbounded memory usage during very long generation runs.
-        self.max_rotary_seq_len = max(self.rotary_seq_len, config.sequence_len * 64)
+        self.max_rotary_seq_len = self.rotary_seq_len
         
         head_dim = config.n_embd // config.n_head
         cos, sin = self._precompute_rotary_embeddings(self.rotary_seq_len, head_dim)


dipeshbabu:
@karpathy what do you think about this? could you review it?

nanochat/gpt.py
```diff
@@ -387,14 +428,16 @@ def setup_optimizer(self, unembedding_lr=0.004, embedding_lr=0.2, matrix_lr=0.02
...
-        # if kv cache exists, we need to offset the rotary embeddings to the current position in the cache
         T0 = 0 if kv_cache is None else kv_cache.get_pos()
-        cos_sin = self.cos[:, T0:T0+T], self.sin[:, T0:T0+T] # truncate cache to current sequence length
+
+        # Ensure RoPE buffers cover absolute positions [T0, T0+T)
+        self._ensure_rope_cache(T0 + T, device=idx.device)
+
+        # Now it's safe to slice
+        assert idx.device == self.cos.device, f"RoPE buffers and idx device mismatch: {idx.device} != {self.cos.device}"
+        assert self.cos.dtype == torch.bfloat16, "RoPE buffers must be bfloat16"
```
svlandeg: restoring original messages, to avoid additional edits and make the diff easier to review
```suggestion
        assert idx.device == self.cos.device, f"Rotary embeddings and idx are on different devices: {idx.device} != {self.cos.device}"
        assert self.cos.dtype == torch.bfloat16, "Rotary embeddings must be in bfloat16"
```

nanochat/gpt.py
svlandeg: This comment should be updated, for consistency

nanochat/gpt.py
```diff
@@ -185,6 +185,47 @@ def __init__(self, config, pad_vocab_size_to=64):
...
+        try:
+            import torch._dynamo
+            if torch._dynamo.is_compiling():
+                raise RuntimeError(
+                    f"RoPE cache too small during torch.compile (need {needed_seq_len}, have {cur_len}). "
+                    f"Increase initial rotary_seq_len or disable compile for generation."
+                )
+        except Exception:
+            # torch._dynamo may not exist in older torch; ignore.
+            pass
```
svlandeg: This doesn't make sense: the `RuntimeError` will immediately be caught & silenced by the blanket `except Exception`. 

Which "older" Torch versions wouldn't support this? Because `nanochat` assumes Torch 2.9.1.

nanochat/gpt.py
```diff
@@ -185,6 +185,47 @@ def __init__(self, config, pad_vocab_size_to=64):
...
+
+        head_dim = self.config.n_embd // self.config.n_head
+        cos, sin = self._precompute_rotary_embeddings(
+            seq_len=new_len, head_dim=head_dim, device=device)
+
+        # Preserve dtype/device invariants (precompute returns bf16 already)
+        cos = cos.to(dtype=self.cos.dtype, device=device)
+        sin = sin.to(dtype=self.sin.dtype, device=device)
+
+        # Overwrite existing registered buffers (no re-register)
```
svlandeg: Why not? (they already are declared as `persistent=False`)

nanochat/gpt.py
dipeshbabu: You're right. The `RuntimeError` would be swallowed by the blanket except. I removed the `torch._dynamo` `try/except` block entirely to keep the fix minimal and correct.

nanochat/gpt.py
dipeshbabu: Agreed. I updated the comment to reflect that the RoPE cache now grows dynamically, so the assert fail if we ever reach that amount note was stale.

nanochat/gpt.py
dipeshbabu: Fair point. I’m not avoiding `register_buffer` for persistence reasons (`persistent=False` is already correct). I kept simple assignment (`self.cos = cos`, `self.sin = sin`) to overwrite the existing buffers in place without re-registering names. I’ll clarify that in the comment or remove the comment if it’s more distracting than helpful.

dipeshbabu:
> Is unbounded growth a good idea?

Unbounded growth is not ideal long-term. It fixes the crash, but without a cap it can silently keep allocating memory during very long generation.

dipeshbabu:
@svlandeg I updated the stale RoPE cache comment, removed the problematic blanket exception pattern, and changed the lazy RoPE cache growth to be bounded `max_rotary_seq_len` to avoid unbounded memory growth in long generation runs. I also simplified `_ensure_rope_cache()` to infer device from the existing buffer and kept the `forward()` fix based on `T0 + T`.

svlandeg:
@dipeshbabu:

Thanks!  Per the contribution guidelines, can you please declare any parts that had substantial LLM contribution, and whether there are any parts that you have not written and that you do not fully understand?

dipeshbabu:
> Hi @dipeshbabu, as per the contribution guidelines, can you please declare any parts that had substantial LLM contribution, and whether there are any parts that you have not written and that you do not fully understand?

I used an LLM in a limited way to help me think through and explain the RoPE/KV-cache indexing issue more clearly specifically around why the cache bound needs to be checked against T0 + T during cached generation, not just T. It was used for explanation/reasoning support and wording, not for code I copied directly.

I wrote the final code changes myself and I understand the fix and the affected code path.

nanochat/gpt.py
```diff
@@ -176,15 +176,61 @@ def __init__(self, config, pad_vocab_size_to=64):
...
         # To support meta device initialization, we init the rotary embeddings here, but it's just "fake" meta tensors only.
-        # As for rotary_seq_len, these rotary embeddings are pretty small/cheap in memory,
-        # so let's just over-compute them by 10X, but assert fail if we ever reach that amount.
-        # In the future we can dynamically grow the cache, for now it's fine.
-        self.rotary_seq_len = config.sequence_len * 10 # 10X over-compute should be enough, TODO make nicer?
+        # Precompute a reasonably large RoPE cache up front (cheap relative to model weights).
+        # The cache may also grow lazily in forward() if generation exceeds this length.
+        self.rotary_seq_len = config.sequence_len * 10
+        # Bound lazy growth to avoid unbounded memory usage during very long generation runs.
+        self.max_rotary_seq_len = max(self.rotary_seq_len, config.sequence_len * 64)
```
svlandeg: I don't like the fact that now there's one more magic number...

svlandeg (commented):
Thanks for the quick follow-up. To be perfectly honest with you, there were a little bit too many small errors / issues with the original implementation, which felt like vibe-coded to me. When I asked Claude, it gave a very similar implementation, code structure & comments.

I think it's good to look into this, as there currently is a "future TODO" in the code on `master` to dynamically grow the cache when we reach the limit here, and as #514 demonstrated this happens with `max_seq_len` at 256, which I can replicate. 

But maybe this is something that Andrej should look into himself...

nanochat/gpt.py
dipeshbabu: Totally fair. I’ll drop the extra factor and set `max_rotary_seq_len = self.rotary_seq_len` so there’s no additional magic number. The fix still holds; it just won’t grow beyond the preallocated cache.

PR #536, opened 2026-02-16
Multi platform CI - Github Workflow tests v4

SermetPekin:
- Updated with the upstream. 
- fixed the test.yml file 
- It  [passes](https://github.com/SermetPekin/nanochat/actions/runs/22054626098) on both Linux and MacOs
- Here is the earlier version of this testing on GH actions [PR](https://github.com/karpathy/nanochat/pull/151)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,24 +29,13 @@ jobs:
 
     - name: Set up uv
       uses: astral-sh/setup-uv@v7
-    # Linux: Request 'cpu' extra to force the pytorch-cpu index (prevents libcublas crash)
     - name: Install dependencies (Linux)
-      if: runner.os == 'Linux'
       run: |
         uv sync --extra cpu
-        uv pip install torch torchvision --index-url https://download.pytorch.org/whl/cpu
-
-    # macOS: Standard sync 
-    - name: Install dependencies (macOS)
-      if: runner.os == 'macOS'
-      run: uv sync
-
     - name: Add nanochat to PYTHONPATH (Unix)
       run: |
         echo "PYTHONPATH=$(pwd):$PYTHONPATH" >> $GITHUB_ENV
-
     - name: Run pytest
       run: |
         uv run --extra cpu pytest tests/ -v 
-        
          


.github/workflows/test.yml
```diff
@@ -0,0 +1,47 @@
...
+      run: uv sync
+
+    - name: Add nanochat to PYTHONPATH (Unix)
+      run: |
+        echo "PYTHONPATH=$(pwd):$PYTHONPATH" >> $GITHUB_ENV
+
+    - name: Run pytest
+      run: |
+        # uv run pytest tests/
+        .venv/bin/pytest tests/
```
svlandeg: Why not use the original `uv run` statement?

.github/workflows/test.yml
```diff
@@ -0,0 +1,47 @@
...
+        uv pip install torch torchvision --index-url https://download.pytorch.org/whl/cpu
+
+    # macOS: Standard sync 
+    - name: Install dependencies (macOS)
+      if: runner.os == 'macOS'
+      run: uv sync
+
+    - name: Add nanochat to PYTHONPATH (Unix)
+      run: |
+        echo "PYTHONPATH=$(pwd):$PYTHONPATH" >> $GITHUB_ENV
```
svlandeg, line 37: @karpathy: I personally think it would be nicer to be able to run `uv pip install -e .` here, instead of messing with env vars. But if we want that then we'd have to add

```
[tool.setuptools]
packages = ["nanochat"]
```

to `pyproject.toml`.

.github/workflows/test.yml
```diff
@@ -0,0 +1,47 @@
+name: Test
+
+on:
+  push:
+  pull_request:
```
svlandeg, line 7:
```suggestion
on:
  push:
    branches:
      - master
  pull_request:
    branches:
      - master
```

.github/workflows/test.yml
SermetPekin: @svlandeg uv run ignores current build and tries to install again. Since we already have an environment ready with what we wanted IMO it is better to run that python executable.

svlandeg: I'm afraid I don't understand/follow that comment. `uv run` right after `uv sync` should just work as intended AFAIK. Do you have any reference docs or links to clarify your point?

SermetPekin: GH Actions were failing on Linux due to the reason that I guess GH Actions do not have cuda therefore it was a quick workaround that I implemented, However if we prefer `uv run` adding `--extra cpu` parameter to `uv run` also worked fine. I changed that line.

.github/workflows/test.yml
```diff
@@ -0,0 +1,52 @@
...
+    - name: Install dependencies (Linux)
+      if: runner.os == 'Linux'
+      run: |
+        uv sync --extra cpu
+        uv pip install torch torchvision --index-url https://download.pytorch.org/whl/cpu
+
+    # macOS: Standard sync 
+    - name: Install dependencies (macOS)
+      if: runner.os == 'macOS'
+      run: uv sync
```
svlandeg: Did a single `uv sync --extra cpu` not work? Because that should also install `torch` from `"https://download.pytorch.org/whl/cpu"`

.github/workflows/test.yml
svlandeg: Right. We should address such issues instead of trying to sweep them under the rug 🙂 It feels like the setup should further be improved so that `uv run` runs seamlessly.

SermetPekin: Not for the purpose of sweeping under the rug but I thought uv sync and uv run is not probably perfect. Cause they had side effects instead of complementing each other.

.github/workflows/test.yml
SermetPekin: - simplifed the workflow since `uv run` fix worked with --extra cpu

PR #543, opened 2026-02-19
MoE

karpathy:
Experimental. Probably won't merge.


dustinwloring1988:
@karpathy   I would love to see this in the master branch. I added this to my fork but allowed the uses to specify sparse or dense. I also updated the tokenizer to more closely match the Harmony style tokenizer.

PR #544, opened 2026-02-19
dataloader: reuse cropped remainders to reduce token waste ~35% → ~23%

handsomesun:
## Summary

When the BestFit-Crop algorithm crops a document to fill remaining row space, the leftover tokens are currently discarded (~35% of all source tokens). This PR puts the cropped remainder (with BOS prepended) back into the document buffer for future rows, reducing effective crop waste to ~23%.

## Changes

**`nanochat/dataloader.py`** (6 lines added in the crop branch):
- After cropping a document to fill a row, the leftover tokens are prepended with BOS and appended back to `doc_buffer`
- Only remainders with >1 token are reused (to avoid degenerate single-token fragments)
- Updated module docstring and function docstring to reflect new ~23% crop rate

**`tests/test_dataloader_remainder.py`** (new file):
- Simulation-based validation comparing original vs. improved packing
- Uses realistic document length distribution (lognormal, median ~800 tokens)
- Verifies source token reduction >5% and speedup >1.05x across multiple sequence lengths

## Validation Results (T=2048, 10K rows)

```
  Source tokens consumed (original):  134,522,086
  Source tokens consumed (improved):  113,867,766
  Source tokens saved:                 20,654,320
  Reduction in source consumption:         15.4%

  Data efficiency (original):  0.7636 (training / source)
  Data efficiency (improved):  0.9023 (training / source)
  Efficiency improvement:            18.2%

  Speedup factor:              1.1816x
  Equivalent time reduction:   15.4%

  Current d24 record:              3.04 hours
  Estimated with optimization:     2.57 hours
  Time saved:                      ~28 minutes
```

Results across sequence lengths:
```
     T |  Orig Source/Train |  Impr Source/Train | Reduction | Speedup
-----------------------------------------------------------------------
   512 |             1.5765 |             1.2316 |      21.9% |  1.2799x
  1024 |             1.3715 |             1.1440 |      16.6% |  1.1989x
  2048 |             1.3098 |             1.1083 |      15.4% |  1.1816x
  4096 |             1.2854 |             1.0947 |      14.8% |  1.1742x
```

## Design Decisions

1. **BOS prepended to remainders**: Each remainder starts a new "context window" in a different row, so it needs its own BOS token for the attention mechanism to work correctly. This costs 1 extra token per remainder but maintains the BOS-aligned property.

2. **Minimum remainder size > 1**: Remainders of just 1 token (the BOS itself after cropping) provide no useful content and would just add overhead.

3. **Append to end of buffer**: Remainders are appended to the end of `doc_buffer` rather than inserted at the front, giving fresh documents priority in the best-fit search. This avoids pathological cases where small remainders keep getting re-split.

## Compatibility

- No changes to the public API or function signatures
- No changes to the training loop (`base_train.py`)
- Resume/checkpoint behavior unchanged (same `state_dict` format)
- The optimization is always-on with no configuration needed

Ref: https://github.com/karpathy/nanochat/discussions/481

### DIFF
--- a/nanochat/dataloader.py
+++ b/nanochat/dataloader.py
@@ -5,12 +5,14 @@ BOS-aligned bestfit:
    - Every row starts with BOS token
    - Documents packed using best-fit algorithm to minimize cropping
    - When no document fits remaining space, crops a document to fill exactly
-   - 100% utilization (no padding), ~35% tokens cropped at T=2048
+   - Cropped remainders are reused in subsequent rows (with BOS prepended)
+   - 100% utilization (no padding), ~23% tokens cropped at T=2048
 
 Compared to the original tokenizing_distributed_data_loader:
-BOS-aligned loses ~35% of tokens to cropping, but ensures that
-there are fewer "confusing" tokens in the train/val batches as every token can
-now attend back to the BOS token and sees the full context of the document.
+BOS-aligned loses ~23% of tokens to cropping (down from ~35% without remainder
+reuse), but ensures that there are fewer "confusing" tokens in the train/val
+batches as every token can now attend back to the BOS token and sees the full
+context of the document.
 
 Fallback to the original if you have very limited data AND long documents:
 https://github.com/karpathy/nanochat/blob/3c3a3d7/nanochat/dataloader.py#L78-L117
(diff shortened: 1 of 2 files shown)


nanochat/dataloader.py
```diff
@@ -5,12 +5,14 @@
...
+   - 100% utilization (no padding), ~23% tokens cropped at T=2048
 
 Compared to the original tokenizing_distributed_data_loader:
-BOS-aligned loses ~35% of tokens to cropping, but ensures that
-there are fewer "confusing" tokens in the train/val batches as every token can
-now attend back to the BOS token and sees the full context of the document.
+BOS-aligned loses ~23% of tokens to cropping (down from ~35% without remainder
+reuse), but ensures that there are fewer "confusing" tokens in the train/val
+batches as every token can now attend back to the BOS token and sees the full
+context of the document.
```
svlandeg, line 15: This comment is kept, but I don't think it's still valid: the cropped/reused remainders don't have access any more to the full context of the document.

tests/test_dataloader_remainder.py
```diff
@@ -0,0 +1,315 @@
+"""
```
svlandeg, line 1: IMO, this file shouldn't be in `tests`, as it's not a genuine regression test that uses the `tokenizing_distributed_data_loader_with_state_bos_bestfit` and that would be able to flag potential regressions in future updates of the actual code in `dataloader.py`. The file is interesting to run some analyses and it's worthwhile mentioning some results in the PR description, but I'm not sure we need to keep it in the repo for posteriority. (I assume this was entirely vibe coded?)

nanochat/dataloader.py
```diff
@@ -148,6 +154,14 @@ def refill_buffer():
                     doc = doc_buffer.pop(shortest_idx)
                     row_buffer[row_idx, pos:pos + remaining] = torch.tensor(doc[:remaining], dtype=torch.long)
                     pos += remaining
+                    # Remainder reuse: if the cropped document has leftover tokens,
+                    # prepend BOS and put the remainder back into the buffer for future rows.
+                    # This reduces token waste from ~35% to ~23% by recycling content that
+                    # would otherwise be discarded.
+                    leftover = doc[remaining:]
+                    if len(leftover) > 1:  # only reuse if remainder is meaningful (>1 token)
```
svlandeg, line 162: Maybe this cutoff should be slightly higher to ensure we have at least `n` tokens (5? 10?)

Issue #541, opened 2026-02-18
make validation loss reproducible


karpathy:
See https://github.com/karpathy/nanochat/pull/516#issuecomment-3879991630

Current validation loss depends on how many shards someone happened to download because the validation shard is chosen as simply the last one. Will have to fix this by designating a specific shard to be val.

mvanhorn:
I've submitted a fix for this in PR #613. Pins the val shard to shard_06542 by explicit filename matching instead of relying on sort order.

svlandeg:
Closing this issue as it's resolved by this [recent edit](https://github.com/karpathy/nanochat/commit/324e69c45d3606095adb6b409078647145165454#diff-ae4e743c6699e9b5468bf7476fbfe1e36764b01a8f1ba612f940864fa2a91270) - we'll review #613 separately.

PR #545, opened 2026-02-19
dataset: pin validation shard 

KartikVashishta:
Addresses #541 Validation shard is fixed so val loss/bpb is stable regardless of how many shards were downloaded.

### DIFF
--- a/nanochat/dataset.py
+++ b/nanochat/dataset.py
@@ -143,9 +143,8 @@ if __name__ == "__main__":
 
     num = MAX_SHARD + 1 if args.num_files == -1 else min(args.num_files, MAX_SHARD + 1)
     ids_to_download = list(range(num))
-    if VAL_SHARD_INDEX not in ids_to_download:
+    if num <= VAL_SHARD_INDEX:
         ids_to_download.append(VAL_SHARD_INDEX)
-    ids_to_download = sorted(set(ids_to_download))
     
     if args.num_files != -1 and args.num_files <= MAX_SHARD:
         print(f"Downloading {len(ids_to_download)} shards ({num} train + 1 val) using {args.num_workers} workers...")


nanochat/dataset.py
```diff
@@ -23,6 +23,12 @@
 BASE_URL = "https://huggingface.co/datasets/karpathy/fineweb-edu-100b-shuffle/resolve/main"
 MAX_SHARD = 1822 # the last datashard is shard_01822.parquet
 index_to_filename = lambda index: f"shard_{index:05d}.parquet" # format of the filenames
+
+# Always use a fixed shard for val so that the metrics don't depend
+# on how many shards are downloaded
+VAL_SHARD_INDEX = MAX_SHARD
```
svlandeg: I personally wouldn't link these two variables to eachother, because the risk would be that in the future for some reason `MAX_SHARD` is changed into a different value without realising the consequences for the validation step. You may as well could do `VAL_SHARD_INDEX=1822`.

I think I'd personally just pick shard 0, which will always be downloaded anyway. This might reduce some if-else logic. The downside being that we won't be able to compare future evaluations with previous ones as soon as we'd merge this.

nanochat/dataloader.py
```diff
@@ -32,10 +32,9 @@ def _document_batches(split, resume_state_dict, tokenizer_batch_size):
     """
     ddp, ddp_rank, ddp_local_rank, ddp_world_size = get_dist_info()
 
-    parquet_paths = list_parquet_files()
-    assert len(parquet_paths) != 0, "No dataset parquet files found, did you run dataset.py?"
-    parquet_paths = parquet_paths[:-1] if split == "train" else parquet_paths[-1:]
 
+    parquet_paths = get_parquet_paths(split) 
+    assert len(parquet_paths) != 0, "No dataset parquet files found, did tiy run dataset.py?"
```
svlandeg:
```suggestion
    assert len(parquet_paths) != 0, "No dataset parquet files found, did you run dataset.py?"
```

nanochat/dataset.py
```diff
@@ -40,15 +46,32 @@ def list_parquet_files(data_dir=None):
...
+    data_dir = DATA_DIR if data_dir is None else data_dir
+
+    val_path = os.path.join(data_dir, VAL_SHARD_FILENAME)
+    if split == "val":
+        if not os.path.exists(val_path):
+            raise FileNotFoundError(f"Validation shard {VAL_SHARD_FILENAME} not found in {data_dir}")
+        return [val_path]
+
+    # train everything except the val shard (if it exists, exclude it if not, nothing to exclude)
+    all_paths = list_parquet_files(data_dir)
```
svlandeg: I think we can just merge these two functions and c/p the functionality of `list_parquet_files` here inline. That also allows you to prevent adding the val shard to the set instead of having to remove it again later on.

nanochat/dataset.py
KartikVashishta: good point - i've decoupled these. `VAL_SHARD_INDEX` is now pinned to 1822. I’m keeping it at shard_01822 specifically for backwards comparability.

nanochat/dataset.py
KartikVashishta: Agree with the intent. I kept list_parquet_files() as a small utility, but added exclude_filenames so get_parquet_paths("train") never includes the val shard in the first place

nanochat/dataset.py
```diff
@@ -23,32 +23,58 @@
...
+    val_path = os.path.join(data_dir, VAL_SHARD_FILENAME)
+    if split == "val":
+        if not os.path.exists(val_path):
+            raise FileNotFoundError(
+                f"Validation shard {VAL_SHARD_FILENAME} not found in {data_dir}. "
+                f"Run: python -m nanochat.dataset -n <N> (downloads the val shard too)."
+            )
+        return [val_path]
+    else:
+        #train split: list files while excluding val (don't add then remove).
```
svlandeg:
```suggestion
        # train split: list files while excluding val
```

nanochat/dataset.py
```diff
@@ -111,13 +137,21 @@ def download_single_file(index):
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="Download FineWeb-Edu 100BT dataset shards")
-    parser.add_argument("-n", "--num-files", type=int, default=-1, help="Number of shards to download (default: -1), -1 = disable")
+    parser.add_argument("-n", "--num-files", type=int, default=-1, help="Number of training shards to download (default: -1 = all). Validation shard is always included")
```
svlandeg: I would remove this comment because it may actually be confusing to users
```suggestion
    parser.add_argument("-n", "--num-files", type=int, default=-1, help="Number of training shards to download (default: -1 = all).")
```

nanochat/dataset.py
```diff
@@ -111,13 +137,21 @@ def download_single_file(index):
...
     parser.add_argument("-w", "--num-workers", type=int, default=4, help="Number of parallel download workers (default: 4)")
     args = parser.parse_args()
 
     num = MAX_SHARD + 1 if args.num_files == -1 else min(args.num_files, MAX_SHARD + 1)
     ids_to_download = list(range(num))
-    print(f"Downloading {len(ids_to_download)} shards using {args.num_workers} workers...")
+    if VAL_SHARD_INDEX not in ids_to_download:
+        ids_to_download.append(VAL_SHARD_INDEX)
+    ids_to_download = sorted(set(ids_to_download))
```
svlandeg: `set` or `sorted` shouldn't be necessary if we're not in the `if` condition, because `ids_to_download = list(range(num))` is just fine. Within this `if`, we also don't need to do `set` because we already checked `not in`.
```suggestion
        ids_to_download.append(VAL_SHARD_INDEX)
        ids_to_download = sorted(ids_to_download)
```

KartikVashishta:
Thanks for the review @svlandeg !

nanochat/dataset.py
svlandeg: @KartikVashishta : can you address this comment?

KartikVashishta: Yep, agreed - removed the set/sorted. range(num) is already unique and ordered, and we only append val when it’s not present

svlandeg (commented):
This has now been fixed by a recent [edit](https://github.com/karpathy/nanochat/commit/324e69c45d3606095adb6b409078647145165454) on `master`, so I'll go ahead and close this one 🙏

PR #549, opened 2026-02-21
do not sample bos in engine due to issues with sft

Jdubedition:


Jdubedition:
> Hi, thanks for the PR!
> 
> Could you provide some more context? When/how have you run into this issue? I'm wondering whether something went wrong during training for this issue to pop up.
> 
> Are you still working on this? (as you've opened it as a draft)

Hi, thanks for responding.

I actually meant to open this on my fork, not on this repo.  I will move it over there.

When I train with the chat_sft.py, the response from chat_cli.py or chat_web.py sampling is just one |bos| and it stops.  If I target the base checkpoint, it works as expected.

I will submit this when I have tested this more.

Thanks.

PR #559, opened 2026-02-22
Add L3 (Large Lookup Layers) implementation

jhspaybar:
## Summary
- Implements L3 (Large Lookup Layers) from arXiv:2601.21461v2, which places per-token lookup tables inside the decoder stack with static routing (determined by token ID, not hidden state)
- Adds `nanochat/l3.py` with LZW-style frequency-based allocation and vectorized L3Layer module (gather+pad+mask approach, tied/untied KV support)
- Integrates L3 into GPT model: layers sit between decoder blocks applied residually, with proper init (w_mix=zeros for no-op at start), optimizer grouping (embeddings→AdamW, matrices→Muon), and FLOPs accounting
- Adds CLI flags (`--l3-after-layers`, `--l3-n-emb`, `--l3-d-up`, `--l3-k-max`) with automatic LZW precomputation from training data

## Test plan
- [x] 17 tests in `tests/test_l3.py` covering:
  - LZW allocation (total, min-one, max-k, distribution, bounds, edge cases)
  - L3Layer (output shape, gradient flow, tied/untied KV, masking, determinism)
  - GPT integration (forward, backward, optimizer coverage, disabled=unchanged)
- [x] All existing tests pass (no regressions)
- [ ] Smoke test: `python -m scripts.base_train --depth=4 --l3-after-layers=2 --l3-n-emb=8192 --num-iterations=5 --device-batch-size=1 --total-batch-size=2048 --max-seq-len=512`

### DIFF
--- a/nanochat/gpt.py
+++ b/nanochat/gpt.py
@@ -24,6 +24,7 @@ from nanochat.optim import MuonAdamW, DistMuonAdamW
 
 # Our custom Flash Attention module that automatically uses FA3 on Hopper+ and SDPA fallback elsewhere
 from nanochat.flash_attention import flash_attn
+from nanochat.l3 import L3Layer
 
 @dataclass
 class GPTConfig:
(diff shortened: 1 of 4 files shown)


jhspaybar:
Opened against wrong repo by mistake, moving to fork.

PR #573, opened 2026-02-28
fix(chat_web): validate ,--num-gpus against available GPUs for clear s…

dhunganapramod9:
Using --num-gpus greater than available GPUs led to a later CUDA error. Added an upfront check and raise a clear ValueError with the requested and available counts.


svlandeg (commented):
What does the "normal" CUDA error look like? I think it's also quite informative/clear, no?

svlandeg (commented):
Thanks for the PR @dhunganapramod9. This has now become obsolete after https://github.com/karpathy/nanochat/commit/da32e1d657b62cc1110f20f86fbc301d6189726e, so I'll go ahead and close this 🙏

PR #580, opened 2026-02-28
fix: correct CSV extraction in scaling_laws.sh

geopti:
## Bug fixes

Two bugs in `runs/scaling_laws.sh` cause the results CSV to be silently wrong.

### Bug 1 — parameter columns always empty

`base_train.py` prints parameter counts with 24-character key padding:
```
wte                     : 33,554,432
value_embeds            : 150,994,944
transformer_matrices    : 167,772,160
...
```

The grep patterns used `"wte:"`, `"value_embeds:"` etc., which never match because there are spaces between the key and the colon. Some parameter columns in the CSV are silently empty.

**Fix:** use `grep -P "wte\s+:"` to handle the padding.

### Bug 2 — tokens_trained always wrong

`tokens_trained` was computed as `NUM_ITERS * 524288`, hardcoding the default batch size. But `base_train.py` auto-computes the optimal batch size based on the FLOPs budget and model size — it may differ from 524288, especially for small models at small FLOPs budgets.

**Fix:** extract `tokens_trained` directly from the log line `"Total number of training tokens: X"`.


hitchho (approved):
Both fixes are correct — the grep pattern needed whitespace handling and tokens_trained should come from the log, not a hardcoded constant. Applied consistently to miniseries.sh too.

PR #588, opened 2026-03-05
Add upload to hf hub at the end of training

Randl:


nanochat/checkpoint_manager.py
```diff
@@ -59,9 +60,9 @@ def save_checkpoint(checkpoint_dir, step, model_data, optimizer_data, meta_data,
         logger.info(f"Saved optimizer state to: {optimizer_path}")
 
 def load_checkpoint(checkpoint_dir, step, device, load_optimizer=False, rank=0):
-    # Load the model state
-    model_path = os.path.join(checkpoint_dir, f"model_{step:06d}.pt")
-    model_data = torch.load(model_path, map_location=device)
+    # Load the model state in safetensors format
```
svlandeg, line 63: This would break bwd-compat, right? You wouldn't be able to load a checkpoint from before this PR anymore?

Randl: We can check for the old path too. I didn't consider people updating code mid training, but I don't have strong opinion about it

Randl:
Thinking of it, it might be convenient to also upload the report and the commit hash of the training code

svlandeg (commented):
To be honest, I'm not sure about the added value here. To upload your files to HF it's really just a few lines of code. Personally I wouldn't make this part of the `base_train` or `chat_sft` scripts, I would only upload specific runs after evaluating them when the run is finished. So I'm not sure it warrants all the additional feature flags etc.

svlandeg (commented):
Discussed with Andrej and decided to close 🙏

PR #603, opened 2026-03-07
feat: aarch64/Jetson support + SFT stability fixes

rohithreddy1095:
## Summary

Enables nanochat to run end-to-end on **aarch64 platforms** (NVIDIA Jetson, ARM servers) and fixes two SFT training stability bugs that affect all platforms.

Tested on **NVIDIA Jetson Orin NX 8GB** — full pipeline (pretrain → SFT → chat) runs in ~20 minutes.

### 1. Conditional `torch.compile` for non-Triton platforms
Triton is not available on aarch64. The current hard-coded `@torch.compile` and `torch.compile(model)` calls crash immediately on these platforms.

**Fix:** Made compilation conditional — falls back to eager mode when Triton is not installed. Zero behavior change on platforms where Triton is available.

Files: `nanochat/optim.py`, `scripts/base_train.py`, `scripts/chat_sft.py`, `nanochat/common.py` (added Orin to FLOPS table)

### 2. NaN guard for all-masked batches in SFT (`nanochat/gpt.py`)
The bestfit conversation packer can produce batches where **all targets are `ignore_index=-1`** (e.g., when most conversations are longer than `max_seq_len` and only padding remains). `F.cross_entropy(reduction='mean')` returns NaN (0/0 division) in this case, which propagates through the model and corrupts all weights permanently.

**Reproducing:** With the default `identity_conversations.jsonl` (avg 526 tokens) and `max_seq_len=256`, 90.9% of conversations can never fit in any packed sequence. After exhausting short conversations (~136 steps), the packer produces all-padding batches → NaN → unrecoverable.

**Fix:** Detect when no valid targets exist and return zero loss (with grad_fn preserved for autograd).

### 3. Gradient clipping in SFT (`scripts/chat_sft.py`)
Added `clip_grad_norm_(max_norm=1.0)` to prevent gradient explosions during SFT. The Muon optimizer with variable-length packed sequences can produce large gradient spikes.

## Testing
- **Platform:** NVIDIA Jetson Orin NX 8GB (Compute 8.7, CUDA 12.6, PyTorch 2.8.0, aarch64)
- **Pretrain:** 500 steps, 10 min, val BPB 3.10→1.88 (d4, 36.7M params)
- **SFT:** 3,146 steps, 10.4 min, val BPB 2.88→1.38, peak 873 MB
- No regressions expected on x86/Triton platforms (conditional paths only)

### DIFF
--- a/nanochat/gpt.py
+++ b/nanochat/gpt.py
@@ -425,7 +425,13 @@ class GPT(nn.Module):
         if targets is not None:
             # training: given the targets, compute and return the loss
             # TODO experiment with chunked cross-entropy?
-            loss = F.cross_entropy(logits.view(-1, logits.size(-1)), targets.view(-1), ignore_index=-1, reduction=loss_reduction)
+            # Guard against NaN from all-masked batches (0/0 in mean reduction)
+            # Use sum reduction on valid targets; if none exist, loss is 0
+            n_valid = (targets.view(-1) != -1).sum()
+            if n_valid == 0:
+                loss = (logits * 0).sum()  # zero loss with grad_fn
+            else:
+                loss = F.cross_entropy(logits.view(-1, logits.size(-1)), targets.view(-1), ignore_index=-1, reduction=loss_reduction)
             return loss
         else:
             # inference: just return the logits directly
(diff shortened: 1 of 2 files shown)


rohithreddy1095:
Thanks for the review @svlandeg! You're right on both points:

1. The all-masked batch issue is better addressed by #486 at the dataloader level
2. This PR was too broad — splitting into focused PRs

I've created **#604** with just the `torch.compile` conditional fix (3 files, clean and focused). Closing this one.

Thanks for the guidance!

PR #604, opened 2026-03-07
feat: conditional torch.compile for non-Triton platforms (aarch64/Jetson)

rohithreddy1095:
## Problem

`torch.compile` requires Triton, which is not available on aarch64 platforms (NVIDIA Jetson, ARM servers). Running nanochat on these platforms crashes immediately:

```
ModuleNotFoundError: No module named 'triton'
```

## Fix

Make `torch.compile` conditional — check for Triton availability and fall back to eager mode when it's missing. Three files changed:

- **`nanochat/optim.py`**: Replace hardcoded `@torch.compile(dynamic=False, fullgraph=True)` decorators with a conditional `_compile_decorator` that becomes a no-op when Triton is unavailable
- **`scripts/base_train.py`**: Wrap `torch.compile(model)` in a Triton import check
- **`nanochat/common.py`**: Add Orin to the peak FLOPS table for MFU reporting on Jetson

## No behavior change on existing platforms

When Triton is available (x86 with CUDA), the code paths are identical to before. The conditional only activates on platforms where `import triton` fails.

## Testing

Tested end-to-end on NVIDIA Jetson Orin NX 8GB (Compute 8.7, CUDA 12.6, PyTorch 2.8.0):
- Pretrain: 500 steps, val BPB 1.88 ✓
- SFT: 3,146 steps, val BPB 1.38 ✓
- Inference/chat: working ✓

---

*Split from #603 per review feedback to keep changes focused.*

### DIFF
--- a/nanochat/common.py
+++ b/nanochat/common.py
@@ -264,6 +264,8 @@ def get_peak_flops(device_name: str) -> float:
         (["5090"], 209.5e12),
         (["4090"], 165.2e12),
         (["3090"], 71e12),
+        # NVIDIA Jetson (unified memory, iGPU)
+        (["orin"], 5.3e12),  # Jetson Orin NX 8GB: ~5.3 TFLOPS BF16 (8 SMs @ 1.0GHz)
     )
     for patterns, flops in _PEAK_FLOPS_TABLE:
         if all(p in name for p in patterns):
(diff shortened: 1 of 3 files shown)


rohithreddy1095:
can this be merged ? Few of my further local implementaiton depends on this

PR #609, opened 2026-03-08
Add Flash Attention 4 (FA4) support for Blackwell GPUs

liuyixin-louis:
## Summary

- Adds FA4 ([flash-attn-4](https://github.com/Dao-AILab/flash-attention) CuTeDSL backend) as the top-priority attention backend
- Blackwell (B200/B100, sm100) and Hopper (H100/H200, sm90) GPUs now use FA4 instead of falling back to SDPA
- Detection priority: **FA4 > FA3 > SDPA** — fully backwards-compatible, zero config needed
- Wraps FA4 with `@torch.compiler.disable` to prevent torch.compile/dynamo from tracing into CuTeDSL internals

### Changes (3 files, +68/-16 lines)

| File | Change |
|------|--------|
| `nanochat/flash_attention.py` | FA4 loader, detection, dispatch, window_size translation |
| `scripts/base_train.py` | Updated FA status logging to show active backend |
| `scripts/chat_sft.py` | Updated FA status logging to show active backend |

### Tested on

2x NVIDIA B200 (Blackwell sm100), depth 20 with FP8: val BPB **0.833**, ~31% bf16 MFU over ~12K steps.

## Test plan

- [x] Verified FA4 auto-detected on Blackwell B200 GPUs
- [x] Verified FA3 fallback still works on Hopper when FA4 is not installed
- [x] Verified SDPA fallback on CPU/MPS (no regression)
- [x] Training loss and val BPB curves look healthy

### DIFF
--- a/nanochat/flash_attention.py
+++ b/nanochat/flash_attention.py
@@ -1,8 +1,10 @@
 """
-Unified Flash Attention interface with automatic FA3/SDPA switching.
+Unified Flash Attention interface with automatic FA4/FA3/SDPA switching.
 
-Exports `flash_attn` module that matches the FA3 API exactly, but falls back
-to PyTorch SDPA on non-Hopper GPUs (including Blackwell), MPS, and CPU.
+Exports `flash_attn` module that matches the FA3 API exactly, but uses:
+  - FA4 (flash-attn-4) on Blackwell (sm100) and Hopper (sm90) GPUs
+  - FA3 (kernels hub) on Hopper (sm90) if FA4 is not installed
+  - PyTorch SDPA fallback on all other hardware (MPS, CPU, older GPUs)
 
 Usage (drop-in replacement for FA3):
     from nanochat.flash_attention import flash_attn
@@ -18,8 +20,22 @@ import torch.nn.functional as F
 
 
 # =============================================================================
-# Detection: Try to load FA3 on Hopper+ GPUs
+# Detection: Try to load FA4 (Hopper + Blackwell), then FA3 (Hopper only)
 # =============================================================================
+def _load_flash_attention_4():
+    """Try to load Flash Attention 4 (supports Hopper sm90 and Blackwell sm100)."""
+    if not torch.cuda.is_available():
+        return None
+    try:
+        major, _ = torch.cuda.get_device_capability()
+        if major not in (9, 10):  # FA4 supports sm90 (Hopper) and sm100 (Blackwell)
+            return None
+        from flash_attn.cute import flash_attn_func as fa4_func
+        return fa4_func
+    except Exception:
+        return None
+
+
 def _load_flash_attention_3():
     """Try to load Flash Attention 3 (requires Hopper GPU, sm90)."""
     if not torch.cuda.is_available():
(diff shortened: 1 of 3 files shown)


svlandeg (commented):
Thanks!

This doesn't specify how `flash_attn.cute` would be installed - it would require local compilation, no?

Huggingface is currently working on getting a community kernel out for FA4 (`kernels-community/flash-attn4`) - we may also opt to wait for that instead of requiring users to compile this themselves.

Either way, if FA4 is being used for Hopper, we'd want to see some benchmarks results that it is indeed faster than the previous implementation using FA3 (I would assume yes, but always good to see the numbers).

liuyixin-louis:
Thanks for the review!

**Installation**: `flash-attn-4` is available as a pre-built pip package — no local compilation needed:

```bash
pip install flash-attn-4==4.0.0b4
```

This pulls in `cuda-python`, `nvidia-cutlass-dsl`, and `quack-kernels` automatically.

**kernels-community/flash-attn4**: Happy to switch to the HuggingFace community kernel once it's ready. The current implementation is structured to make that swap easy — just change the import in `_load_flash_attention_4()`.

**Benchmarks**: I tested on 2×B200 (Blackwell, sm100), depth 20, ~1K steps. On Blackwell, FA3 is not available (Hopper-only kernels), so the relevant comparison is FA4 vs SDPA fallback:

| Config | Val BPB | MFU | Throughput (tok/s) | Speedup |
|--------|---------|-----|--------------------|---------|
| SDPA (baseline) | 0.925 | 18.3% | 274K | 1.00× |
| FA4 (bf16) | 0.925 | 31.1% | 466K | **1.70×** |
| FA4 + FP8 | 0.933 | 31.7% | 474K | 1.73× |

FA4 matches SDPA quality exactly (BPB difference of 0.0003) while delivering a **1.70× throughput improvement**. The FP8 variant trades a small quality delta for slightly higher throughput, as expected.

I don't have access to H100 hardware to benchmark FA4 vs FA3 on Hopper directly, but the FA4 README reports comparable or better performance vs FA3 on sm90.

svlandeg:
> Installation: flash-attn-4 is available as a pre-built pip package — no local compilation needed:

But my point is - your PR won't work out of the box, right? You haven't edited the requirements of this project? A `speedrun` should just run from scratch with `uv sync` without requiring any implicit prerequisites.

Ofc I get that the main point of this PR is improving things for Blackwell but the code does change the setup for Hopper as well, so we'd need to get this tested first before merging, to make sure there are no regressions etc. I know it's a bit tough though, as H100's are hard to get by these days...

Randl:
Just to add a datapoint, I've run a d24 on 8xB200 machine, with cuda 13 and pytorch 2.11.

Baseline gave me 1,543,690 tok/s and 41% MFU, with a device batch size of 64 and fp8.
Baseline without fp8 (batch 32) gave ~1,405,000 tok/s.

FA4 with fp8 (this PR + new pytorch/cuda) OOMs at batch size 64 and at batch size 32 gives ~1,485,000 tok/s.
FA4 without fp8 ~1,360,000 tok/s.

So for me FA4 is ~4% slower both with and without fp8.

PR #613, opened 2026-03-09
pin validation shard by filename instead of sort order

### LINKED ISSUE #541: make validation loss reproducible

mvanhorn:
Fixes #541

The val split used `parquet_paths[-1:]` which relies on the last file in sorted directory listing. This is fragile - the val shard should always be `shard_06542` regardless of which other shards exist.

Changes:
- Add `VAL_SHARD` constant and `split_parquet_paths()` helper to `dataset.py`
- Both `dataset.py` and `dataloader.py` now use explicit filename matching instead of position-based slicing
- Clear assert message if the val shard is missing

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/nanochat/dataloader.py
+++ b/nanochat/dataloader.py
@@ -20,7 +20,7 @@ import torch
 import pyarrow.parquet as pq
 
 from nanochat.common import get_dist_info
-from nanochat.dataset import list_parquet_files
+from nanochat.dataset import list_parquet_files, split_parquet_paths
 
 def _document_batches(split, resume_state_dict, tokenizer_batch_size):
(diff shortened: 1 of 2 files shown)


mvanhorn:
You're right - the current dataset naming makes this scenario pretty much impossible, and I didn't consider that it might break legacy dataset usage. Closing this one.

PR #626, opened 2026-03-10
Fix wandb cosmetic issues

filipviz:
This PR fixes a few cosmetic wandb issues:
1. Update the pinned `wandb` library version. The old version raises when given new `wandb` API keys!
2. Move the `step` argument to the right place in `wandb.log` calls. The signature is `wandb.log(data: dict, step: int, commit: bool)` - previously, step counts were being included in the data dict, meaning wandb metrics incorrectly had x-axes corresponding to the number of calls to `.log` instead of the number of training steps.
3. Move `wandb.init` later in `chat_sft.py` and `base_train.py` to include config values that are calculated or read from a checkpoint.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ dependencies = [
     "torch==2.9.1",
     "transformers>=4.57.3",
     "uvicorn>=0.36.0",
-    "wandb>=0.21.3",
+    "wandb>=0.25.0",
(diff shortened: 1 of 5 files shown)


filipviz:
I also considered adding a CLI option to enable `wandb.watch` for periodic weight/gradient distribution statistic logging, but decided it was out of scope. Let me know if you'd like me to add that to the PR.

PR #627, opened 2026-03-10
Add bounds checking to KVCache.advance() method

hiSandog:
This PR adds bounds checking to the KVCache.advance() method to prevent cache overflow beyond max_seq_len.

Changes:
- Added validation in KVCache.advance() to check if new cache position would exceed max_seq_len
- Raises ValueError with clear error message if cache overflow is detected
- Helps catch potential bugs early during inference


svlandeg (commented):
Please clean your branch/diff, overlap with https://github.com/karpathy/nanochat/pull/602

PR #633, opened 2026-03-12
perf: BackLite sparse backward pass (1.76 hours, 2.10% speedup)

omershlo:
## Summary

- Integrates [BackLite](https://github.com/moonmath-ai/BackLite), a drop-in Hopper kernel that accelerates the backward pass of Flash Attention 3 by stochastically skipping attention tiles that contribute negligible probability mass 
- Achieves **1.76 hours** time-to-GPT-2 (down from 1.80 hours in Run 5), a **2.10% wall-clock speedup**
- Average CORE = **0.2586** (above GPT-2 threshold of 0.2565), average val\_bpb = **0.7200**

## Run details

**Launch command** (same as Run 5 + `--backlite-negl-prob=0.04` + `--negl-prob-warmup-steps=100`):

```
OMP_NUM_THREADS=1 torchrun --standalone --nproc_per_node=8 -m scripts.base_train -- \
    --depth=24 \
    --run="d24-mar10-prob0.04-ratio8.7" \
    --model-tag="d24_mar10_prob0.04_ratio8.7" \
    --sample-every=-1 \
    --save-every=-1 \
    --core-metric-every=999999 \
    --target-param-data-ratio=8.7 \
    --device-batch-size=16 \
    --fp8 \
    --backlite-negl-prob=0.04 \
    --negl-prob-warmup-steps=100
```

**Results (averages):**

| Metric | Run 5 (baseline) | Run 6 (BackLite) | Delta |
|---|---|---|---|
| CORE | 0.2690 | 0.2586 | -3.9% (still above GPT-2 0.2565) |
| val\_bpb | 0.71808 | 0.72003 | +0.27% (slight regression) |
| Training time | 1.80 hr | ~1.76 hr | **-2.10%** |

We ran this command 4 individual times on 4 different machines. The spread of CORE scores:

```
0.2588
0.2516
0.2593
0.2648
```

Mean is 0.25863 (above the GPT-2 threshold of 0.2565).

## How it works

BackLite modifies the Flash Attention 3 backward pass. During the forward pass, per-tile log-sum-exp (LSE) statistics are saved. In the backward pass, these statistics are used to generate a block sparsity mask that identifies tiles contributing negligible probability mass (below `negl_prob` threshold). Those tiles are skipped entirely in the gradient computation, reducing backward-pass FLOPS with minimal impact on gradient quality. For more technical details, see [here](https://moonmath.ai/posts/introducing-backlite/).

## Changes (3 files)

- **`nanochat/flash_attention.py`** — Adds BackLite as the highest-priority attention backend (BackLite > FA3 > SDPA). New `set_backlite_negl_prob()` API. BackLite only applies to training (forward+backward); inference always uses vanilla FA3.
- **`scripts/base_train.py`** — New CLI arg `--backlite-negl-prob`. Disables BackLite during eval/sampling. Pre-compiles both code paths to avoid `torch.compile` recompilation spikes.
- **`dev/BACKLITE.md`** — Documentation for BackLite install and usage: how to clone/build the Hopper kernel, and how to launch a BackLite training run with `--backlite-negl-prob`.


omershlo:
Thanks for taking the time to review and for discussing it with Andrej. 

A few clarifications on the intent of the PR for readers: 

- This implementation should be viewed as a proof-of-concept. Our code is open source and we welcome feedback and suggestions on how to improve or simplify it.
- The algorithm itself is general and can be applied to any LLM training setup; nanochat was simply a convenient environment to demonstrate the technique.
- We specifically chose nanochat because it is a small model, which allows us to test and demonstrate ideas quickly and cheaply. We appreciate the work of the authors and maintainers of this repository for providing such a clean and useful benchmark.

Regarding the performance:
- Because sequence lengths in nanochat are VERY small, there is limited sparsity to exploit, which naturally caps the achievable speedup. This is why the observed improvement is only ~2%.
- In larger models, we expect significantly larger gains where sparsity becomes more meaningful.

Regarding the metrics:
- We used basic calibration only, just enough to remain above the required threshold (there was no requirement to improve the core of entry 5).
- With full offline tuning, it should be possible to recover most or all of the quality delta while keeping the training speed improvement.

Finally, @svlandeg:  we understand that merging this PR may not be possible  given the concerns you raised. That said, we wonder if it might still be possible to include the result as an honorary mention on the leaderboard. To the best of our understanding, one of the goals of nanochat was to provide a clean playground where teams such as ours can experiment with new ideas and explore different optimization approaches. In that spirit, it would be great if the effort could still receive some acknowledgment and more importantly - make others aware of the new algorithm who can benefit many improving their training pipeline. As Andrej wrote when introducing the benchmark, the idea was to iterate on the “time to GPT-2” leaderboard and encourage contributions from others, with the hope that nanochat becomes a nice experimental harness for prototyping ideas, learning, and having fun. Our submission was intended in exactly that spirit. cc: @karpathy 

Thanks again for taking the time to review the work! 

P.S. we measured the performance of backlite compared to the new entry6 (99min) - while the sequence length got even smaller, BackLite still shaves around 1 minute of running time, while staying above threshold.

Issue #637, opened 2026-03-13
confusion on mmlu dataset


mathieu-lacage:
The current code in tasks/mmlu.py is confused about the mmlu dataset splits. It incorrectly assumes it has the 4 splits: test, validation, dev, and train. Instead, this dataset has the 4 following splits: test, validation, dev, and auxiliary_train.

Note above: the correct split name is "auxiliary_train", not "train" which is indeed confusing but it is what it is: 

```
>>> ds = datasets.load_dataset('cais/mmlu', 'all')
>>> ds
DatasetDict({
    test: Dataset({
        features: ['question', 'subject', 'choices', 'answer'],
        num_rows: 14042
    })
    validation: Dataset({
        features: ['question', 'subject', 'choices', 'answer'],
        num_rows: 1531
    })
    dev: Dataset({
        features: ['question', 'subject', 'choices', 'answer'],
        num_rows: 285
    })
    auxiliary_train: Dataset({
        features: ['question', 'subject', 'choices', 'answer'],
        num_rows: 99842
    })
})

```

The confusion comes from the fact that this dataset also has defined a subset = "auxiliary_train":
```
>>> ds = datasets.load_dataset('cais/mmlu', 'auxiliary_train')
Warning: You are sending unauthenticated requests to the HF Hub. Please set a HF_TOKEN to enable higher rate limits and faster downloads.
>>> ds
DatasetDict({
    train: Dataset({
        features: ['train'],
        num_rows: 99842
    })
```
})

which appears to be exactly equal to the split "auxiliary_train" of the "all" subset.

My suggestion is to change the current code as below which I find significantly less confusing: note the "I don't understand" comment is gone because it was due to the above-described confusion.

```
diff --git a/scripts/chat_sft.py b/scripts/chat_sft.py
index c1adbb6..ab886a7 100644
--- a/scripts/chat_sft.py
+++ b/scripts/chat_sft.py
@@ -166,7 +166,7 @@ train_tasks = [
     SmolTalk(split="train"), # 460K rows of general conversations
     CustomJSON(filepath=identity_conversations_filepath), # 1000 rows of synthetic identity conversations
     CustomJSON(filepath=identity_conversations_filepath), # 2 epochs of these
-    *[MMLU(subset="auxiliary_train", split="train") for _ in range(args.mmlu_epochs)], # 100K rows per epoch
+    *[MMLU(subset="all", split="auxiliary_train") for _ in range(args.mmlu_epochs)], # 100K rows per epoch
[...]

PR #639, opened 2026-03-13
MMLU main split is named auxiliary_train, not train

mathieu-lacage:
fixes issue #637


karpathy:
Thanks @mathieu-lacage! Great catch and clean fix. The "I don't understand why" comment was the giveaway that something was off.

Verified by loading both paths in Python — `subset="all", split="auxiliary_train"` returns the same 99,842 rows with clean columns (`question`, `subject`, `choices`, `answer`) directly, no unwrapping needed. All four splits under `subset="all"` load correctly.

PR #644, opened 2026-03-14
Mup implementation for nanochat

amritbulusu:
Addressing the muP discussion with @karpathy: https://github.com/karpathy/nanochat/discussions/481#discussioncomment-15666695

The changes on top of SP are minimal:

  Forward pass: logits *= base_width / n_embd. It keeps the loss shape width invariant by curbing logit magnitudes - O(√width).

  Init: lm_head init std scaled by √(base_width / n_embd) to match.

  LR scaling: None changes needed. muP prescribes 1/width LR scaling for hidden and output layers, but with Muon+AdamW both are redundant:
  - Muon: Polar Express orthogonalization produces ‖update‖_F ≈ 1 regardless of matrix dimensions, so updates are naturally width-independent.
  - lm_head (AdamW): The logit scaling factor backpropagates into the gradient, and Adam's second moment normalizes it, an additional 1/width LR factor is not needed.

  Verified with coord check (activation magnitudes flat across widths 128 - 2048 ) and transfer check (optimal LR multiplier = 1.0 across widths 128 - 2048 under muP, vs shifting from 0.5→0.125 under SP). Details and derivations in muP_changes.md. (Thanks to Claude Code)
  
  
  
image
image


svlandeg (commented):
Hi!

I tried running this with just `--use-mup`; trained a d14 for 2192 iterations (data ratio 7.0) on an A100. It runs fine, but compared to `master`, the result is slightly worse:

- same training time: 3u23 -> 3u24
- slightly worse `val_bpb`: 0.838 -> 0.841
- slightly worse CORE score: 0.16 -> 0.14

Did you obtain better results? Should I be setting `--base_width` to something non-default? Or is this really only meant as an extra knob to turn for `autoresearch`?

amritbulusu:
> Hi!
> 
> I tried running this with just `--use-mup`; trained a d14 for 2192 iterations (data ratio 7.0) on an A100. It runs fine, but compared to `master`, the result is slightly worse:
> 
> * same training time: 3u23 -> 3u24
> * slightly worse `val_bpb`: 0.838 -> 0.841
> * slightly worse CORE score: 0.16 -> 0.14
> 
> Did you obtain better results? Should I be setting `--base_width` to something non-default? Or is this really only meant as an extra knob to turn for `autoresearch`?

thanks for testing! a few things:

-   The hyperparameters that gpt.py currently has were tuned for standard parametrization (SP).  Maximal-update parametrization (muP) changes the effective scaling of learning rates, initialization.  So the existing HPs aren't expected to be optimal under muP, and the slight regression you're seeing is expected.

-   I haven't done HP tuning for the muP implementation yet, happy to do that. The --base-width default is 256, which means LRs should be tuned at width 256 first, then those values become the new defaults. 

-   Yes, there is value in using it as an autoresearch knob. The mu-transfer property means: we tune HPs on a small/cheap model (e.g., d6 or d8 at base_width=256), and those HPs should transfer as you scale up width. The transfer check script
  (scripts/mup_transfer_check.py) validates this, optimal LR multipliers stay flat across widths under muP while they spread under SP. I haven't explicitly tested depth transfer (scaling depth while keeping width fixed), but with Muon + muP we should
  expect reasonable HP transfer across model sizes. 
  
  Reference: 
  https://www.cerebras.ai/blog/the-practitioners-guide-to-the-maximal-update-parameterization
  https://arxiv.org/pdf/2203.03466

Muesli1:
Hey!

A muP implementation sounds like  a very nice addition to nanochat, especially since the changes are so minimal.
But as far as I understand, muP is only really designed/researched for AdamW, not Muon, correct?

Also, as the project currently stands, there are fixed hyperparameters for the betas of the AdamW hyperparameter groups, as well as momentum for the Muon optimizer. Can these stay fixed even when scaling up the model?
Additionally, batch size is an important factor that has a significant relationship with the learning rate. I'm not sure how compatible this is with current muP knowledge. 

But the question I'm most worried about is the transferability of the weight decay scheduling. The paper [Muon is Scalable for LLM training](https://arxiv.org/html/2502.16982v1) seems to suggest that weight decay is a critical hyperparameter for training with Muon at scale that is dependent on the number of tokens used for training (which obviously are directly derived from the model size with Chinchilla-like scaling). 

I am not up-to-date with all the newest information on muP, so if you have more information on this, I would be very happy to get my hands on it. 

This is not a direct critique of your implementation or your idea to integrate this into nanochat; rather, it's a general concern I have regarding muP when using it in a Muon setting. However, the same problems obviously still exist with standard parameterization, so muP should most likely be preferred in any case!

amritbulusu:
> Hey!
> 
> A muP implementation sounds like a very nice addition to nanochat, especially since the changes are so minimal. But as far as I understand, muP is only really designed/researched for AdamW, not Muon, correct?
> 
> Also, as the project currently stands, there are fixed hyperparameters for the betas of the AdamW hyperparameter groups, as well as momentum for the Muon optimizer. Can these stay fixed even when scaling up the model? Additionally, batch size is an important factor that has a significant relationship with the learning rate. I'm not sure how compatible this is with current muP knowledge.
> 
> But the question I'm most worried about is the transferability of the weight decay scheduling. The paper [Muon is Scalable for LLM training](https://arxiv.org/html/2502.16982v1) seems to suggest that weight decay is a critical hyperparameter for training with Muon at scale that is dependent on the number of tokens used for training (which obviously are directly derived from the model size with Chinchilla-like scaling).
> 
> I am not up-to-date with all the newest information on muP, so if you have more information on this, I would be very happy to get my hands on it.
> 
> This is not a direct critique of your implementation or your idea to integrate this into nanochat; rather, it's a general concern I have regarding muP when using it in a Muon setting. However, the same problems obviously still exist with standard parameterization, so muP should most likely be preferred in any case!



hey, thanks for the comment, you raised some important questions - some of which is addressed in the literature. 

- on muP + Muon compatibility: they're actually two sides of the same coin (check Bernstein's blog - https://jeremybernste.in/writing/deriving-muon). Muon's orthogonalization step forces updates to have unit spectral norm, which is exactly the condition μP's LR scaling exists to enforce (check spectral condition paper - https://arxiv.org/abs/2310.17813). So Muon satisfies it automatically for hidden layers, making most μP LR prescriptions redundant. Only the output logit multiplier survives as essential - this is my takeaway from this exploration so far. 
- On the fixed betas/momentum:  the μP paper  (https://arxiv.org/abs/2203.03466) categorizes optimization HPs like Adam betas as transferable since they shape dynamics, not regularization. They hold steady across width empirically, so we can keep them unchanged. 
- batch size and training tokens are explicitly scaling axes not transferable HPs, so you tune at one batch size and transfer across width at that batch size. 
- weight decay is in the "does not transfer" bucket since it's primarily regularization and μP doesn't know data size. The Moonshot finding that WD is critical and token-dependent is fully consistent with this. for batch size + token horizon + depth transfer specifically see (https://arxiv.org/abs/2505.01618) and (https://arxiv.org/abs/2512.22382) which extend HP transfer across all these axes (I'm yet to explore this fully) 
- your last point is exactly right though. all these problems exist with SP too, except SP doesn't even give you width transfer. μP is strictly better as a starting point even with these open edges.

svlandeg (commented):
So to circle back, not sure we'll want to merge this if we can't obtain a solid improvement in training time, because the PR does also bring quite a lot of added complexity.

For reference, the graph for the run I talked about earlier:

Image

amritbulusu:
@svlandeg - I tried to reproduce your runs. After some hyperparameter tuning, I found that lr=0.005 and wd=0.2 works well for the mup parametrization. However, SP still works a little better.

Here's what I got:

mup run (lr=0.005, wd=0.2) - : https://gist.github.com/amritbulusu/460d33a479d132abbbd287d379a47c31
sp run: https://gist.github.com/amritbulusu/69bafd44011f743b76cc771f39cafbe0

  | | muP (lr=0.05, wd=0.2) | SP |
  |---|---|---|
  | val bpb | 0.8399 | 0.8382 |
  | CORE metric | 0.1481 | 0.1602 |

HP Sweep Details:

  | LR \ WD | 0.0    | 0.1    | 0.2    | 0.3    | 0.4    | 0.6    |
  |---------|--------|--------|--------|--------|--------|--------|
  | 0.003   | 4.6515 | 4.6400 | 4.6272 | 4.6156 | 4.6031 | 4.6416 |
  | 0.005   | 4.2361 | 4.2230 | 4.2069 | 4.1954 | 4.1802 | 4.1859 |
  | 0.007   | 4.0033 | 3.9899 | 3.9732 | 3.9596 | 3.9447 | 3.9265 |
  | 0.010   | 3.8014 | 3.7881 | 3.7707 | 3.7577 | 3.7478 | 3.7254 |
  | 0.015   | 3.6400 | 3.6265 | 3.6133 | 3.6026 | 3.5939 | 3.5695 |
  | 0.020   | 3.5607 | 3.5470 | 3.5334 | 3.5255 | 3.5147 | 3.4983 |
  | 0.030   | 3.4806 | 3.4657 | 3.4592 | 3.4535 | 3.4473 | 3.4402 |
  | 0.050   | 3.4278 | 3.4169 | 3.4129 | 3.4140 | 3.4186 | 3.4132 |
  
Sweep Configuration:

- width=256, depth=24, head_dim=128, n_head=2, vocab=32768, seq=2048, window=SSSL, muP=True
- batch=16, grad_accum=16, total_batch=524288 tok/step, steps=300, seed=42
- LRs (base): matrix=0.02, emb=0.3, unemb=0.008, scalar=0.5
- WD (base): 0.28
- Schedule: warmup=40, warmdown_ratio=0.65, final_lr_frac=0.05
- Muon momentum: 0.85→0.97 (400 steps), warmdown→0.90

Takeaways: 

- I may need to tweak the HPs a little more to find the optimal ones in the new parametrization. Obviously, I have only swept LR and WD.
- I set muon_lr_exponent as 0 which was only tested at small scale, the optimal exponent could be different. 
- My suspicion is that the sweep optimal HPs may not transfer to a full training duration, also width=256 as a sweep proxy is perhaps too small for reliable transfer to d14.
- The muP parameterization itself appears correct, coord check and mu-transfer checks both look good, so the regression is an HP tuning issue rather than a parameterization issue.
- Given that SP is currently outperforming and muP brings adds complexity, I think it makes sense to hold off on merging for now. A more thorough sweep is necessary to bridge the gap.

PR #646, opened 2026-03-16
Attention Residuals addition code

ar0cket1:
## Summary
- add an `attnres_block` residual mode that keeps the current nanochat residual path intact and applies AttnRes as a zero-init gated correction
- patch checkpoint loading for the new AttnRes parameters and expose the relevant training flags in `scripts/base_train.py`
- add local d12 result figures for the initial 0-500 run and the resumed 500-1000 continuation

## Notes
- baseline mode is unchanged current nanochat
- AttnRes starts exactly at the baseline behavior because the new per-layer gate is initialized to 0
- the 500-1000 figure is a checkpoint continuation from step 500, so its scheduler is not identical to a fresh 0-1000 run planned from step 0

## Local result note
See `dev/attention_residuals_local_results.md` for the two figures and checkpoint values.


Jake-Song:
I ran training a d16 model up to 1000 steps comparing baseline.
validation bpb is slightly worse than baseline. 

- GPU: NVIDIA H100 80GB HBM3 x 1
- Command: python-m scripts.base_train --depth=16 --residual-mode=attnres_block --fp8 --run=attnres_exp-attnres-depth16 --num-iterations=1000

## Observation
- Minimum validation bpb is slightly worse than baseline.
- Peak memory usage much higher than baseline.
- MFU(Model FLOPS Utilization) is so bad in attnres. training time is about 1.5x longer. 

Minimum validation bpb
- baseline: 0.880029
- attnres:  0.897636

Peak memory usage
- baseline: 49087.68 MiB
- attnres: 77796.47 MiB

MFU(Model FLOPS Utilization) 
- baseline: about 51%
- attnres: about 33%

Training time
- baseline: 27.05 minutes
- attnres: 42.00 minutes

attnres-d16

svlandeg:
I'm getting similar results, not being able to replicate a performance gain with this branch. In fact, on an A100, it OOMs for a d18, which doesn't happen on `master`

500 iterations for a d14 on an A100, comparing `master` with `baseline` and `attnres_block`:

image

image

image

image

svlandeg (commented):
As we don't see an improvement in `val_bpb` vs training time, I'll go ahead and close this PR. Thanks for the effor @ar0cket1 and thanks for helping to test @Jake-Song !

PR #648, opened 2026-03-17
Add optional Attention Residuals (Block AttnRes)

AaronXPeng:
## Summary

Adds optional **Block Attention Residuals** (AttnRes) from [MoonshotAI](https://github.com/MoonshotAI/Attention-Residuals) as an alternative to `resid_lambdas`/`x0_lambdas`.

- `--attn-res` flag enables learned depth-attention over block-level representations
- Two `nn.Parameter(n_layer, D)` pseudo-query vectors stored on GPT (one before attn, one before MLP)
- Block boundary logic in `GPT.forward`, keeping `Block` clean and unchanged
- Uses nanochat's existing parameterless `norm()` for key normalization (no learnable RMSNorm)
- ~0.06% parameter overhead, ~7% throughput cost, negligible memory impact

**GPU training results** (depth=4, 20 steps, RTX 6000 Ada):
| Metric | Standard | AttnRes |
|--------|----------|---------|
| Val bpb | 2.80 | **2.61** |
| Loss (final) | 10.13 | **9.36** |
| Throughput | ~840K tok/sec | ~780K tok/sec |

### Files changed
- `nanochat/gpt.py`: +68/-11 — `block_attn_res()` function, AttnRes path in `GPT.forward`
- `nanochat/checkpoint_manager.py`: +5 — backward compatibility patching
- `scripts/base_train.py`: +3 — `--attn-res` and `--attn-res-block-size` CLI args
- `tests/test_attn_res.py`: +267 — 18 comprehensive tests

## Test plan
- [x] All 31 tests pass locally (macOS, float32)
- [x] All 31 tests pass on GPU server (Linux, bf16 CUDA)
- [x] GPU training: standard path unchanged
- [x] GPU training: AttnRes converges faster
- [x] torch.compile compatible (Inductor)
- [x] KV cache inference works with AttnRes

### DIFF
--- a/nanochat/checkpoint_manager.py
+++ b/nanochat/checkpoint_manager.py
@@ -26,6 +26,11 @@ def _patch_missing_config_keys(model_config_kwargs):
     if "window_pattern" not in model_config_kwargs:
         model_config_kwargs["window_pattern"] = "L"
         log0(f"Patching missing window_pattern in model config to 'L'")
+    # AttnRes defaults to disabled
+    if "attn_res" not in model_config_kwargs:
+        model_config_kwargs["attn_res"] = False
+    if "attn_res_block_size" not in model_config_kwargs:
+        model_config_kwargs["attn_res_block_size"] = 4
 
 def _patch_missing_keys(model_data, model_config):
     """Add default values for new parameters that may be missing in old checkpoints."""
--- a/nanochat/gpt.py
+++ b/nanochat/gpt.py
@@ -10,6 +10,7 @@ Notable features:
 - no bias in linear layers
 - Group-Query Attention (GQA) support for more efficient inference
 - Flash Attention 3 integration
+- Optional Block Attention Residuals (AttnRes): https://github.com/MoonshotAI/Attention-Residuals
 """
 
 from functools import partial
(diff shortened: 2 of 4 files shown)


AaronXPeng:
> When I try to run this on an A100, I'm getting Torch warnings:
> 
> > /home/ubuntu/git/nanochat/.venv/lib/python3.10/site-packages/torch/_inductor/lowering.py:7242: UserWarning:
> 
> > Online softmax is disabled on the fly since Inductor decides to split the reduction. Cut an issue to PyTorch if this is an important use case and you want to speed it up with online softmax.
> 
> (but otherwise it's running)

Hi @svlandeg , thanks for testing this on A100! That warning is a known PyTorch Inductor behavior. This warning only fires from the new AttnRes code path (`logits.softmax(0) `in `block_attn_res()`), not from standard nanochat. The base model uses Flash Attention for attention softmax and fused `F.cross_entropy` for logits, both of which bypass Inductor's softmax lowering entirely.

The fallback itself is functionally correct and the perf impact here should be negligible (the softmax is over `n_layer`, a small dimension). But it may be worth noting in the PR description that this warning is expected with `--attn-res` on certain GPU configs until PyTorch 2.11+.

AaronXPeng:
> Ok so I've done a d14 run on an A100, and unfortunately haven't been able to replicate these results yet, compared to `master` as baseline:
> 
> * tok/sec from 140 to 80
> * wall clock time from 3u23 to 5u25
> * val bpb from 0.838 to 0.888
> * CORE from 0.16 to 0.13
> 
> This is just running with `--attn-res`, leaving all else on default, which is what you did as well, right?
> 
> Image

Hi @svlandeg , you were right to question this! Was excited to try out the AttnRes and jumped the gun on this one!  I went back and reran things properly with 200 steps instead of 20, and also tried a little tuning (smaller block size, higher LR). Here's what I found:
image
Turns out AttnRes does converge faster early on, which is exactly what I saw and got excited about. But I should have kept training longer before opening this. Baseline catches up around step 74 and wins from there. With tuning I got the gap down to almost nothing on quality (+0.002 BPB), but throughput still drops about 9%, so there's no real win.

Re-reading the original paper, I think the issue is fundamental. AttnRes was designed for Kimi Linear 48B where PreNorm dilution at extreme depth is a real problem. The paper recommends ~8 blocks. At d4 we only get 2, and even at d20 nanochat's existing `resid_lambdas` and `x0_lambdas` already handle depth scaling cheaply. There just isn't enough depth for the mechanism to pay for itself here.

Sorry for the noise! I should have done a full run before PRing! Suggesting we close this one. Thanks a lot!

svlandeg:
No worries! Thanks for reporting back, we're all learning 🙂

PR #540, opened 2026-02-18
Use HF's community kernel for FA3 for Ada and Ampere

svlandeg:
We currently use `varunneal/flash-attention-3`, but this is only supported on Hopper. The officially supported kernel `kernels-community/flash-attn3` however also works on Ada (RTX 40xx) and on Ampere (A100 and RTX 30xx).

There was a small bug before that is fixed now, cf https://github.com/huggingface/kernels-community/issues/312.

For Hopper, we stay with `varunneal/flash-attention-3` as it's quicker, cf below in thread comments 👇 

## Experiments on 8x A100

I just ran a `speedrun` on an 8x A100 which gave me 

```
GPU: NVIDIA A100-SXM4-80GB | Peak FLOPS (BF16): 3.12e+14
✓ Using Flash Attention 3: efficient, new and awesome.
```

and runs at about 315 `tok/sec`.

On the same branch and the same 8x A100, but using the SDPA fallback and setting `window-pattern="L"`:

```
GPU: NVIDIA A100-SXM4-80GB | Peak FLOPS (BF16): 3.12e+14
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
WARNING: Flash Attention 3 not available, using PyTorch SDPA fallback
WARNING: Training will be less efficient without FA3
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
gives about 309 `tok/sec`  (using the default `window-pattern` SSSL would be about 226 `tok/sec`.)

So not a huge improvement, but by using `kernels-community/flash-attn3` we can run `speedrun` with the default settings on the A100 & RTX 40/30xx architectures and it will work nicer out of the box.

~~I haven't double checked on an H100 yet, but I would expect the same performance as `varunneal/flash-attention-3`.~~

PR #663, opened 2026-03-22
Refactor for FA varlen

chrisjmccormick:
# Flash Attention Varlen

## Speedrun Result

**1.62 hours** (97.38 min), down from 1.65 hours (99 min), ~1.8% faster. 
* Val bpb `0.7248` 
* CORE `0.2614` 
(See also `dev/LEADERBOARD.md`). 

Model checkpoint is on huggingface [here](https://huggingface.co/ChrisMcCormick/nanochat-varlen-d24-2026-03-22), including the full [training log](https://huggingface.co/ChrisMcCormick/nanochat-varlen-d24-2026-03-22/blob/main/logs/d24_speedrun_20260322_150106.log).

---

## Why varlen

An [earlier investigation of varlen](https://github.com/karpathy/nanochat/blob/master/dev/LOG.md#2026-01-13-varlen-attention-negative-result) showed no quality improvement--isolating documents in attention didn't meaningfully reduce "noise" from cross-document bleeding. But isolating documents also means that flops are not wasted on cross-document attention. With many short documents packed into a long sequence, the varlen approach is faster. 

To maximize the benefit, I sized `max_num_docs` by analyzing the HillClimbMix dataset to ensure we aren't over-padding the `cu_seqlens` buffer. Leaving empty document positions in that tensor has a surprising impact on performance. (I also submitted a [PR](https://github.com/KellerJordan/modded-nanogpt/pull/246) to modded-nanogpt simply by tuning this size).

## Dataloader simplification

The current `bestfit` dataloader sorts and bin-packs documents into multiple rows, discarding a substantial number of tokens. 

The varlen loader is simpler: documents are greedily packed into a flat buffer with `cu_seqlens` tracking boundaries. Only the final document in the 1D batch is ever truncated, so far fewer tokens are wasted. 

The `cu_seqlens` tensor has a fixed size (for `torch.compile` performance) and is padded with "ghost segments" if needed.  `max_num_docs` is provided by the caller, and can be calculated by dividing the batch size by the average document size in the dataset and rounding up for padding. The calculation in `base_train.py` is tailored for HillClimbMix. 

## RoPE Embeddings

For varlen, we require enough embeddings to span the full 1D batch. For the d24 speedrun configuration this is 32K tokens. At d12 you can easily fit 64K. Claude sized this to 128K for headroom (64 * max_seq_len).

Peak memory usage went up `120 MiB` from `60285 MiB` to `60405 MiB`, perhaps due to this change.

## SFT pre-tokenization

`chat_sft.py` previously interleaved tokenization with training so the training loop couldn't know the total number of iterations upfront, leading to `approx_progress` globals and a guessed LR schedule. Pre-tokenizing all conversations at startup fixes this: it gives exact step counts, a clean LR schedule, and (the main reason I included the change) allows us to calculate a good value for `max_num_docs` instead of hardcoding.

## Evaluation updates

All evaluation paths (CORE, BPB, chat eval) were updated to use the varlen interface, fully removing the old batched code path. 

The one compromise is a `ModelWrapper._forward_varlen()` shim in `base_eval.py` for HuggingFace models. HF models don't support varlen natively, so the wrapper unpacks 1D varlen to padded `(num_docs, max_doc_len)` for the HF forward pass, then repacks. 

## FA2 support and fallback

This PR adds FA2 via `kernels-community/flash-attn2`, extending flash attention support for users on Ampere (A100) and Ada (RTX 4090) GPUs. 

For hardware without either FA2 or FA3, the SDPA fallback reshapes packed `(T, H, D)` back to `(B, T_seq, H, D)` and uses standard `is_causal=True` attention--no document isolation, but functional on all hardware including Mac/CPU/Blackwell. This matches the behavior of the previous codebase.

## AI Code

All of the changes were implemented by Claude Opus 4.6 and partially reviewed by me. Happy to review it more thoroughly if the overall change is accepted.

### DIFF
--- a/dev/LEADERBOARD.md
+++ b/dev/LEADERBOARD.md
@@ -200,3 +200,36 @@ This commit is special because all of the improvements that went into [this comm
 ## Run 6
 
 Achieved Mar 14, 2026 on commit `a825e63`. Exactly the same launch command as Run 4 except `--target-param-data-ratio=8`. Improvements in the architecture are allowing us to train shorter and shorter time. Instead of an undertrained d24 I attempted to train an overtrained d22 but it was worse. This set of changes came from autoresearch round 2, where I asked it to reference the modded-nanogpt repo for inspiration. So the exploration tried out a number of ideas and in particular found a way to incorporate the backout and smear in such a way that they are helpful (I had previously tried them manually a long time ago and they caused regressions). The smear idea in particular is a little bit heavier and bloaty because it is essentially an "early fusion" of context across tokens, producing a kind of a bigram input into the network and allowing it to focus on higher ngrams earlier. But for this reason the code gets a bit more complex and required some changes to inference. I verified with a unit test that the Engine inference is correct compared to the naive inference of `GPT.generate()`. The average of 5 runs was CORE 0.262634 and each of them lasted 1.65 hours (99 minutes).
+
+## Run 7
+
+Achieved Mar 22, 2026 on commit `a075166`. Launch command (same as `runs/speedrun.sh`):
+
+```
+OMP_NUM_THREADS=1 torchrun --standalone --nproc_per_node=8 -m scripts.base_train -- \
+    --depth=24 \
+    --run="d24-varlen" \
+    --model-tag="d24-varlen" \
+    --device-batch-size=16 \
+    --sample-every=-1 \
+    --save-every=-1 \
+    --core-metric-max-per-task=-1 \
+    --core-metric-every=999999 \
+    --target-param-data-ratio=8 \
+    --fp8
+```
+
+Result:
+
+```
+step 05567/05568 (99.98%) | loss: 2.388741 | lrm: 0.05 | dt: 1048.30ms | tok/sec: 1,000,264 | bf16_mfu: 60.37 | epoch: 1 pq: 117 rg: 64 | total time: 97.38m | eta: 0.0m
+Step 05568 | Validation bpb: 0.724772
+Step 05568 | CORE metric: 0.2614
+Peak memory usage: 52865.94MiB
+Total training time: 97.38m
+Minimum validation bpb: 0.724772
+```
+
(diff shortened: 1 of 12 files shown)


svlandeg (commented):
Hi, thanks for the submission! I'll try and replicate this on an A100 as a quick sanity check.

Just to note that `val_bpb` is quite a bit worse than current run 6 (0.71800), which may play into the decision on whether or not to accept this for the leaderboard. But that's ultimately for Andrej to decide.

chrisjmccormick:
@svlandeg Thanks, Sofie!

It occurred to me that the bpb here isn't directly comparable to the baseline--the changes to the dataloader mean that, even when using the same data shard for validation, the tokens will be different because of how many are discarded by the current best-fit loader.

My model weights should work without any of the varlen changes, so we could try loading the weights into the current codebase to get the correct bpb measurement.

karpathy:
Wow, very large change! I'd have to look pretty closely I'm not 100% sure if +2% is worth this

svlandeg (commented):
FYI - my results running this branch on an A100 (d14):

The training time & tok/sec are better than `master`, but in fact almost equal to PR #540 which simply adds `"kernels-community/flash-attn3"` as fallback for Ada & Ampere. So the improvement on the A100 is entirely due to using an FA kernel, in which case I'd prefer to merge #540 which is more minimal.

Haven't tested this on a H100 yet.

svlandeg:
Also, as a quick note:

> All of the changes were implemented by Claude Opus 4.6 and partially reviewed by me. Happy to review it more thoroughly if the overall change is accepted.

I do think the author of the PR should put in the effort of at least reviewing in detail the code they're submitting. Waiting for "acceptance" first is a bit like the chicken-or-the-egg problem. If you review the PR yourself first, and improve it and try to make sure the changes are minimal & easy reviewable, then your PR will also get a better change of getting accepted 😉

PR #667, opened 2026-03-25
update optim.py with compute_dtype

Krank41:
Change the data type conversion of gradient tensor to use COMPUTE_DTYPE instead of harcoded bfloat16.

### DIFF
--- a/nanochat/optim.py
+++ b/nanochat/optim.py
@@ -10,6 +10,7 @@ Further contributions from @karpathy and @chrisjmccormick.
 import torch
 import torch.distributed as dist
 from torch import Tensor
+from nanochat.common import COMPUTE_DTYPE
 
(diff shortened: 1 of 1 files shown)


karpathy:
Thanks for the suggestion! We agree the hardcoded bf16 should be COMPUTE_DTYPE-aware. However, using `COMPUTE_DTYPE` directly is unsafe when it's fp16 — the Polar Express (Newton-Schulz iteration) relies on bf16's wide exponent range for numerical stability, and fp16's limited range could cause overflow.

We landed a slightly different version in c0dbf1f: cast to bf16 when `COMPUTE_DTYPE == torch.bfloat16`, otherwise skip the cast entirely and let it run in the gradient's native dtype (fp32).

PR #669, opened 2026-03-26
96.2 min, 0.7177 valbp, 0.2613 CORE - Add sparse funnel and frontend feature controls

jn2clark:
# **Speedrun Result**

  Best 2 runs, metrics from the eval command:

  - 96.21 min, 0.717725, 0.2617
  - 96.20 min, 0.717730, 0.2609

#### At D24
  - Sparse funnel attention
  - G=6 global layer layout, using the chirped sparse-funnel rule with
    gamma=0.7; at depth=24 this resolves to 5,10,13,17,20,23
  - VE on the resolved global layers
  - Local attention window 128
  - local_mlp_ratio=4
  - rope_base=200000
  - Smear front-end
  - Hashed bigram path with bigram_vocab_size=4096
  - bigram_dim=128
  - Gated attention
  - device_batch_size=16
  - FP8 training with the default recipe
  - Shortened training horizon: 5368 iterations, which is 200 fewer than the
    previous result

  The evaluation was run following the same base evaluation used in the
  speedrun flow:

```
  torchrun --standalone --nproc_per_node=8 -m scripts.base_eval -- --device-
  batch-size=16
```

  ### Autoresearch

  These changes were found using an internal autoresearch system, with large parts of the code written by an LLM. There was a lot of exploration at `d=8`before moving to `d=12` and `d=18`, pruning candidates along the way. Both architecture and optimization ideas were explored, although no optimization changes made it into the final recipe. I do not have the exact number, but it was probably around 100-200 experiments.

  The search used many inputs, including this repo. I think there is still quite a bit to improve and simplify. The main objective was to find features that gave a disproportionate improvement in model quality per token, relative to any reduction in throughput, while also looking for changes that were neutral on quality but improved throughput. FA3 and FP8 changed which candidates were attractive quite a bit. Hardware-specific capabilities and kernels clearly influence which architectural directions are worth taking.

  ### Methodology

  The final architecture was selected using the methodology below. I did not ablate every permutation, but I did try to isolate a number of the main effects. Many of the final features were only slightly positive in isolation, but were still worthwhile when considering the tradeoff between model quality per token and system throughput.

  The candidate changes were tested at `d=8, d=12, d=18`, and finally `d=24`. The key criterion was that the relative improvement in `valbp/tok` had to more than compensate for any reduction in token density or iteration throughput, while also looking for ways to improve throughput without hurting performance.

  Performance over time can be decomposed as:

`  (valbp/tok) * (tok/sec) -> valbp/sec`

  and for a fixed time budget:

`  (valbp/tok) * (tok/sec) * (sec) -> valbp`

  Since the run is iteration-defined, throughput can be decomposed further
  as:

`  tok/sec = (tok/iter) * (iter/sec)`

  so:

`  (valbp/tok) * (tok/iter) * (iter/sec) -> valbp/sec`

  and:

`  (valbp/tok) * (tok/iter) * (iter/sec) * (sec) -> valbp`

  The interesting tradeoff is between:

  - `valbp/tok`: model efficiency / capacity
  - `tok/iter`: token density per step, including batch size, sequence length,
    and packing
  - `iter/sec`: system throughput

 Early candidate generation focused on improvements at a fixed iteration count (i.e improving `valbp/tok` ), which does not require Hopper, before shifting toward the fixed wall-clock budget (improving `valbp/sec`).

  Training longer appeared to be worse value in this regime, so selection was biased toward changes that improved `valbp/tok` materially while only minimally hurting iter/sec so total iterations could be reduced to reach a fixed valbp.

Note: I'll add in the scaling laws for any new parameters and then remove the extra cli arguments.

### DIFF
--- a/nanochat/engine.py
+++ b/nanochat/engine.py
@@ -102,11 +102,14 @@ class KVCache:
         self.cache_seqlens = torch.zeros(batch_size, dtype=torch.int32, device=device)
         # Previous token's normalized embedding for smear (set by model forward pass)
         self.prev_embedding = None
+        # Previous token ids for bigram hash features during decoding
+        self.prev_token_ids = None
 
     def reset(self):
         """Reset cache to empty state."""
         self.cache_seqlens.zero_()
         self.prev_embedding = None
+        self.prev_token_ids = None
 
     def get_pos(self):
         """Get current position (assumes all batch elements at same position)."""
@@ -135,6 +138,8 @@ class KVCache:
         # Copy smear state: expand batch=1 prev_embedding to num_samples
         if other.prev_embedding is not None:
             self.prev_embedding = other.prev_embedding.expand(self.batch_size, -1, -1).clone()
+        if other.prev_token_ids is not None:
+            self.prev_token_ids = other.prev_token_ids.expand(self.batch_size).clone()
 
 # -----------------------------------------------------------------------------
 @torch.inference_mode()
--- a/runs/speedrun.sh
+++ b/runs/speedrun.sh
@@ -69,8 +69,8 @@ python -m scripts.tok_eval
 echo "Waiting for dataset download to complete..."
 wait $DATASET_DOWNLOAD_PID
 
-# d24 model (slightly undertrained to beat GPT-2 => decrease data:params ratio from compute optimal 10.5 (default) to 8)
-torchrun --standalone --nproc_per_node=8 -m scripts.base_train -- --depth=24 --target-param-data-ratio=8 --device-batch-size=16 --fp8 --run=$WANDB_RUN
+# d24 sparse recipe (baked into the branch defaults), using the recorded 5,318-step horizon
+torchrun --standalone --nproc_per_node=8 -m scripts.base_train -- --depth=24 --target-param-data-ratio=8 --num-iterations=5318 --device-batch-size=16 --fp8 --run=$WANDB_RUN
 # evaluate the model: CORE metric, BPB on train/val, and draw samples
 torchrun --standalone --nproc_per_node=8 -m scripts.base_eval -- --device-batch-size=16
 
--- a/scripts/base_train.py
+++ b/scripts/base_train.py
@@ -25,7 +25,7 @@ import wandb
 import torch
 import torch.distributed as dist
 
-from nanochat.gpt import GPT, GPTConfig, Linear
+from nanochat.gpt import GPT, GPTConfig, Linear, default_sparse_n_global, compute_sparse_global_layers
 from nanochat.dataloader import tokenizing_distributed_data_loader_bos_bestfit, tokenizing_distributed_data_loader_with_state_bos_bestfit
 from nanochat.common import compute_init, compute_cleanup, print0, DummyWandb, print_banner, get_base_dir, autodetect_device_type, get_peak_flops, COMPUTE_DTYPE, COMPUTE_DTYPE_REASON, is_ddp_initialized
 from nanochat.tokenizer import get_tokenizer, get_token_bytes
@@ -52,6 +52,20 @@ parser.add_argument("--aspect-ratio", type=int, default=64, help="model_dim = de
 parser.add_argument("--head-dim", type=int, default=128, help="target head dimension for attention")
 parser.add_argument("--max-seq-len", type=int, default=2048, help="max context length")
 parser.add_argument("--window-pattern", type=str, default="SSSL", help="sliding window pattern tiled across layers: L=full, S=half context (e.g. 'SSL')")
+# Sparse funnel architecture
+parser.add_argument("--sparse-funnel", action=argparse.BooleanOptionalAction, default=True, help="use sparse funnel architecture instead of window pattern")
+parser.add_argument("--n-global", type=int, default=0, help="number of global (full-context) layers, 0 = auto depth//4")
+parser.add_argument("--chirp-gamma", type=float, default=0.7, help="chirp exponent for global layer placement")
+parser.add_argument("--local-window", type=int, default=128, help="local attention window size")
+parser.add_argument("--global-layers", type=str, default="", help="explicit global layer indices, comma-separated (overrides chirp)")
+parser.add_argument("--rope-base", type=int, default=200000, help="RoPE base frequency")
+parser.add_argument("--smear", action=argparse.BooleanOptionalAction, default=True, help="restore cheap bigram-like embedding mixing before the transformer trunk")
+parser.add_argument("--smear-channels", type=int, default=24, help="number of embedding channels used to predict smear gate")
+parser.add_argument("--bigram-vocab-size", type=int, default=4096, help="hashed bigram embedding buckets (0 disables)")
+parser.add_argument("--bigram-dim", type=int, default=128, help="bigram embedding dim before projection")
+parser.add_argument("--gated-attn", action=argparse.BooleanOptionalAction, default=True, help="learn lightweight per-head attention output gates")
+parser.add_argument("--attn-gate-channels", type=int, default=12, help="embedding channels used to predict attention gates")
+parser.add_argument("--ve-layers", type=str, default="", help="explicit value-residual layer indices, comma-separated")
 # Training horizon (only one used, in order of precedence)
 parser.add_argument("--num-iterations", type=int, default=-1, help="explicit number of optimization steps (-1 = disable)")
 parser.add_argument("--target-flops", type=float, default=-1.0, help="calculate num_iterations to reach target_flops (-1 = disable)")
@@ -78,6 +92,14 @@ parser.add_argument("--save-every", type=int, default=-1, help="save checkpoints
 # Output
 parser.add_argument("--model-tag", type=str, default=None, help="override model tag for checkpoint directory name")
 args = parser.parse_args()
+
+if args.sparse_funnel and args.n_global <= 0:
+    args.n_global = default_sparse_n_global(args.depth)
+if args.sparse_funnel and not args.ve_layers:
+    args.ve_layers = ",".join(
+        str(x) for x in compute_sparse_global_layers(args.depth, args.n_global, args.chirp_gamma)
+    )
+
 user_config = vars(args).copy()  # for logging
 # -----------------------------------------------------------------------------
 # Compute init and wandb logging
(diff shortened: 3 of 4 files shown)


scripts/base_train.py
```diff
@@ -52,6 +52,21 @@
...
+parser.add_argument("--rope-base", type=int, default=-1, help="RoPE base frequency (-1 = baked-in sparse default)")
+parser.add_argument("--bigram-vocab-size", type=int, default=-1, help="hashed bigram embedding buckets (-1 = baked-in sparse default, 0 disables)")
+parser.add_argument("--bigram-dim", type=int, default=-1, help="bigram embedding dim before projection (-1 = baked-in sparse default)")
+parser.add_argument("--gated-attn", dest="gated_attn", action="store_true", help="enable lightweight per-head attention output gates")
+parser.add_argument("--no-gated-attn", dest="gated_attn", action="store_false", help="disable lightweight per-head attention output gates")
+parser.set_defaults(gated_attn=None)
+parser.add_argument("--attn-gate-channels", type=int, default=-1, help="embedding channels used to predict attention gates (-1 = baked-in sparse default)")
+parser.add_argument("--attn-gate-layers", type=str, default="", help="explicit attention-gate layer indices, comma-separated")
+parser.add_argument("--ve-layers", type=str, default="", help="explicit value-residual layer indices, comma-separated")
```
svlandeg, line 68: Oof, these are a lot of extra settings! But in the end you only set `--sparse-funnel` for your best run, right?

scripts/base_train.py
```diff
@@ -111,9 +161,12 @@
     else:
         print0("WARNING: Flash Attention 3 not available, using PyTorch SDPA fallback")
     print0("WARNING: Training will be less efficient without FA3")
-    if args.window_pattern != "L":
+    if not args.sparse_funnel and args.window_pattern != "L":
         print0(f"WARNING: SDPA has no support for sliding window attention (window_pattern='{args.window_pattern}'). Your GPU utilization will be terrible.")
         print0("WARNING: Recommend using --window-pattern L for full context attention without alternating sliding window patterns.")
+    elif args.sparse_funnel:
```
svlandeg: Nitpick: these statements would be nicer if you do first
`if args.sparse_funnel:`

and then 
`elif args.window_pattern != "L":`

svlandeg: Actually, hold on, shouldn't the condition also check for SDPA fallback? Now it is printing the new warning whenever `args.sparse_funnel` is `True`.

Anyway, these are details that can be discussed if the PR is accepted 🙂

nanochat/engine.py
```diff
@@ -177,13 +182,9 @@ def generate(self, tokens, num_samples=1, max_tokens=None, temperature=1.0, top_
...
-        # NOTE: setting the dtype here and in this way is an ugly hack.
-        # Currently the repo assumes that cuda -> bfloat16 and everything else -> float32.
-        # We need to know the dtype here to call __init__ on KVCache and pre-allocate its tensors.
-        # As a quick hack, we're making generate() function inherit and know about this repo-wise assumption.
-        # I think there has to be a bigger refactor to deal with device/dtype tracking across the codebase.
-        # In particular, the KVCache should allocate its tensors lazily
-        dtype = torch.bfloat16 if device.type == "cuda" else torch.float32
+        # Infer KV-cache dtype from the model instead of hard-coding by device type.
+        dtype = getattr(getattr(self.model, "transformer", None), "wte", None)
+        dtype = dtype.weight.dtype if dtype is not None else (torch.bfloat16 if device.type == "cuda" else torch.float32)
```
svlandeg: Flagging that this is a refactor that needs reviewing separate from the leaderboard submission

README.md
```diff
@@ -20,6 +20,7 @@ Presently, the main focus of development is on tuning the pretraining stage, whi
 | 4 | 2.02 | 0.71854 | 0.2571 | change dataset to NVIDIA ClimbMix | Mar 4 2026 | 324e69c | @ddudek @karpathy |
 | 5 | 1.80 | 0.71808 | 0.2690 | autoresearch [round 1](https://x.com/karpathy/status/2031135152349524125) | Mar 9 2026 | 6ed7d1d | @karpathy |
 | 6 | 1.65 | 0.71800 | 0.2626 | autoresearch round 2 | Mar 14 2026 | a825e63 | @karpathy |
+| 7 | 1.60 | 0.71773 | 0.2609 | d24 sparse-funnel reference recipe, reproduced twice | Mar 26 2026 | pending | pending |
```
svlandeg: Another nitpick: this entry shows different values than the ones in the title of the PR - probably best to pick one of the two runs as the "submission".

jn2clark: thank you, ill correct that. it needs to be the average rather than the individual.

nanochat/engine.py
jn2clark: thanks, ill remove this then

scripts/base_train.py
jn2clark: correct, the run script has the command. ill remove these args and bake it in like the other previous runs have been doing

runs/speedrun.sh
```diff
@@ -69,8 +69,19 @@ python -m scripts.tok_eval
 echo "Waiting for dataset download to complete..."
 wait $DATASET_DOWNLOAD_PID
 
-# d24 model (slightly undertrained to beat GPT-2 => decrease data:params ratio from compute optimal 10.5 (default) to 8)
-torchrun --standalone --nproc_per_node=8 -m scripts.base_train -- --depth=24 --target-param-data-ratio=8 --device-batch-size=16 --fp8 --run=$WANDB_RUN
+# d24 sparse-funnel reference model
+torchrun --standalone --nproc_per_node=8 -m scripts.base_train -- \
+  --depth=24 \
+  --model-tag=d24_sparse_funnel \
```
svlandeg: If this is submitted to `master`, I wouldn't change the default `model-tag`?

svlandeg (commented):
I've been trying to replicate, but haven't been able to get hold of a 8xH100 yet.

On a 1xA100 (no fp8, no FA3, d14):

Image

so sparse funnel starts off better, but eventually gets to more or less the same `val_bpb` as `master`: both 0.838, but taking a significant longer time:

image

For a partial d24 run on a H100 we see a similar trend plotting `val_bpb` versus steps:

Image

but here there is no regression on tok/sec:

image

svlandeg (commented):
@jn2clark: will you be addressing the review comments? Currently, the PR is pretty bloated and it would be good to trim this down to the minimal edits necessary to run your leaderboard submission.

That said, it's still worrying to me that on an A100, there's such a huge regression in tok/sec. Even if we don't see this on an H100, it wouldn't be ideal to get like a 30-40% slowdown on other architectures.

jn2clark:
@svlandeg sorry about the delay, i am planning on updating this shortly. I am running some final experiments now, waiting on a 8xH100 for final testing. re A100, the FA3 and hopper specific optimizations do really limit the architectural choices. it would be good to clarify if the goal is as stated - faster time on 8xH100 for same or better CORE and valbp vs something broader (all GPU's) because this will change what can be done a lot

svlandeg:
Then again, there shouldn't be a regression on Ampere etc if the sparse funnel flag isn't used, right? Or is there? I haven't double checked yet.

jn2clark:
no there shouldn't be if its not used. @svlandeg i think what i will do is close this and re-open an updated one with all this feedback and some additional results based on these discussions.  thanks!

jn2clark:
@svlandeg I will update this tomorrow as I finished re-doing the experiments.

scripts/base_train.py
jn2clark: they are set now at the defaults, can remove if not needed and bake directly into the config

jn2clark:
@svlandeg I ran a bunch more experiments to validate the previous result and also re-check the implementation. There were a few translation errors in the previous version that should be resolved now. 

## Results Summary

Best individual run: Dense (before init fix) at **94.74 min**, step 5318, CORE **0.2581**
Best CORE overall: Sparse after (init fixed) at **96.76 min** with CORE **0.2721**
Earliest mean crossing: Sparse before init fix at step 5318 (**95.54 min** mean and CORE **0.2593**) 

### Top-5 Fastest Individual Runs (above GPT-2 baseline)

| Rank | Time (min) | Step | Side   | Init Phase | CORE  | Val BPB |
|------|------------|------|--------|------------|-------|---------|
| 1    | 94.74      | 5318 | dense  | before fix | 0.2581 | 0.7172  |
| 2    | 95.52      | 5318 | sparse | before fix | 0.2652 | 0.7180  |
| 3    | 95.52      | 5318 | sparse | after fix  | 0.2573 | 0.7179  |
| 4    | 95.53      | 5318 | sparse | before fix | 0.2575 | 0.7180  |
| 5    | 95.59      | 5368 | dense  | before fix | 0.2595 | 0.7168  |

---

### Top-5 Fastest Mean of Repeated Runs (above GPT-2 baseline)

| Rank | Time Mean (min) | Step | Side   | Init Phase | n | CORE Mean | CORE Min | CORE Max | Val BPB Mean |
|------|------------------|------|--------|------------|---|-----------|----------|----------|---------------|
| 1    | 95.54           | 5318 | sparse | before fix | 3 | 0.2593    | 0.2553   | 0.2652   | 0.7180        |
| 2    | 95.59           | 5368 | dense  | before fix | 2 | 0.2571    | 0.2547   | 0.2595   | 0.7168        |
| 3    | 95.64           | 5368 | dense  | after fix  | 2 | 0.2590    | 0.2529   | 0.2650   | 0.7170        |
| 4    | 95.91           | 5388 | dense  | before fix | 2 | 0.2591    | 0.2524   | 0.2657   | 0.7168        |
| 5    | 95.97           | 5388 | dense  | after fix  | 3 | 0.2583    | 0.2563   | 0.2597   | 0.7169        |

### Top 5 Individual Runs by CORE (above GPT-2 baseline)

| Rank | CORE  | Time (min) | Step | Side   | Init Phase | Val BPB |
|------|-------|------------|------|--------|------------|---------|
| 1    | 0.2721 | 96.76     | 5388 | sparse | after fix  | 0.7174  |
| 2    | 0.2690 | 96.44     | 5368 | sparse | after fix  | 0.7175  |
| 3    | 0.2687 | 97.30     | 5418 | sparse | after fix  | 0.7172  |
| 4    | 0.2657 | 95.90     | 5388 | dense  | before fix | 0.7168  |
| 5    | 0.2652 | 95.52     | 5318 | sparse | before fix | 0.7180  |

---

### Top 5 Group Means by CORE (above GPT-2 baseline)

| Rank | CORE Mean | Time Mean (min) | Step | Side   | Init Phase | n | Val BPB Mean |
|------|------------|------------------|------|--------|------------|---|---------------|
| 1    | 0.2639     | 99.14           | 5568 | dense  | after fix  | 2 | 0.7158        |
| 2    | 0.2628     | 96.42           | 5368 | sparse | after fix  | 3 | 0.7175        |
| 3    | 0.2617     | 97.26           | 5418 | sparse | after fix  | 3 | 0.7172        |
| 4    | 0.2615     | 96.71           | 5388 | sparse | after fix  | 3 | 0.7174        |
| 5    | 0.2602     | 99.90           | 5568 | sparse | after fix  | 2 | 0.7163        |


So the first crossing happens at 94.74 minutes and the mean group is at 95.54 minutes.  Note, I ran training both before and after the init fix listed here https://github.com/karpathy/nanochat/pull/686.
It did appear in general to improve the final CORE but not necessarily the valbp. The data are distinguished in the plots (below) and tables as being before or after the init fix. I suspect there is a better init scheme but this wasn't explored and the actual inits for the successful runs before the fix were unfortunately not recorded. 

## Expanded Results

In terms of results, here is a plot showing the dense baseline vs the proposed sparse (this PR) as a function of training time/number of iterations. The plot shows the mean at each point as the training run was repeated multiple times to assess the variability. The upper/lower bounds of the whiskers represent the max/min within that group. 

### Iterations
record_series_panel_step_all

### Time

record_series_panel_time_all


The sparse version does better at final CORE compared to the current dense baseline. However this is not true for the final valbp, where we see the sparse is consistently higher. One thing to note is that I am using the valbp and CORE values ran using the eval script after training. I think there are three different places these values are calculated and do appear to be subtly different - it would be good to clarify what the source of truth should be. 

If we look at the individual training runs, then we can plot out the frontier in terms of time to GPT-2 CORE across both dense and sparse variants. The highest CORE came from the sparse after init fix variant, but the fastest crossing of the GPT-2 baseline came from the dense before init fix variant. The earliest mean crossing of the GPT-2 baseline was from the sparse before init fix variant. 

record_curated_core_vs_time_pareto

record_curated_valbpb_vs_time_pareto

jn2clark:
@svlandeg i'm going to close this as I think there are some better improvements in later PR's.

PR #670, opened 2026-03-26
[codex] Fix calculator timeout concurrency in engine

unclejobs-ai:
## Summary
This patch fixes the calculator helper in `nanochat/engine.py`, which previously relied on `signal.signal(SIGALRM, ...)` and `signal.alarm()` to time out expression evaluation.

Those APIs are process-global, so concurrent chats could interfere with each other by overwriting the active alarm handler or cancelling another request's timeout. In practice that made the calculator path nondeterministic under concurrency.

The fix removes the signal-based timeout/eval flow and replaces it with a restricted AST evaluator that only accepts the calculator feature's supported operations:
- numeric `+`, `-`, `*`, `/`
- unary `+` and `-`
- string `.count(...)`

This keeps the feature behavior while avoiding process-global timeout state in the request path.

## Validation
I reran the focused engine tests and the full local test suite:

```bash
cd /tmp/nanochat-review && PYTHONPATH=/tmp/nanochat-review ./.venv/bin/python -m pytest tests/test_engine.py -q
cd /tmp/nanochat-review && PYTHONPATH=/tmp/nanochat-review ./.venv/bin/python -m pytest -q
```

Results:
- `9 passed` for `tests/test_engine.py`
- `14 passed, 10 skipped` for the full suite

I also added a regression test covering calculator evaluation off the main thread.


unclejobs-ai:
Closing this out since this change is not intended for upstream contribution.

PR #673, opened 2026-03-27
Have fastapi and uvicorn as optional dependencies only to run chat_web.py

svlandeg:
- Takes `fastapi` and `uvicorn` out of the standard list of dependencies
- Make them available by running `uv sync --extra web`
- in `speedrun.sh` and `runcpu.sh`, only suggest running them when the user wants to deploy the chat interface
- print a helpful message when `chat_web.py` is executed without `fastapi` in the environment

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ For development (adds pytest, matplotlib, ipykernel, transformers, etc.):
 uv sync --extra gpu --group dev
 ```
 
-If you plan on running scripts.chat_web:
+If you plan on running `scripts.chat_web`:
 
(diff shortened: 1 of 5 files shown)


hitchho (approved):
Clean separation of web dependencies. The module-level try/except with a clear exit message is the right approach. Shell script comments are helpful.

svlandeg (commented):
Closing, this has become obsolete after https://github.com/karpathy/nanochat/commit/da32e1d657b62cc1110f20f86fbc301d6189726e

PR #674, opened 2026-03-27
Make transformers and tokenizers optional dependencies in the "hf" extra

svlandeg:
Move `transformers` and `tokenizers` into a separate `hf` extra. Most speedruns work just fine without having to install them. This reduces footprint of the repo & minimizes supply chain attack risk.

To avoid import errors on normal runs (that don't need HF), we then need to move the relevant import statements to inline within the functions that use them.

### DIFF
--- a/nanochat/tokenizer.py
+++ b/nanochat/tokenizer.py
@@ -31,10 +31,6 @@ SPLIT_PATTERN = r"""'(?i:[sdmt]|ll|ve|re)|[^\r\n\p{L}\p{N}]?+\p{L}+|\p{N}{1,2}| 
 
 # -----------------------------------------------------------------------------
 # Generic GPT-4-style tokenizer based on HuggingFace Tokenizer
-from tokenizers import Tokenizer as HFTokenizer
-from tokenizers import pre_tokenizers, decoders, Regex
-from tokenizers.models import BPE
-from tokenizers.trainers import BpeTrainer
 
 class HuggingFaceTokenizer:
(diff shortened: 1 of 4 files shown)


hitchho (approved):
Aligns well with the direction of trimming dependencies. The lazy import pattern with `_try_import()` is clean. Minor nit: `train_from_iterator` skips the `_try_import()` guard so you'd get a raw ImportError instead of the friendly message there.

nanochat/tokenizer.py
```diff
@@ -31,33 +31,45 @@
...
-from tokenizers.trainers import BpeTrainer
 
 class HuggingFaceTokenizer:
     """Light wrapper around HuggingFace Tokenizer for some utilities"""
 
     def __init__(self, tokenizer):
         self.tokenizer = tokenizer
 
+    @staticmethod
+    def _try_import():
```
svlandeg, line 42: Note for Andrej: this helper function is just to improve UX, if we want to keep the file/diff smaller we can also just let the import error be shown to the user as such.

svlandeg:
Thanks for the review @hitchho ! I actually think that `HuggingFaceTokenizer.train_from_iterator` is currently an unused code path, but just for consistency I did add a `_try_import` call there, too.

karpathy:
I might just delete all of huggingface entirely, and get rid of this functionality... thinking

svlandeg:
Closing, became obsolete after https://github.com/karpathy/nanochat/commit/41865401f73ff1c5321ae53297bceb2b78d4c8b4.

PR #682, opened 2026-03-30
inital support for intel GPU

suspicious-pineapple:
havent done extensive correctness testing, but base_train seems to run fine on an Intel Arc A770 16gb

followed these steps to get torch to work with xpu:
<https://docs.pytorch.org/docs/stable/notes/get_start_xpu.html>

hope this isnt considered bloat, i figured the changes i needed are non-invasive enough to include upstream

draft for now since i'm waiting for a few test runs to finish to completion in case there are any issues i haven't encountered yet.
..as it turns out A770 is not the strongest card so it might be a moment.

(hardware utilisation is probably also not fantastic, the fans hardly spin up, but at least it runs)

### DIFF
--- a/nanochat/common.py
+++ b/nanochat/common.py
@@ -27,6 +27,10 @@ def _detect_compute_dtype():
         # fp16 training requires GradScaler (not yet implemented), so fall back to fp32.
         # Users can still force fp16 via NANOCHAT_DTYPE=float16 if they know what they're doing.
         return torch.float32, f"auto-detected: CUDA SM {capability[0]}{capability[1]} (pre-Ampere, bf16 not supported, using fp32)"
+    if torch.xpu.is_available():
+        if torch.xpu.is_bf16_supported():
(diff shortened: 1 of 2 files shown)


suspicious-pineapple:
base_train and chat_sft seem to run without errors, loss goes down as one would hope

Inference seems a bit sketchy. sample_every during base training seems to work, but chat_cli tends to crash, roughly like this:

```
User: how are you

Assistant: AssertHandler::printMessage
/pytorch/third_party/torch-xpu-ops/src/ATen/native/xpu/sycl/TensorCompareKernels.cpp:180: operator(): global id: [0,0,0], local id: [0,0,0] Assertion `input_[0] != 0` failed.
Aborted (core dumped)
```

will have to figure out what the inference calls during pretraining do different from chat_cli.

when I got inference to run, the base model seemed fine, but the sft checkpoint didn't output anything, at least nothing resembling the usual gibberish of a D6 model.

andomeder:
Nice work, will try it(training) on B580 and see how it goes

PR #683, opened 2026-03-31
Small doc fixes for consistency

svlandeg:
All of these edits are for consistency across the code base:

- Update speedrun from 3h to 1.5h in the docs
- Fix comment in `runcpu` that referred to the wrong depth
- Add `-i` to example `chat_eval` commands, as it's required and will crash if not specified
- Fix fwe-train and fwe-val labels to Climbmix, taken from PR #764 
- Fix link in `leaderboard.md`, taken from PR #690


hitchho (approved):
All changes are factually correct — depth comment, training time, and the missing print0() call. Clean and minimal.

PR #684, opened 2026-04-01
Fix speedrun.sh: add uv to PATH for container instances

### LINKED ISSUE #529: speedrun.sh updates
A couple suggestions for `speedrun.sh` and the speedrun requirements:

**1. Installer path**

The `speedrun.sh` script seems to not work on container instances--it errors saying that it can't find `uv`. The added `export PATH` line below fixes it. 

```python
# install uv (if not already installed)
command -v uv &> /dev/null || curl -LsSf https://astral.sh/uv/install.sh | sh
# add uv to PATH (the installer puts it in ~/.local/bin)
export PATH="$HOME/.local/bin:$PATH"
# create a .venv local virtual environment (if it doesn't exist)
[ -d ".venv" ] || uv venv
# install the repo dependencies
uv sync --extra gpu
# activate venv so that `python` uses the project's venv instead of system python
source .venv/bin/activate
```

**2. Validation shard**

Currently, the validation shard is chosen as the last shard downloaded. Following speedrun.sh, this is 370. 

I ran some tests (partially shared in a comment on the batch size PR) where I got very low variance on the validation `bpb` across 8 runs using shard 370. 

However, when I ran a quick test of calculating bpb for a single model on multiple different shards, the variance was "much higher" (I didn't capture or share these results, though, because I wasn't running the baseline model).

I wanted to point this out in case it's a concern for maintaining a fair comparison across speedrun submissions. If so, it might be helpful to include a note about this, or even isolate shard 370 as the dedicated validation shard.

JoseRodriguez26:
## Summary
- Adds `export PATH="$HOME/.local/bin:$PATH"` after the `uv` install step in `speedrun.sh`
- The `uv` installer places the binary in `~/.local/bin`, which is not on `PATH` by default in container environments, causing the script to fail immediately after install

Fixes #529 (installer path fix)

## Test plan
- [ ] Run `speedrun.sh` on a fresh container instance where `~/.local/bin` is not on PATH
- [ ] Verify `uv` commands succeed after the install step

### DIFF
--- a/runs/speedrun.sh
+++ b/runs/speedrun.sh
@@ -20,6 +20,8 @@ mkdir -p $NANOCHAT_BASE_DIR
 
 # install uv (if not already installed)
 command -v uv &> /dev/null || curl -LsSf https://astral.sh/uv/install.sh | sh
+# add uv to PATH (the installer puts it in ~/.local/bin)
+export PATH="$HOME/.local/bin:$PATH"
 # create a .venv local virtual environment (if it doesn't exist)
 [ -d ".venv" ] || uv venv
 # install the repo dependencies


svlandeg (commented):
Duplicate of #656, closing 🙏

PR #686, opened 2026-04-03
Initialize smear and backout lambdas in init_weights()

marcinbogdanski:
Currently `smear_lambda` and `backout_lambda` are declared in `GPT.__init__()`, but they are not initialized in `init_weights()`.

In `base_train.py` model is initialized on `meta` device and materialized with `to_empty()`, so these params may end up not being initialized.

In my testing, both `smear_lambda` and `backout_lambda` seem to be implicitly initialized to `0.0` by some mechanism in PyTorch, so this issue doesn't present itself easily. Adding `torch.use_deterministic_algorithms(True)` before model init in `base_train.py` makes issue repoducible on my setup. In this case both values become `nan`, which in turn results in training failure (loss=nan).

To be aligned with the constructor, this PR initalizes to `0.0` and `0.2` respectively.

### DIFF
--- a/nanochat/gpt.py
+++ b/nanochat/gpt.py
@@ -237,6 +237,8 @@ class GPT(nn.Module):
         # Decaying x0 init: earlier layers get more input embedding blending
         for i in range(n_layer):
             self.x0_lambdas.data[i] = 0.20 - (0.15 * i / max(n_layer - 1, 1))
+
+        # Smear/backout scalars and smear gate must be explicitly initialized 
         torch.nn.init.zeros_(self.smear_lambda)
         torch.nn.init.constant_(self.backout_lambda, 0.2)
         torch.nn.init.uniform_(self.smear_gate.weight, 0.0, 0.02)
--- a/nanochat/gpt.py
+++ b/nanochat/gpt.py
@@ -237,6 +237,8 @@ class GPT(nn.Module):
         # Decaying x0 init: earlier layers get more input embedding blending
         for i in range(n_layer):
             self.x0_lambdas.data[i] = 0.20 - (0.15 * i / max(n_layer - 1, 1))
+        self.smear_lambda.fill_(0.0)
+        self.backout_lambda.fill_(0.2)
 
         # Value embeddings (init like c_v: uniform with same std)
         for ve in self.value_embeds.values():
--- a/nanochat/gpt.py
+++ b/nanochat/gpt.py
@@ -237,8 +237,9 @@ class GPT(nn.Module):
         # Decaying x0 init: earlier layers get more input embedding blending
         for i in range(n_layer):
             self.x0_lambdas.data[i] = 0.20 - (0.15 * i / max(n_layer - 1, 1))
-        self.smear_lambda.fill_(0.0)
-        self.backout_lambda.fill_(0.2)
+        torch.nn.init.zeros_(self.smear_lambda)
+        torch.nn.init.constant_(self.backout_lambda, 0.2)
+        torch.nn.init.uniform_(self.smear_gate.weight, 0.0, 0.02)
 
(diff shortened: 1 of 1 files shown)


marcinbogdanski:
Good point from Sofie: the `smear_gate.weight` should also be initialized in `init_weights`.

svlandeg:
Yep just for reference and to repeat part of the conversation on Discord :)

I ran into this a few days ago and Claude fixed it for me like so: https://github.com/svlandeg/nanochat/commit/71460c166d70818585d5bbcc9b5c659453fa47b0. I hadn't reviewed this or tested this thoroughly yet though. IIRC this came up in a cpu run, producing nans like you said. 

Thanks for the PR! We should definitely get this fixed.

hitchho (approved):
Clean 2-line fix. Values match the constructor, placement is consistent with how other lambdas are initialized. LGTM.

svlandeg:
> Clean 2-line fix. Values match the constructor, placement is consistent with how other lambdas are initialized. LGTM.

You don't think `smear_gate.weight` should be initialized too?

nanochat/gpt.py
```diff
@@ -237,6 +237,8 @@ def init_weights(self):
         # Decaying x0 init: earlier layers get more input embedding blending
         for i in range(n_layer):
             self.x0_lambdas.data[i] = 0.20 - (0.15 * i / max(n_layer - 1, 1))
+        self.smear_lambda.fill_(0.0)
+        self.backout_lambda.fill_(0.2)
```
svlandeg:
```suggestion
        torch.nn.init.zeros_(self.smear_lambda)
        torch.nn.init.constant_(self.backout_lambda, 0.2)
        torch.nn.init.uniform_(self.smear_gate.weight, 0.0, 0.02)
```

svlandeg (commented):
I suggest to use `torch.nn.init` as we're in `init_weights` and all other initializations use `torch.nn.init` as well.

Also, Claude suggests to set `smear_gate.weight` to an initial uniform distribution as well.

nanochat/gpt.py
```diff
@@ -237,6 +237,9 @@ def init_weights(self):
         # Decaying x0 init: earlier layers get more input embedding blending
         for i in range(n_layer):
             self.x0_lambdas.data[i] = 0.20 - (0.15 * i / max(n_layer - 1, 1))
+        torch.nn.init.zeros_(self.smear_lambda)
```
svlandeg, line 242:
```suggestion

        # Smear/backout scalars and smear gate must be explicitly initialized 
        torch.nn.init.zeros_(self.smear_lambda)
```

svlandeg (commented):
When I run a CPU `base_train` on my local machine, half of the time it starts producing `nan`'s from the very first step. Dissecting the problem together with Claude got me to a very similar fix as what this PR originally suggested. With this fix, I can't produce any `nan`'s more on the CPU run.

I'm also now running a larger d14 on the A100 to compare with `master`, will report back when it's done.

marcinbogdanski:
Few quick tests with `--device-type=cpu` seem to confirm that *without* fix `smear_lambda` / `backout_lambda` are left uninitialized. with fix they are ok. I suspect `smear_gate` is similar

Example w/o fix and `--device-type=cpu`:
```bash
model.smear_lambda = -306407059892182187780341760.0000 | model.backout_lambda = -306407059892182187780341760.0000
```

svlandeg (commented):
The A100 run confirmed my assumption: on GPU this fix creates a slightly lower `val_bpb` in the beginning because there's appropriate initialization. The final `val_bpb` is equal to `master` (0.838), the CORE score is even slightly better (0.1632 vs 0.159). So, this validates the fix for a proper GPU run.

Image

karpathy:
doh, thank you for the fix.

Issue #550, opened 2026-02-21
[BUG] core_eval.py throws ValueError when core-metric-max-per-task < num_fewshot


suraj-self:
### Bug Description
When running `scripts.base_train` with a severely restricted task size (e.g., `--core-metric-max-per-task=1` for rapid testing), the evaluation suite crashes with a `ValueError` in `core_eval.py`. This seems to happen because tasks like Jeopardy hardcode `num_fewshot=10`, causing `random.sample` to request more items than exist in the restricted population.

### System Config
- **OS:** Linux (Ubuntu 22.04)
- **GPU:** NVIDIA RTX 3060 (6GB)

### Reproduction Command
```bash
python -m scripts.base_train \
  --depth=4 \
  --aspect-ratio=64 \
  --max-seq-len=512 \
  --num-iterations=1 \
  --core-metric-max-per-task=1
```
### Logs

```
(nanochat) ss@ss-Predator-PH315-53:~/nanochat$ python -m scripts.base_train   --depth=4   --aspect-ratio=64   --max-seq-len=512   --num-iterations=1   --core-metric-max-per-task=1

                                                       █████                █████
                                                      ░░███                ░░███
     ████████    ██████   ████████    ██████   ██████  ░███████    ██████  ███████
    ░░███░░███  ░░░░░███ ░░███░░███  ███░░███ ███░░███ ░███░░███  ░░░░░███░░░███░
     ░███ ░███   ███████  ░███ ░███ ░███ ░███░███ ░░░  ░███ ░███   ███████  ░███
     ░███ ░███  ███░░███  ░███ ░███ ░███ ░███░███  ███ ░███ ░███  ███░░███  ░███ ███
     ████ █████░░████████ ████ █████░░██████ ░░██████  ████ █████░░███████  ░░█████
    ░░░░ ░░░░░  ░░░░░░░░ ░░░░ ░░░░░  ░░░░░░   ░░░░░░  ░░░░ ░░░░░  ░░░░░░░░   ░░░░░
    
Autodetected device type: cuda
2026-02-21 12:58:23,246 - nanochat.common - INFO - Distributed world size: 1
2026-02-21 12:58:23,247 - nanochat.common - WARNING - Peak flops undefined for: NVIDIA GeForce RTX 3060 Laptop GPU, MFU will show as 0%
GPU: NVIDIA GeForce RTX 3060 Laptop GPU | Peak FLOPS (BF16): inf
[...]

PR #551, opened 2026-02-21
fix: clamp few-shot sample size to prevent ValueError in core_eval

suraj-self:
Fixes #550 

**Impact:**
Prevents `ValueError: Sample larger than population` when a user restricts dataset size for testing (e.g., `--core-metric-max-per-task=1`) on tasks requiring multiple few-shots (like Jeopardy 10-shot).

**Change:**
Wrapped `num_fewshot` in a `min()` function against `len(available_indices)` to safely bound the random sampler.

### Log
```
(nanochat) ss@ss-Predator-PH315-53:~/nanochat$ python -m scripts.base_train   --depth=4   --aspect-ratio=64   --max-seq-len=512   --num-iterations=1   --core-metric-max-per-task=1

                                                       █████                █████
                                                      ░░███                ░░███
     ████████    ██████   ████████    ██████   ██████  ░███████    ██████  ███████
    ░░███░░███  ░░░░░███ ░░███░░███  ███░░███ ███░░███ ░███░░███  ░░░░░███░░░███░
     ░███ ░███   ███████  ░███ ░███ ░███ ░███░███ ░░░  ░███ ░███   ███████  ░███
     ░███ ░███  ███░░███  ░███ ░███ ░███ ░███░███  ███ ░███ ░███  ███░░███  ░███ ███
     ████ █████░░████████ ████ █████░░██████ ░░██████  ████ █████░░███████  ░░█████
    ░░░░ ░░░░░  ░░░░░░░░ ░░░░ ░░░░░  ░░░░░░   ░░░░░░  ░░░░ ░░░░░  ░░░░░░░░   ░░░░░
    
Autodetected device type: cuda
2026-02-21 12:33:28,161 - nanochat.common - INFO - Distributed world size: 1
2026-02-21 12:33:28,161 - nanochat.common - WARNING - Peak flops undefined for: NVIDIA GeForce RTX 3060 Laptop GPU, MFU will show as 0%
GPU: NVIDIA GeForce RTX 3060 Laptop GPU | Peak FLOPS (BF16): inf
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
WARNING: Flash Attention 3 not available, using PyTorch SDPA fallback
WARNING: Training will be less efficient without FA3
WARNING: SDPA has no support for sliding window attention (window_pattern='SSSL'). Your GPU utilization will be terrible.
[...]

PR #688, opened 2026-04-05
fix: cap few-shot count when dataset is subsampled below num_fewshot

hiSandog:
## Problem

When running `base_train` with `--core-metric-max-per-task` set to a small value (e.g. 5), tasks with a high `num_fewshot` like Jeopardy (10-shot) crash with:

```
ValueError: Sample larger than population or is negative
```

This happens because the dataset is subsampled to `max_per_task` items (line 156-157 of `base_eval.py`), but the few-shot sampling in `core_eval.py` still tries to draw `num_fewshot` examples from the now-smaller pool.

## Fix

Cap the effective few-shot count to the number of available indices (excluding the current evaluation example). If the cap is lower than the configured `num_fewshot`, print a one-line warning so the user knows the task is running with fewer examples than intended.

## Impact

- Minimal: only affects evaluation code path
- No behavior change when dataset is large enough (normal case)
- Enables rapid CORE eval testing with small `--core-metric-max-per-task` values

Relates to #550.

### DIFF
--- a/nanochat/core_eval.py
+++ b/nanochat/core_eval.py
@@ -177,7 +177,11 @@ def evaluate_example(idx, model, tokenizer, data, device, task_meta):
     if num_fewshot > 0:
         rng = random.Random(1234 + idx)
         available_indices = [i for i in range(len(data)) if i != idx]
-        fewshot_indices = rng.sample(available_indices, num_fewshot)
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
Duplicate of #551, closing

PR #691, opened 2026-04-08
PyTorch 2.11 + FlexAttention + SDPA support (B300 performance)

ademeure:
Start of my work trying to get maximum performance on Blackwell on nanochat & autoresearch :)

---

Same as autoresearch PR: https://github.com/karpathy/autoresearch/pull/504

PyTorch 2.11 is required for good performance on B300 and also adds support for FlexAttention via Flash Attention 4 This provides slightly better loss-per-minute than SDPA full-causal on B300 (and massively faster than SDPA sliding) while keeping support for FA3 on H100.

uv has just added support for torch-backend = "auto" which will result in 2.11+cu130 on Blackwell, 2.9+cu128 on older GPUs, and CPU PyTorch if there is no CUDA GPU - but this is very new, so this does create a risk it might not automatically work for e.g. H100 stuck on older drivers with an older uv. The nanochat version has more user control which might be a better approach for now, depending on the compatibility vs simplicity trade-off you want.

### DIFF
--- a/LOCAL_STATE.md
+++ b/LOCAL_STATE.md
@@ -1,0 +1,35 @@
+# Local State — nanochat (karpathy fork)
+
+Documented 2026-04-09 before machine teardown.
+
+## Branch: fa3-flex-sdpa (current)
+- Tracking: `fork/fa3-flex-sdpa` (ademeure/nanochat) — pushed and up to date
+- 1 commit ahead of upstream master: `3d0dec5 FA3/FlexAttention/SDPA attention + PyTorch 2.11/CUDA 13.0`
+
+## Branch: pytorch-2.11-cu130
+- Tracking: `fork/pytorch-2.11-cu130` — pushed and up to date
+- 2 commits ahead of master
+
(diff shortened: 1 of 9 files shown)


hiSandog:
One thing I noticed in the new backend selection logic: `ATTN_BACKEND` can report `flex`, but `flash_attn_func()` still routes full-causal attention through plain SDPA.

In `nanochat/flash_attention.py`, the new branch returns early on `if w < 0 or w >= T:` with `F.scaled_dot_product_attention(..., is_causal=True)`, and the `if USE_FLEX:` block only runs for sliding-window mode. So on the common `window_pattern="L"` full-context path, `scripts/base_train.py` / `scripts/chat_sft.py` will print `Attention backend: flex` even though the actual compute path is still SDPA.

If the intent is that FlexAttention only accelerates sliding-window attention, I think the status messages / README should say that explicitly. Otherwise this looks like a behavior gap: Blackwell users following the new `uv sync` path may expect full-causal runs to benefit from FlexAttention, but this implementation never exercises it there.

PR #695, opened 2026-04-11
[codex] Fix chat validation message and engine KV cache dtype

missflash:
## What changed
- fixed the chat API validation message so it matches the supported roles
- made Engine KV-cache allocation follow the repo-wide `COMPUTE_DTYPE`
- added narrow regression tests for both fixes

## Why
- the web API error text claimed `system` messages were allowed even though the handler rejects them
- inference cache dtype was using a hardcoded device heuristic instead of the existing global dtype policy

## Validation
- reviewer pass: no clear regression found in the touched paths
- runtime tests were not executed here because this environment does not have `python`, `pytest`, or `uv` installed

### DIFF
--- a/nanochat/engine.py
+++ b/nanochat/engine.py
@@ -17,7 +17,7 @@ import signal
 import warnings
 from contextlib import contextmanager
 from collections import deque
-from nanochat.common import compute_init, autodetect_device_type
+from nanochat.common import compute_init, autodetect_device_type, COMPUTE_DTYPE
 from nanochat.checkpoint_manager import load_model
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


svlandeg (commented):
> runtime tests were not executed here because this environment does not have python, pytest, or uv installed

Please do run the tests, especially if you're introducing new ones 🙏

missflash:
@svlandeg Ran the tests in a fresh `uv`-managed environment after syncing dependencies.

Validated:
- `tests/test_regressions.py`
- `tests/test_engine.py`

Result: `10 passed`

Thanks for the reminder - I'll keep including the actual test command and result in validation notes.

svlandeg:
Closing in favor of #795 🙏

PR #702, opened 2026-04-12
Exclusive Self Attention

RohanKhanBD:
added exclusive self attention from https://arxiv.org/pdf/2603.09078

PR #708, opened 2026-04-15
Implement Exclusive Self Attention in the forward pass.

Mithun-033:
Adds Exclusive Self-Attention (XSA) from “Exclusive Self Attention” by Shuangfei Zhai (Apple) , which removes the component of the attention output aligned with its own value vector to eliminate attention similarity bias and improve context modeling. This is a 2 line change in the Attention Block


babaic:
@Mithun-033 pls add tests, tnx

svlandeg:
Actually, more importantly: run this code, and provide us with the results of the run (val_bpb, timing, CORE). Cf also #702.

PR #714, opened 2026-04-16
feat(chat-api): conversation orchestration + SSE streaming proxy (#6)

manmohan659:
## Summary

Implements the Chat API Service (GitHub Issue #6) — the orchestration layer between the frontend, the auth service, and the inference service.

- FastAPI + SQLAlchemy 2.0 async + asyncpg service that owns `conversations` and `messages` in PostgreSQL and streams assistant responses back over SSE.
- Every endpoint outside `/api/health` runs through `require_user`, which calls the auth service `POST /auth/validate` with the shared `X-Internal-API-Key` and caches results in a 5-minute TTL cache.
- Main endpoint `POST /api/conversations/{id}/messages`: validates ownership, persists the user message, builds a history slice bounded by `MAX_CONVERSATION_HISTORY` + `MAX_TOKEN_BUDGET`, opens an httpx streaming POST to the inference service, pipes the SSE events unchanged to the client via `sse-starlette`, and persists the accumulated assistant response (with `token_count` + `inference_time_ms`) after the client disconnects. First-message auto-titling is applied in the same post-stream step.
- `POST /api/conversations/{id}/regenerate` drops the last assistant message and re-streams using the same machinery.
- `GET /api/models` and `POST /api/models/swap` proxy to the inference service (swap requires `is_admin` on the validated user).
- Structured JSON logging via `structlog`; `trace_id` and `user_id` are attached via ContextVars to every log line for the request lifecycle.

### Security posture

- All conversation/message queries filter by `Conversation.user_id == authenticated_user_id`; cross-user access returns 404 rather than leaking existence.
- Bearer tokens are validated against the auth service, not decoded locally, so key rotation requires no changes here.
- CORS is locked to `FRONTEND_URL`.

## Test plan

- [x] `pytest` passes locally (16/16) — CRUD, streaming send + persistence, regenerate, user scoping, models proxy admin gate, stream proxy error path.
- [ ] Smoke-test end-to-end with the auth service + inference service + Postgres via `docker-compose`.
- [ ] Verify SSE stream is forwarded without buffering behind a real reverse proxy.
- [ ] Exercise `Authorization: Bearer <expired>` to confirm 401 pass-through and that the TTL cache does not cache 401s.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


manmohan659:
Apologies — accidentally opened against upstream. Reopening against my fork.

Issue #578, opened 2026-02-28
ZeroDivisionError in scaling law math when --target-param-data-ratio is set to 0


suraj-self:
Setting ```--target-param-data-ratio 0``` causes a crash during the auto-computation of the optimal batch size. Since this ratio is used as a denominator in the Chinchilla scaling math, a zero value results in a ```ZeroDivisionError``` instead of a graceful exit or a helpful error message.

### Steps to reproduce:
```
python -m scripts.base_train --target-param-data-ratio 0
```
### Logs

```
(nanochat) ss@ss-Predator-PH315-53:~/nanochat$ python -m scripts.base_train   --depth 1   --aspect-ratio 1   --target-param-data-ratio 0

                                                       █████                █████
                                                      ░░███                ░░███
     ████████    ██████   ████████    ██████   ██████  ░███████    ██████  ███████
    ░░███░░███  ░░░░░███ ░░███░░███  ███░░███ ███░░███ ░███░░███  ░░░░░███░░░███░
     ░███ ░███   ███████  ░███ ░███ ░███ ░███░███ ░░░  ░███ ░███   ███████  ░███
     ░███ ░███  ███░░███  ░███ ░███ ░███ ░███░███  ███ ░███ ░███  ███░░███  ░███ ███
     ████ █████░░████████ ████ █████░░██████ ░░██████  ████ █████░░███████  ░░█████
    ░░░░ ░░░░░  ░░░░░░░░ ░░░░ ░░░░░  ░░░░░░   ░░░░░░  ░░░░ ░░░░░  ░░░░░░░░   ░░░░░
    
Autodetected device type: cuda
2026-02-28 13:00:51,784 - nanochat.common - INFO - Distributed world size: 1
2026-02-28 13:00:51,785 - nanochat.common - WARNING - Peak flops undefined for: NVIDIA GeForce RTX 3060 Laptop GPU, MFU will show as 0%
GPU: NVIDIA GeForce RTX 3060 Laptop GPU | Peak FLOPS (BF16): inf
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
WARNING: Flash Attention 3 not available, using PyTorch SDPA fallback
WARNING: Training will be less efficient without FA3
WARNING: SDPA has no support for sliding window attention (window_pattern='SSSL'). Your GPU utilization will be terrible.
[...]

PR #579, opened 2026-02-28
fix: guard target-param-data-ratio against zero to avoid ZeroDivision…

suraj-self:
This PR adds a sanity check for the ```--target-param-data-ratio``` argument. Fix for https://github.com/karpathy/nanochat/issues/578

- Prevents ```ZeroDivisionError``` when the user accidentally provides ```0```.
- Ensures the script fails fast with a clear message: ```"target-param-data-ratio must be positive (or -1 to disable)"```.
- Maintains compatibility with the ```-1``` flag used to disable scaling-based iterations.

### Logs

```
(nanochat) ss@ss-Predator-PH315-53:~/nanochat$ python -m scripts.base_train   --depth 1   --aspect-ratio 1   --target-param-data-ratio 0

                                                       █████                █████
                                                      ░░███                ░░███
     ████████    ██████   ████████    ██████   ██████  ░███████    ██████  ███████
    ░░███░░███  ░░░░░███ ░░███░░███  ███░░███ ███░░███ ░███░░███  ░░░░░███░░░███░
     ░███ ░███   ███████  ░███ ░███ ░███ ░███░███ ░░░  ░███ ░███   ███████  ░███
     ░███ ░███  ███░░███  ░███ ░███ ░███ ░███░███  ███ ░███ ░███  ███░░███  ░███ ███
     ████ █████░░████████ ████ █████░░██████ ░░██████  ████ █████░░███████  ░░█████
    ░░░░ ░░░░░  ░░░░░░░░ ░░░░ ░░░░░  ░░░░░░   ░░░░░░  ░░░░ ░░░░░  ░░░░░░░░   ░░░░░
    
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/ss/nanochat/scripts/base_train.py", line 82, in <module>
    assert args.target_param_data_ratio > 0 or args.target_param_data_ratio == -1, "target-param-data-ratio must be positive (or -1 to disable)"
AssertionError: target-param-data-ratio must be positive (or -1 to disable)
(nanochat) ss@ss-Predator-PH315-53:~/nanochat$
```

PR #720, opened 2026-04-17
Add validation for --target-param-data-ratio to prevent ZeroDivisionError

hiSandog:
## Summary

When `--target-param-data-ratio` is set to `0`, the training script crashes with a division error because the value is used as a multiplier for computing `target_tokens` and `D_REF`, which then causes division by zero in the batch size and weight decay calculations.

## Changes

1. **Early validation check**: Rejects invalid values (`<=0`) with a clear error message before any scaling law calculations begin, preventing cryptic downstream crashes.

2. **Fixed assertion bug**: The training horizon assertion used `> 0` comparisons with `-1` sentinel values. While this did not cause runtime issues in practice (since `-1 > 0` is `True` in Python), it is semantically incorrect and misleading. Changed to use proper sentinel-aware comparisons (`!= -1`).

## Reproduction

```bash
python -m scripts.base_train --target-param-data-ratio 0
```

Before: cryptic division error during scaling law computation
After: clear `ValueError: --target-param-data-ratio must be > 0 or -1 to disable, got 0.0`

Related: #578

### DIFF
--- a/scripts/base_train.py
+++ b/scripts/base_train.py
@@ -257,6 +257,10 @@ num_params = param_counts['total']
 num_flops_per_token = model.estimate_flops()
 print0(f"Estimated FLOPs per token: {num_flops_per_token:e}")
 
+# Validate training horizon arguments
+if args.target_param_data_ratio != -1 and args.target_param_data_ratio <= 0:
+    raise ValueError(f"--target-param-data-ratio must be > 0 or -1 to disable, got {args.target_param_data_ratio}")
+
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
Duplicate of https://github.com/karpathy/nanochat/pull/579, so closing this one. Feel free to leave a review on #579 instead!

svlandeg:
@hiSandog : this is the third duplicate PR you've opened for the same fix, please be more mindful 🙏

PR #721, opened 2026-04-18
add Flash Attention 2 as a middle tier between FA3 and SDPA

matt-langston:
on sm80+ non-Hopper GPUs (Blackwell, Ada, Ampere) with the flash-attn package installed, FA2 kernels replace the SDPA fallback. priority is FA3 > FA2 > SDPA. measured 28% faster than SDPA on GB10, and makes sliding-window attention fast on Blackwell (where FA3 is unavailable). no effect on H100: USE_FA3 wins whenever available so runs/speedrun.sh on 8xH100 runs the same kernels as before. tests/test_attention_fallback.py::TestFA2VsSDPA compares FA2 and SDPA output on any sm80+ GPU with flash-attn installed.

context: https://github.com/karpathy/nanochat/discussions/710 (the writeup was produced from my dgx-spark branch at https://github.com/matt-langston/nanochat/tree/dgx-spark, which carries my two PRs #721 and #722 plus a DGX-Spark-Bundle-specific speedrun script I kept separate)

Developed in pair-programming sessions with Claude (Anthropic). I reviewed, tested, and debugged every change myself on a 2x GB10 DGX Spark Bundle. I also tried to validate on 8xH100 via Lambda Cloud but couldn't get capacity after several days of polling. I understand all of the code in this PR.

### DIFF
--- a/nanochat/flash_attention.py
+++ b/nanochat/flash_attention.py
@@ -1,8 +1,9 @@
 """
-Unified Flash Attention interface with automatic FA3/SDPA switching.
+Unified Flash Attention interface with automatic FA3/FA2/SDPA switching.
 
 Exports `flash_attn` module that matches the FA3 API exactly, but falls back
 to PyTorch SDPA on non-Hopper GPUs (including Blackwell), MPS, and CPU.
+If the flash-attn package is installed, FA2 kernels are used instead of SDPA on sm80+ GPUs (FA3 wins if both are usable).
 
 Usage (drop-in replacement for FA3):
     from nanochat.flash_attention import flash_attn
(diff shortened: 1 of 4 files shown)


svlandeg (commented):
How does this compare to the community kernel FA3 (https://github.com/karpathy/nanochat/pull/540) which also runs on Ampere/Ada?

matt-langston:
> How does this compare to the community kernel FA3 (#540) which also runs on Ampere/Ada?

Thanks for flagging #540 - I followed the thread. Short version: the two PRs are complementary - they partition non-Hopper coverage. #540 extends FA3 to sm_80/86/89 (A100, RTX 30/40). My #721 adds FA2 as a middle tier for sm_80+ when no FA3 cubin is available. The user I'm targeting is the homelab contributor on consumer Blackwell (RTX 5090, GB10) who wants to tinker with nanochat without renting datacenter hardware.

I tested `kernels-community/flash-attn3` on GB10 (`sm_121a` i.e. my DGX Spark) yesterday. The kernel resolves the matching variant and loads (as expected), but first launch raises `no kernel image is available for execution on the device` (`flash_fwd_launch_template.h:188`). The pre-built cubins don't include `sm_121a`, matching your note that the build targets sm80/86/89/90. FA3 is built around Hopper-class warp specialization (TMA, WGMMA), and Blackwell's tensor-core and TMA semantics differ enough that a port would be source-level work, not a cubin update - to my knowledge no one has done that yet.

#721 is effectively the FA2 form of the PS you wrote on #540 ("try `has_kernel(flash-attn2)` as a fallback first when FA3 is not available"). Priority is FA3 > FA2 > SDPA, so on Ampere/Ada where #540 gives FA3, the FA2 branch stays dormant. On GB10 the FA2 branch is 28% faster than SDPA at d=24 with the default SSSL window pattern.

Happy to rebase #721 on top of #540 whenever your PR merges - the combined detection (varunneal Hopper > community Ampere/Ada > FA2 sm_80+ > SDPA) falls out of two short edits in `_load_flash_attention_3`. Whichever order works best for you.

PR #728, opened 2026-04-23
feat: double default and max generation budget

manmohan659:
## Summary
- chat-api `inference_default_max_tokens`: 512 → 1024
- Modal `serve.py` default: 512 → 1024, hard cap: 2048 → 4096

Longer answers (especially in thinking mode) were getting clipped mid-sentence because the 512-token default was spent before the final response finished. Doubling both the default and the cap gives thinking-mode room to reason *and* answer without the user having to pass `max_tokens` explicitly.

Modal side is already deployed; this PR merges the chat-api config change so CI rebuilds the image and rolls it to EC2.

## Test plan
- [ ] CI build + deploy-ec2 succeeds
- [ ] Long multi-section prompt in thinking mode on samosachaat.art completes with a natural ending (no mid-clause cutoff)
- [ ] Short non-thinking prompt latency unchanged


manmohan659:
Wrong repo — meant to target my own fork. Closing.

PR #734, opened 2026-04-27
Validate engine prompt token input

hiSandog:
## Summary
- reject empty prompt token lists before Engine.generate indexes into them
- validate every prompt token is an int instead of only checking the first element
- add a focused engine test for invalid prompt tokens

## Validation
- python -m py_compile nanochat/engine.py tests/test_engine.py
- python -m pytest tests/test_engine.py -v (not run: system Python lacks pytest)
- uv run --extra cpu --group dev python -m pytest tests/test_engine.py -v (attempted, but dependency environment setup stalled before pytest/torch were installed)


hiSandog:
Closing this stale PR because it has been open for more than 15 days.

PR #742, opened 2026-05-02
Fix smear patches for ckpt loading

art-test-stack:
Hey, 

Just noticed that smear (`smear_gate` and `smear_lambda`) and `backout_lambda` weights were not handled if loading model checkpoints that are from commits before commit a825e63

### DIFF
--- a/nanochat/checkpoint_manager.py
+++ b/nanochat/checkpoint_manager.py
@@ -27,25 +27,25 @@ def _patch_missing_config_keys(model_config_kwargs):
         model_config_kwargs["window_pattern"] = "L"
         log0(f"Patching missing window_pattern in model config to 'L'")
 
-def _patch_missing_keys(model_data, model_config):
+def _patch_missing_keys(model_data, model_config, device):
     """Add default values for new parameters that may be missing in old checkpoints."""
     n_layer = model_config.n_layer
     # resid_lambdas defaults to 1.0 (identity scaling)
     if "resid_lambdas" not in model_data:
-        model_data["resid_lambdas"] = torch.ones(n_layer)
+        model_data["resid_lambdas"] = torch.ones(n_layer).to(device)
         log0(f"Patching missing resid_lambdas in model data to 1.0")
     # x0_lambdas defaults to 0.0 (disabled)
     if "x0_lambdas" not in model_data:
-        model_data["x0_lambdas"] = torch.zeros(n_layer)
+        model_data["x0_lambdas"] = torch.zeros(n_layer).to(device)
         log0(f"Patching missing x0_lambdas in model data to 0.0")
-    if "smear_gate.weights" not in model_data:
-        model_data["smear_gate.weights"] = torch.ones(1, 24)
-        log0(f"Patching missing smear_gate.weights in model data to 1.0")
+    if "smear_gate.weight" not in model_data:
+        model_data["smear_gate.weight"] = torch.ones(1, 24).to(device)
+        log0(f"Patching missing smear_gate.weight in model data to 1.0")
     if "smear_lambda" not in model_data:
-        model_data["smear_lambda"] = torch.zeros(1)
+        model_data["smear_lambda"] = torch.zeros(1).to(device)
         log0(f"Patching missing smear_lambda in model data to 0.0")
     if "backout_lambda" not in model_data:
-        model_data["backout_lambda"] = 0.2 * torch.ones(1)
+        model_data["backout_lambda"] = 0.2 * torch.ones(1).to(device)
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
Hi, thanks for the PR!

You're right that the introduction of smear & backout made current `master` incompatible with older models.

However, there's a few issues with this PR. 

First, it should be `smear_gate.weight`, not `smear_gate.weights`. 

Further, and I think this was already an issue with the other code in `_patch_missing_keys`, the device isn't applied correctly. With `assign=True`, `model.load_state_dict` directly assigns tensors from `model_data` into the model rather than copying them into existing buffers. This means the device of the incoming tensors is preserved as-is. The patched tensors are created as plain CPU tensors, so they get assigned onto CPU even though the model is on CUDA. I've gone ahead and fixed this by passing through the `device` and moving them there before registering into `model_data`.

I'll push these edits now.

svlandeg (approved):
Ok, I think this now looks good. To recap, I tried sampling from an old model on file, from before smear & backout were introduced.

This would give
```
RuntimeError: Error(s) in loading state_dict for GPT:
        Missing key(s) in state_dict: "smear_lambda", "backout_lambda", "smear_gate.weight".
```

The original state of this PR then would error out with 
```
RuntimeError: Error(s) in loading state_dict for GPT:
        Missing key(s) in state_dict: "smear_gate.weight".
        Unexpected key(s) in state_dict: "smear_gate.weights".
```

and after fixing that typo, it would error with
```
  File "/home/ubuntu/git/nanochat/nanochat/gpt.py", line 50, in forward
    return F.linear(x, self.weight.to(dtype=x.dtype))
RuntimeError: Expected all tensors to be on the same device, but got mat2 is on cpu, different from other tensors on cuda:0 (when checking argument in method wrapper_CUDA_bmm)
```

with the `to(device)` fix added in this PR, it works fine, and the model's output looks reasonable as well:

> <|bos|>The capital of France is Paris, and the capital city of France is Paris Michel Fouquet Paris Michel Fou
>
> <|bos|>The chemical symbol of gold is Au. Gold is a soft, malleable, ductile, and lustrous
>
> <|bos|>If yesterday was Friday, then tomorrow will be Saturday. The weather is going to be sunny and warm. The sun will be

art-test-stack:
Hi, 
Thanks for the review (and sorry about the typo). And thanks for the explanation about the device allocation, I was not even aware of it. 
I was just concerned by one thing relative to this change (but it is not mentioned in the code base), but shouldn't we want to freeze the weights of parameters that were not included in the model dict? I mean, if we started training a model before commit #a825e63, and we resume the training from a commit after #a825e63, shouldn't we prefer not to train the model on the new weights? By setting `required_grads=False`.
That said, there is a similar issue with inference. The values of the weights that I put in the PR have been chosen from model initialization. But I wonder if we would not prefer weight values that would make those layers act as an identity function?
If you think the same, I can open a PR to commit these changes, or add them here

svlandeg:
Let's not create a new PR, the context is all here.

> But I wonder if we would not prefer weight values that would make those layers act as an identity function?

So smear lambda set to zero disables the smear entirely which is fine, right? Maybe `backout_lambda` should be zeros as well?

> if we started training a model before commit #a825e63, and we resume the training from a commit after #a825e63

Right, yea, though that's a bit of an edge case (I would assume people training & resuming to do some from the same branch/commit). Maybe Andrej can weigh in on this one (pun intended).

art-test-stack:
> So smear lambda set to zero disables the smear entirely which is fine, right? Maybe `backout_lambda` should be zeros as well?

I agree to both (moreover, `smear_gate` init is not a concern since `smear_lambda` = 0)

> Right, yea, though that's a bit of an edge case (I would assume people training & resuming to do some from the same branch/commit). Maybe Andrej can weigh in on this one (pun intended).

Mmh, that's right, but I was wondering if it was the role of the backward compatibility to make it "fully" functioning — because it would just introduce unexpected bias in the model (btw, regarding training, I just realized that it not only concerns resuming, but it also affects [sft](scripts/chat_sft.py) and [rl](scripts/chat_rl.py) pipelines)

karpathy:
i think won't fix, lmk if there is a good reason to

art-test-stack:
hi,

the good reason I see is that `checkpoint_manager` already has a bwd compatibility path for missing ckpt keys, but it currently only covers `resid_lambdas` and `x0_lambdas`. After smear/backout were added, older checkpoints now fail under `strict=True` because `smear_lambda`, `backout_lambda`, and `smear_gate.weight` are missing.

There is also a device issue because loading uses `assign=True`: patched tensors need to be created on the target device, otherwise CUDA loading can leave some assigned params on CPU.

Issue #459, opened 2026-01-23
Python capability in nanochat?


leolbyy:
Hi, recently I’ve been working on building my own version of **nanochat**, and I’m currently looking into the **Engine module**.

In [`nanochat/engine.py`](https://github.com/karpathy/nanochat/blob/master/nanochat/engine.py#L252-L267), the engine checks for special tokens to detect whether the model is inside a Python block, and then routes the execution to the calculator.

However, after looking through the training data and dataloader, I couldn’t find any examples that actually contain these Python-related special tokens.

So I’m a bit confused about the intended design here:

- Is Python / calculator usage currently trained / included in nanochat project?
- Or is there some preprocessing or data source that introduces these tokens which I might have missed?

svlandeg:
(Hi! I'll move this to a discussion thread - feel free to continue the conversation over there)

PR #744, opened 2026-05-05
get the code for evaluation of sample into a funtion to avoid duplicate code.

xrdddd:
1. since for the evaluation of sample, there are duplicate code in base_train and base_eval, so here to get the code be a funtion (keep the logic the same, for example, in train, still only prompt and with disable_fp8 while in eval both prompt and uncondition with no disable_fp8). 
2. the context funtion moved to base_eval since it seems for evaluation.

### DIFF
--- a/scripts/base_eval.py
+++ b/scripts/base_eval.py
@@ -31,6 +31,10 @@ import tempfile
 import argparse
 import torch
 
+from typing import Callable
+from contextlib import nullcontext, contextmanager
+from nanochat.gpt import Linear
+
 from nanochat.common import compute_init, compute_cleanup, print0, get_base_dir, autodetect_device_type, download_file_with_lock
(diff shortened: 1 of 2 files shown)


svlandeg (commented):
Avoiding duplicate code can make sense, but I'm a little confused as to why this PR still obtains the same number of line count (+89, -89) ?

xrdddd:
> Avoiding duplicate code can make sense, but I'm a little confused as to why this PR still obtains the same number of line count (+89, -89) ?

Thank you @svlandeg for your time to review. indeed, I never noticed this number thing, very interesting and coincidence. 
I calculated again, since the duplicated code is not so much and some new import statements and empty lines added, so I guess it's offset by that. (e.g. the lines reduction gain from old line range #459 - #473, around 14 lines reduced.)

Best,
xrdddd

PR #733, opened 2026-04-27
Run 7: d22 + MuonClip + warmdown=0.85 + final-lr=0 (88 min, 10.8% faster than Run 6)

giovannizinzi:
## Summary

New `runs/speedrun.sh` recipe that crosses GPT-2 CORE 0.256525 in **5294 s (88.2 min)** vs Run 6's 5934 s (99.0 min) — **10.8% faster**, single-shot CORE **0.2665** (vs Run 6's 0.2643), val_bpb 0.7242.

| | Run 6 (`a825e63`) | this PR |
|---|---|---|
| depth | 24 | **22** |
| training horizon | ratio=8 (≈ 5568 iter) | **ratio=11.05** (≈ 6000 iter) |
| total-batch-size | auto = 1 M | 1 M (explicit) |
| warmdown-ratio | 0.65 (default) | **0.85** |
| final-lr-frac | 0.05 (default) | **0.0** |
| muon-qk-clip-tau | — | **100** (new flag) |
| total_training_time | 5934 s | **5294 s** ← −10.8% |
| val_bpb | 0.7190 | 0.7242 |
| CORE (single-shot) | 0.2643 | **0.2665** |

CORE clears the 0.2626 reference (Run 6's value) by +0.004 single-shot. val_bpb sits 0.7% above Run 5's 0.71800 reference but is acceptable in the wider tolerance band; Run 6 itself is at 0.7190.

### Reproducibility note

CORE is single-shot-noisy. Three runs of the same submitted recipe land at 0.2665 / 0.2620 / 0.2617 (mean 0.263, σ ≈ 0.0028). The headline 0.2665 is a high draw; the median is 0.262, still above Run 6's 0.2643 single-shot. val_bpb is much tighter (0.7242 / 0.7240 / 0.7242 across the same three runs).

## What changed in the code

One new flag: `--muon-qk-clip-tau` for [Kimi K2 §A QK-Clip](https://arxiv.org/abs/2507.20534). Default 0 (off), bit-identical to current behavior. Total **+66 LOC across 3 files**:

- `nanochat/optim.py` (+44): `_apply_qk_clip()` helper, called after `MuonAdamW.step()` and `DistMuonAdamW.step()`.
- `nanochat/gpt.py` (+20): `setup_optimizer(muon_qk_clip_tau=0.0, …)` pulls `c_q`/`c_k` into a dedicated Muon group with `is_qk=True, qk_tau=tau` when tau > 0.
[...]

PR #747, opened 2026-05-07
87 minutes -> D22 Muon+Eq-row + bigram 

jn2clark:
# Muon+Eq-row Bigram D22 Speedrun Summary

## Proposed Baseline

This branch uses the same D22 direction as PR733 and adds three small changes:

- Muon+ / MuonEq-row optimizer tweaks
- a hashed bigram residual embedding
- `torch.compile` mode `max-autotune-no-cudagraphs`

The current submission defaults are set directly in `scripts/base_train.py` and
are used by `runs/speedrun.sh`.

## Proposed Results

Three runs:

| Run | train_time_min | train_final_val_bpb | train_core | eval_train_bpb | eval_val_bpb | eval_core |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | `87.01` | `0.722670` | `0.2615` | `0.721894` | `0.719980` | `0.2568` |
| 2 | `87.04` | `0.722645` | `0.2627` | `0.721876` | `0.719992` | `0.2609` |
| 3 | `86.88` | `0.722732` | `0.2566` | `0.721974` | `0.720107` | `0.2553` |

Average:

| Metric | Average |
|---|---:|
| train_time_min | `86.9767` |
| train_final_val_bpb | `0.722682` |
| train_core | `0.260267` |
| eval_train_bpb | `0.721915` |
| eval_val_bpb | `0.720026` |
| eval_core | `0.257667` |

I have had difficulty working out which metrics are being used for reporting evaluation since the bp appears in three places, at the end of training, as train bp and val bp in the eval output. I included all those metrics below and can provide the raw logs if needed. Because of this I re-ran PR733 to generate my own set of metrics. In general, its a difficult benchmark because the training horizon is chosen a priori to choose a target time but performance may end up not being good enough. Additional results are shown below comparing to the previous D24 candidate in Run6 demonstrating better valbp under the same time. 

## PR733 Rerun Results

Two reruns of PR733:

| Run | train_time_min | train_final_val_bpb | train_core | eval_train_bpb | eval_val_bpb | eval_core |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | `88.08` | `0.723062` | `0.2523` | `0.722318` | `0.720365` | `0.2473` |
| 2 | `88.06` | `0.722954` | `0.2557` | `0.722248` | `0.720268` | `0.2540` |

Average:

| Metric | Average |
|---|---:|
| train_time_min | `88.0700` |
| train_final_val_bpb | `0.723008` |
| train_core | `0.254000` |
| eval_train_bpb | `0.722283` |
| eval_val_bpb | `0.720317` |
| eval_core | `0.250650` |

## Average Delta vs PR733

| Metric | Proposed avg | PR733 avg | Delta | Read |
|---|---:|---:|---:|---|
| train_time_min | `86.9767` | `88.0700` | `-1.0933` | `1.24%` faster |
| train_final_val_bpb | `0.722682` | `0.723008` | `-0.000326` | better |
| train_core | `0.260267` | `0.254000` | `+0.006267` | better |
| eval_train_bpb | `0.721915` | `0.722283` | `-0.000368` | better |
| eval_val_bpb | `0.720026` | `0.720317` | `-0.000290` | better |
| eval_core | `0.257667` | `0.250650` | `+0.007017` | better |

The proposed branch is faster on all three runs and improves every reported BPB
metric and CORE metric on average.

### Compared to before init fix.

| Metric | this_work d24_5568 (time) | before-fix best (time) | this_work better by |
|---|---|---|---|
| Train final val BPB | 0.716189 (98.92 min) | 0.719319 (98.53 min) | 0.003130 |
| Eval train BPB | 0.715502 (98.92 min) | 0.718524 (98.53 min) | 0.003022 |
| Eval val BPB | 0.713584 (98.92 min) | 0.716681 (98.53 min) | 0.003097 |

### Compared to after init fix

| Metric | this_work d24_5568 (time) | after-fix best (time) | this_work better by |
|---|---|---|---|
| Train final val BPB | 0.716189 (98.92 min) | 0.718374 (99.14 min) | 0.002185 |
| Eval train BPB | 0.715502 (98.92 min) | 0.717520 (99.14 min) | 0.002018 |
| Eval val BPB | 0.713584 (98.92 min) | 0.715741 (99.14 min) | 0.002157 |

Both comparisons are at essentially the same training-time budget (~99 min), so the deltas reflect score gains rather than reduced compute cost.                                            

## Plots
Below are the plots of the bp metrics with time for the previous master baselines, PR733 and this work.
This PR sets a new loss frontier compared to the previous runs and compared to run 6 .

image

image

image

### Muon

Adds `--muon-plus`: one Frobenius renormalization after orthogonalization:

```python
g *= sqrt(min(m, n)) / ||g||_F
```

Reference: Muon+ paper, https://arxiv.org/abs/2602.21545

Adds `--muon-eq=row|col`: simple pre-orthogonalization row/column norm
equilibration.

Reference direction: MuonEq, https://arxiv.org/abs/2603.28254

The submitted recipe uses row equilibration because it worked best in screening.

### Bigram Residual

Adds two model/config knobs:

- `bigram_embed_factor`
- `bigram_lambda_init`, default `0.05`

When `bigram_embed_factor > 0`, the model creates a separate bigram embedding
table with `vocab_size * bigram_embed_factor` entries. For each token position,
the current token id and previous token id are hashed into that table. The
resulting embedding is added as a residual input before every transformer block:

```python
x = x + bigram_lambdas[i] * x0_bigram
```

The first token in each sequence uses a sentinel bucket because it has no
previous token. During KV-cache decoding, the previous token is stored in and
read from the cache so generation matches the training-time bigram definition.

The `GPTConfig` default for `bigram_embed_factor` remains `0`, but the
submission training default in `scripts/base_train.py` is `5`.

### Compile Mode

Short d16/40 throughput probes on the minimal branch:

| Compile mode | Avg logged tok/s, excluding first | Avg logged step time, excluding first | Total time |
|---|---:|---:|---:|
| default `torch.compile` | `324,995` | `1613.250ms` | `0.78m` |
| `max-autotune-no-cudagraphs` | `333,261` | `1573.250ms` | `0.76m` |

On this d16 probe, `max-autotune-no-cudagraphs` was about `2.5%` faster than
default `torch.compile` in steady state. The full new-vs-PR733 training-time
gain is smaller, about `1.24%`. 

## Metric Notes

There are several related metrics in the logs:

- `train_final_val_bpb`: the final `Validation bpb` printed during
  `scripts.base_train`
- `train_core`: the CORE metric printed during training
- `eval_train_bpb`: train split BPB from `scripts.base_eval`
- `eval_val_bpb`: val split BPB from `scripts.base_eval`
- `eval_core`: CORE metric from `scripts.base_eval`

The leaderboard primarily cares about wall-clock training time and CORE above
the GPT-2 threshold (`0.256525`). BPB is still useful because it is smoother and
less noisy than CORE.

The proposed branch has average `eval_core = 0.257667`, above the GPT-2
threshold. 

## Summary Claim

Compared to the PR733 reruns, the proposed branch is faster on all three runs
and improves train-time validation BPB, final eval train BPB, final eval val
BPB, and CORE on average. The final eval CORE average is above the GPT-2
threshold, while val BPB provides the smoother quality signal across repeats. It also beats
the previous best run 6 when run at D24 in terms of valbp metrics.


svlandeg (commented):
I currently don't have access to an 8xH100, and the proposed changes here are probably fine-tuned towards that architecture.

That said, on an A100 and the settings from the `speedrun.sh` here but without fp8 and for a d14, I get a slightly longer wall clock time (3h30 instead of 3h23) for a slightly worse CORE score (0.1507 instead of 0.1574).

For the leaderboard submission: just noting here that the current best result (run 6) obtains a `val_bpb` of 0.71800. I've gotten somewhat confused by the PR description here which first lists a lot of `val_bpb` numbers around 0.72 (in section "Proposed Results") and then different numbers in the section "Compared to Run 6". There's also a lot of detail in the comparison to #733, but ultimately we'd just need to rerun this on an 8xH100 to benchmark against run 6.

@jn2clark : what are the final proposed numbers for your submission, in terms of wall time, `val_bpb` and CORE score for the run as represented by the `speedrun.sh` changes in this PR?

As a side note, it looks like the autotuning is generating quite a bit of output on the console, so it might be nice to silence that somehow.

karpathy:
i don't want to merge the bigram embeddings. i know it works a bit better. im even suspicious of the smear gate already and might delete it. i merged the Muon suggestions, when i rerun the next nanochat speedrun i will credit in the table.

jn2clark:
apologies for the late reply, sounds good @svlandeg and @karpathy . thanks!

PR #758, opened 2026-05-26
Fix RL sampling when num_samples has a remainder

hiSandog:
## Summary
- generate RL rollouts in a final partial batch when num_samples is not divisible by device_batch_size
- allow the training pass loop to handle that final partial batch instead of requiring exact divisibility

## Validation
- python -m py_compile scripts/chat_rl.py


hiSandog:
Closing this stale PR because it is older than 15 days. I am leaving the branch intact.

PR #760, opened 2026-06-03
docs: improve skill description for better trigger accuracy

yanghaod2278827:
## Summary\n\n

Update the skill's \`description\` frontmatter to follow Claude Code skill development best practices:\n\n- Use **third-person** format ("This skill should be used when...")\n- Include **quoted trigger phrases** that match what users actually say\n- Make trigger conditions specific and concrete\n\n## Reference\n\nSee [Claude Code skill development guide](https://docs.anthropic.com/en/docs/claude-code/skills) for description best practices.\n\n---\nGenerated with Claude Code\n

### DIFF
--- a/.claude/skills/read-arxiv-paper/SKILL.md
+++ b/.claude/skills/read-arxiv-paper/SKILL.md
@@ -1,6 +1,9 @@
 ---
 name: read-arxiv-paper
-description: Use this skill when asked to read an arxiv paper given an arxiv URL
+description: >-
+  This skill should be used when the user provides an arXiv URL and asks to "read this paper",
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
I love how Claude Code is telling us how it wants it skills to be described 🙃 

Anyway, you give the reference to https://code.claude.com/docs/en/skills, but can you point me to where exactly it says to use third person and quoted phrases?

To me, it's a little counterintuitive to use quoted phrases, like an LLM should be perfectly capable of paraphrasing/interpreting an abstract context like "when asked to read..." so in a sense it feels weird to me to go through some lengths summing up different ways of saying exactly that.

Issue #590, opened 2026-03-05
[Bug] scripts/chat_sft.py produces loss: nan from step 00001 on small device-batch-size (≤8) due to fully-masked micro-batches


suraj-self:
### Description:
When running Supervised Fine-Tuning (SFT) on hardware with limited VRAM (e.g., 6GB/8GB GPUs), setting a small ```device-batch-size``` (e.g., 4 or 8) consistently leads to ```NaN``` loss values starting from Step 1. This appears to be a regression or unhandled edge case in how the SFT trainer handles micro-batches that contain no active training targets.

### Log
```
(nanochat) ss@ss-Predator-PH315-53:~/nanochat$ WANDB_RUN=nano_d12_sft_ss_v_stable_final \
PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True \
python -m scripts.chat_sft \
  --model-tag nanochat_v1_d12 \
  --model-step 2205 \
  --device-batch-size 8 \
  --matrix-lr 0.0002 \
  --warmup-ratio 0.05 \
  --run nano_d12_sft_ss_v_stable_final
Autodetected device type: cuda
2026-03-05 21:46:46,861 - nanochat.common - INFO - Distributed world size: 1
COMPUTE_DTYPE: torch.bfloat16 (auto-detected: CUDA SM 86 (bf16 supported))
2026-03-05 21:46:46,861 - nanochat.common - WARNING - Peak flops undefined for: NVIDIA GeForce RTX 3060 Laptop GPU, MFU will show as 0%
GPU: NVIDIA GeForce RTX 3060 Laptop GPU | Peak FLOPS (BF16): inf
wandb: Currently logged in as: suraj03143 (suraj-self) to https://api.wandb.ai. Use `wandb login --relogin` to force relogin
wandb: Tracking run with wandb version 0.21.3
wandb: Run data is saved locally in /home/ss/nanochat/wandb/run-20260305_214648-ehs5d2d2
wandb: Run `wandb offline` to turn off syncing.
wandb: Syncing run nano_d12_sft_ss_v_stable_final
wandb: ⭐️ View project at https://wandb.ai/suraj-self/nanochat-sft
wandb: 🚀 View run at https://wandb.ai/suraj-self/nanochat-sft/runs/ehs5d2d2
WARNING: Flash Attention 3 not available, using PyTorch SDPA fallback. Training will be less efficient.
[...]

PR #775, opened 2026-06-12
chat_sft: fix --num-iterations to count optimizer steps, not micro-batches

walbis:
## Bug

`--num-iterations` is documented as the number of **optimization steps**:

```python
parser.add_argument("--num-iterations", type=int, default=-1, help="number of optimization steps (-1 = full epoch)")
```

But the SFT dataloader counts **yielded micro-batches** and compares that count directly against `args.num_iterations`:

```python
it += 1
if 0 < args.num_iterations <= it and split == "train":
    last_step = True
...
approx_progress = it / args.num_iterations
```

The training loop calls `next(train_loader)` `grad_accum_steps` times per optimizer step, so whenever `grad_accum_steps > 1` two things go wrong:

1. **Training stops `grad_accum_steps`× too early** — `--num-iterations=N` yields N micro-batches, which is only ~N / grad_accum_steps optimizer steps.
2. **The LR multiplier goes negative and the loss goes NaN.** `approx_progress` advances `grad_accum_steps`× too fast and shoots past 1.0 during the *first* optimizer step. `get_lr_multiplier` doesn't clamp, so the warmdown branch extrapolates below zero: `lrm = 1 - (progress - 0.5) / 0.5` → e.g. **-127** at progress 64.5. One optimizer step at a large negative LR destroys the weights; `last_step` then fires and the run "completes", saving the ruined checkpoint.

### Observed

Single GPU, `total_batch_size=524288` inherited from the pretrained checkpoint, 4096-token micro-batches → `grad_accum_steps=128`, with `--num-iterations=2`. After the first optimizer step the generator has already yielded 129 micro-batches, so progress = 129/2 = 6450%:

```
step 00001 (6450.00%) | loss: nan | lrm: -127.00 | ...
```

### Why the official runs never hit it

- `runs/speedrun.sh` doesn't pass `--num-iterations` to chat_sft — it uses the epoch-driven stop (`-1`), where progress comes from dataset consumption instead of `it`.
- `runs/runcpu.sh` passes `--num-iterations=1500` but sets `total_batch_size = device_batch_size × max_seq_len` (16384 = 32 × 512), i.e. `grad_accum_steps == 1`, where micro-batches and optimizer steps coincide.

The natural way to hit it is any single-GPU run that inherits `total_batch_size=524288` from the checkpoint, uses a smaller device batch, and passes `--num-iterations` to shorten the run.

## Fix

Scale the generator-side stop/progress target by `grad_accum_steps`, so `--num-iterations` means optimizer steps, matching its help text and the meaning of an iteration in `base_train.py`. No behavior change when `grad_accum_steps == 1` (all official configs) or in the default `--num-iterations=-1` epoch mode.

(Not related to #590 — that NaN comes from fully-padded rows in the bestfit packer and reproduces regardless of `--num-iterations`.)

### DIFF
--- a/scripts/chat_sft.py
+++ b/scripts/chat_sft.py
@@ -267,16 +267,19 @@ def sft_data_generator_bos_bestfit(split, buffer_size=100):
             rows.append(row[:row_capacity])
             mask_rows.append(mask_row[:row_capacity])
 
-        # Stopping condition to respect num_iterations, if given
+        # Stopping condition to respect num_iterations, if given.
+        # num_iterations counts optimizer steps, but this generator yields micro-batches:
+        # grad_accum_steps of them per optimizer step.
         it += 1
-        if 0 < args.num_iterations <= it and split == "train":
+        target_micro_iters = args.num_iterations * grad_accum_steps
+        if 0 < target_micro_iters <= it and split == "train":
             last_step = True
 
         # Update progress tracking (based on consumed, not cursor, to account for buffering)
         if split == "train":
             current_epoch = epoch
             if args.num_iterations > 0:
-                approx_progress = it / args.num_iterations
+                approx_progress = it / target_micro_iters
             else:
                 approx_progress = consumed / dataset_size
             # Trigger last_step when we've consumed enough (instead of when cursor wraps)


walbis:
Independent reproduction on a completely different setup (Windows 11, single RTX 5070 Ti, torch 2.9.1+cu128) — identical symptom, posting in case a second data point on different hardware/OS is useful.

Prereqs: a couple of shards + tokenizer (`python -m nanochat.dataset -n 2` then `python -m scripts.tok_train`).

```bash
# tiny base model at a short context
python -m scripts.base_train --depth=4 --max-seq-len=512 --device-batch-size=8 \
  --num-iterations=20 --eval-every=-1 --core-metric-every=-1 --sample-every=-1

# SFT with a smaller device batch + an explicit --num-iterations
python -m scripts.chat_sft --model-tag=d4 --device-batch-size=4 --num-iterations=2 \
  --eval-every=-1 --chatcore-every=-1
```

The d4 checkpoint inherits `total_batch_size` from pretraining, so `device_batch_size=4 × max_seq_len=512` gives `grad_accum_steps=128`. First optimizer step:

```
step 00001 (6450.00%) | loss: nan | lrm: -127.00 | ...
```

— matching the PR description exactly. With this PR's `target_micro_iters` change, progress and LR are sane again (`lrm ≈ 0.99`, progress ≤ 100%) and the run trains instead of saving a wrecked checkpoint.

One note for anyone testing at this tiny `max_seq_len`: a residual `loss: nan` can still appear after this fix, but that one is the *separate* fully-padded-rows packing bug (#590 / #486) — it reproduces independently of `--num-iterations`, and capping `render_conversation` at `max_seq_len + 1` clears it. So the two are cleanly separable, exactly as the PR already states.

marcinbogdanski:
Hello

I can confirm this is a real issue and quite catastrophic on runs with `--num-iterations` and `grad_accum > 1`.

There are a couple of subtler related issues:

- off-by-one / off-by-two issues with stop condition and LR scaling
- slightly incorrect debiased loss reporting

I have put details and combined fix in [#816](https://github.com/karpathy/nanochat/pull/816)

marcinbogdanski:
Just clarifying the difference between this PR and the alternative in #816.

This PR addresses the issue where grad_accum > 1 causes the dataloader to incorrectly count iterations.

My PR #816 addresses two items:

- the grad_accum > 1 bug mentioned here
- off-by-one last_step and off-by-two LR schedule caused by batch pre-fetch

I'm addressing them with a slight refactor: moving progress accounting out of the dataloader into the train loop. I believe it is a simpler overall fix for all items. But if a minimal fix for grad_accum > 1 only is preferred, I think this PR is also correct as an isolated fix.

PR #779, opened 2026-06-18
Update

putta-saikiran:


svlandeg (commented):
This makes no sense - `.gitignore` shouldn't be gitignored 😁 

Closing.

PR #784, opened 2026-06-23
fix: make calculator timeout work on Windows and fix signal race condition

putta-saikiran:
fix: make calculator timeout work on Windows and fix signal race condition

signal.SIGALRM doesn't exist on Windows so the entire inference stack was crashing on startup with AttributeError. Also, signal.alarm(0) was sitting in the except block instead of finally, meaning a late-firing alarm could bleed into unrelated code on the same thread.

Replaced the Unix-only signal approach with a platform check — SIGALRM on Linux/macOS (now with proper finally + handler restore), and a stdlib ThreadPoolExecutor fallback on Windows. No new dependencies.

### DIFF
--- a/nanochat/engine.py
+++ b/nanochat/engine.py
@@ -13,6 +13,7 @@ The whole thing is made as efficient as possible.
 
 import torch
 import torch.nn.functional as F
+import platform
 import signal
 import warnings
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
Hi, thanks for the PR! Officially, Windows isn't a supported platform, but if it doesn't add much more complexity I think we could look into supporting it. 

Did you perform a full training & eval run on Windows? Did it run to completion?

svlandeg:
Sorry for not being clear, but I was asking you (not your AI) whether you actually ran the repo (eg a speedrun or cpu run) with one of the provided bash scripts, and whether you were successfully able to train a nanochat model in Windows with this branch.

PR #786, opened 2026-06-24
Match calculator tool output to training format (drop trailing .0)

devYRPauli:
## What

When the model calls the calculator tool, `Engine` injects the result back into the context between `<|output_start|>` / `<|output_end|>`. Those output tokens are unsupervised - the model only conditions its continuation on them - so their formatting needs to match what the model saw during training.

During SFT the calculator output comes from the GSM8K annotations (`tasks/gsm8k.py`), which write whole-number results as integers, e.g. `<<0.2*50=10>>` -> the output part is `"10"`.

At inference, `nanochat/engine.py` rendered the result with `str()`:

```python
result_tokens = self.tokenizer.encode(str(result))
```

`use_calculator("0.2*50")` returns the float `10.0`, and `str(10.0)` is `"10.0"`. So the model is conditioned on `"10"` during training but sees `"10.0"` at inference. This hits any expression that divides or multiplies to a whole number, which is very common in GSM8K.

## Fix

Render integer-valued floats without the trailing `.0` so the injected tool output matches the training annotations. Fractional results (e.g. `12/60 -> 0.2`) are unchanged.

```python
def format_calculator_result(result):
    if isinstance(result, float) and result.is_integer():
        return str(int(result))
    return str(result)
```

## Test

Added `test_calculator_result_matches_training_format` to `tests/test_engine.py`, covering the canonical `0.2*50 -> "10"` case, a division case, and unchanged fractional/integer results.

### DIFF
--- a/nanochat/engine.py
+++ b/nanochat/engine.py
@@ -78,6 +78,19 @@ def use_calculator(expr):
     # Evaluate with timeout
     return eval_with_timeout(expr)
 
+def format_calculator_result(result):
+    """
+    Render a calculator result the same way it appears in training data.
+    GSM8K writes whole-number results as integers (e.g. "10", not "10.0"), but
+    eval returns a float for division and many products (0.2*50 -> 10.0). Without
+    this, the engine would inject "10.0" at inference while the model was trained
+    on "10", a train/inference mismatch on the tool-output tokens. Fractional
+    results are left unchanged.
+    """
+    if isinstance(result, float) and result.is_integer():
+        return str(int(result))
+    return str(result)
+
 # -----------------------------------------------------------------------------
 class KVCache:
     """
@@ -263,7 +276,7 @@ class Engine:
                         expr = self.tokenizer.decode(state.python_expr_tokens)
                         result = use_calculator(expr)
                         if result is not None:
-                            result_tokens = self.tokenizer.encode(str(result))
+                            result_tokens = self.tokenizer.encode(format_calculator_result(result))
                             state.forced_tokens.append(output_start)
                             state.forced_tokens.extend(result_tokens)
                             state.forced_tokens.append(output_end)
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -5,7 +5,7 @@ python -m pytest tests/test_engine.py -v
 """
 
 import torch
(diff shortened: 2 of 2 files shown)


svlandeg (commented):
Thanks for the PR!

Have you found this makes a difference in terms of evaluation score?

devYRPauli:
Good question - I have not measured an eval-score delta, so I do not want to claim one. I framed this as a train/inference consistency fix rather than a tuning change.

The mismatch is on the tool-output tokens the model conditions on. In training (tokenizer render_conversation, the python_output part), the output text is taken verbatim from the GSM8K annotation - for <<0.2*50=10>> that part is "10", injected as unsupervised context. At inference, engine.py injected str(use_calculator(expr)); use_calculator evals the expression, so 0.2*50 -> 10.0 and str() -> "10.0". So the model is trained to condition on "10" but sees "10.0" at test time.

It only bites when the result is a whole-number float: division (100/10 -> 10.0) or a decimal product (0.2*50 -> 10.0). Pure int expressions (5*2 -> 10) were already fine, and genuinely fractional results (12/60 -> 0.2) are unchanged by the fix. Those whole-number cases are common in GSM8K.

Whether that moves the final score is empirical and I have not run it - I do not have a trained checkpoint and eval set up on my end. If you would like a before/after on GSM8K I am happy to run it if you can point me at the expected harness/checkpoint.

devYRPauli:
Following up with something more concrete than my earlier answer, though it is still not an eval score.

I ran the GSM8K train split (7,473 examples) through nanochat's own annotation parser (`tasks/gsm8k.py`) and `use_calculator` (`nanochat/engine.py`), comparing what inference emits against the annotation text used in training. Of 23,716 calculator calls, 6,348 (26.8%) render differently before this change (`str(result)` making `10` into `10.0`); with it, that drops to 689 (2.9%). 57% of train examples contain at least one affected call. The residual 2.9% is a separate dataset quirk (annotations like `144.00` or `.4`) that this PR does not target.

To be clear on what this is: a static match rate against training annotations, not a measured accuracy delta. I still have no trained checkpoint to run `chat_eval`, so I cannot say whether GSM8K accuracy moves. It only shows the mismatch is not a rare edge case. Happy to run the real eval if there is a checkpoint I can use.

PR #793, opened 2026-06-27
docs: clarify bf16 works on MPS (Apple Silicon) for ~25% memory savings

eaglstun:
The precision docs and the _detect_compute_dtype comment implied MPS is an fp32-only path ("no reduced-precision tensor cores"). In practice recent macOS runs bfloat16 on MPS fine: NANOCHAT_DTYPE=bfloat16 lowers peak memory ~25% in both training and inference with no loss of training stability (measured ~26%: 1412MB -> 1041MB for a 125.8M-param model on M4 Max / macOS 26.4.1).

Split the CPU and MPS rows in the README precision table, document the env-var opt-in and its memory-vs-throughput tradeoff (no FA3/bf16 fast path on MPS, so the win is memory headroom, not speed), and note it's left opt-in because bf16 on MPS was unreliable on older macOS. No behavior change.

Refs karpathy/nanochat#685

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,8 @@ nanochat does not use `torch.amp.autocast`. Instead, precision is managed explic
 |----------|--------------|-----|
 | CUDA SM 80+ (A100, H100, ...) | `bfloat16` | Native bf16 tensor cores |
 | CUDA SM < 80 (V100, T4, ...) | `float32` | No bf16; fp16 available via `NANOCHAT_DTYPE=float16` (uses GradScaler) |
-| CPU / MPS | `float32` | No reduced-precision tensor cores |
+| CPU | `float32` | No reduced-precision tensor cores |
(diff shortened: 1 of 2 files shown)


mparrett:
Hi there, stopping by to mention this pairs well with #741, the optimizer crash fix that makes bf16 on MPS run successfully end to end. On the quality side of your "no loss of training stability" point: I measured bf16 vs fp32 at 5000 steps on d6: val_bpb **1.169138 vs 1.168621**, +0.0005 (0.04%), within per-eval sampling noise. So the parity holds at horizon scale, not just the memory win.

svlandeg (approved):
Thanks @eaglstun, this looks good to me! 🙏 
I would only merge this after merging #741 though.

karpathy:
Thanks for measuring this! Landed a condensed version in 5e43f1e (a README table note + a one-line comment in common.py) — kept it short since device/OS-specific measurements tend to age quickly in docs. Closing, thank you!

PR #795, opened 2026-06-30
fix(engine): allocate inference KV cache at COMPUTE_DTYPE (fixes bf16 on MPS)

mparrett:
## What

`Engine.generate()` allocates the inference KV cache with `bfloat16 if device.type == "cuda" else float32`, which ignores `NANOCHAT_DTYPE`. On a non-CUDA device with `NANOCHAT_DTYPE=bfloat16` (the supported MPS bf16 path — see #741, #685), the model forward emits bf16 `q` against an fp32 cache, and `F.scaled_dot_product_attention` raises on MPS:

```
RuntimeError: Expected query, key, and value to have the same dtype,
but got query.dtype: c10::BFloat16 key.dtype: float and value.dtype: float instead.
```

CUDA promotes mixed dtypes silently, so this never surfaced there.

## Fix

Allocate the cache at `COMPUTE_DTYPE` — the repo-wide compute dtype that already tracks `NANOCHAT_DTYPE` — so it matches the model's compute dtype. `generate_batch()` delegates to `generate()`, so this one site covers both paths.

This is the `COMPUTE_DTYPE` approach @missflash proposed in #695, narrowed to the single change that fixes this — no `chat_web.py` refactor.

## Behavior

| scenario | before | after |
|---|---|---|
| CUDA Ampere+, no override | bf16 | bf16 (unchanged) |
| CPU / MPS, no override | fp32 | fp32 (unchanged) |
| MPS + `NANOCHAT_DTYPE=bfloat16` | fp32 → crash | bf16 |
| CUDA pre-Ampere, no override | bf16 vs fp32 forward | fp32 |
| CUDA + `NANOCHAT_DTYPE=float32` | bf16 vs fp32 forward | fp32 |

Mainstream CUDA is unchanged. The last two rows were latent mismatches the old heuristic got away with only because CUDA promotes silently; aligning to `COMPUTE_DTYPE` corrects them and halves the cache footprint at bf16.

## Validation

On an M2 (MPS, torch 2.11), `NANOCHAT_DTYPE=bfloat16`, running `Engine.generate_batch`:

- with this change: generation completes
- reverting just this line to the old heuristic: the same run raises the SDPA dtype error above at the first decode step

The fix is weight-independent, so this was exercised with a small model on the current arch rather than a specific checkpoint.

## Related

- #741 — the `optim.py` MPS bf16 fix; this is the inference-side companion
- #685 — MPS bf16 support request
- #695 — prior art for the `COMPUTE_DTYPE` approach

### DIFF
--- a/nanochat/engine.py
+++ b/nanochat/engine.py
@@ -17,7 +17,7 @@ import signal
 import warnings
 from contextlib import contextmanager
 from collections import deque
-from nanochat.common import compute_init, autodetect_device_type
+from nanochat.common import compute_init, autodetect_device_type, COMPUTE_DTYPE
 from nanochat.checkpoint_manager import load_model
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
Makes sense, thanks! I suggest merging this together with #741.

karpathy:
Thanks — this was the right fix. Landed in ca366eb with the comment trimmed down to keep things lean. Closing, thank you!

PR #798, opened 2026-07-03
Add optional Mixture of Experts (MoE) addon, distilled from the moe2 …

karpathy:
…branch

DeepSeekV3-style MoE as a drop-in replacement for the MLP: sigmoid top-k router, one always-on shared expert, aux-loss-free load balancing via bias nudging, grouped-GEMM expert execution, 3D expert params trained with Muon. Off by default (--num-experts=0 keeps the exact dense model); enable with e.g. --num-experts=8 --top-k=1. Experts are sized iso-FLOP with the dense MLP so quality comparisons are compute-fair.

Defaults are the winners of the d16 sweep (see dev/LOG.md on moe2): router LR 0.005, balance coeff 1e-3, k+1 iso-FLOP sizing. Verdict from that investigation: at nanochat's ratio-8 speedrun horizon MoE improves bpb-per-FLOP but loses on wall clock (the ~1.3-1.6x memory-traffic tax of sparse experts exceeds the ~1.1x compute multiplier), hence an optional addon branch rather than a default.

### DIFF
--- a/nanochat/gpt.py
+++ b/nanochat/gpt.py
@@ -21,6 +21,7 @@ import torch.nn.functional as F
 
 from nanochat.common import get_dist_info, print0, COMPUTE_DTYPE
 from nanochat.optim import MuonAdamW, DistMuonAdamW
+from nanochat.moe import MoE, update_expert_biases, compute_moe_stats
 
 # Our custom Flash Attention module that automatically uses FA3 on Hopper+ and SDPA fallback elsewhere
 from nanochat.flash_attention import flash_attn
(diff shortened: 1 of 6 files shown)


karpathy:
Won't merge, but I did some work on MoE again. I still can't make it better than the baseline dense model. Possibly the nanochat models are just too small and the overhead is too much. This version (moe3) is better than the last (moe), and has minor fixes and improvements.

bharatsachya:
## 🔦 ByteBell — PR review

**Add Mixture of Experts (MoE) support** · _feature_

Add Mixture of Experts (MoE) support across the entire repository, making the MoE layer a drop-in replacement for the dense MLP in each transformer block, with dedicated optimizer handling for router parameters, active-parameter accounting, and per-iteration balancing — while preserving the dense-only training path when --num-experts=0.

✅ **No exceptions** — the change looks clean against the indexed base.

---
_Reviewed `moe3` against `92d63d4e8bb4df75c3b71618f31ddde2378b2bcd` · 6 files · posted from ByteBell._

PR #800, opened 2026-07-04
[WIP] The experiment refactor: nanochat as a miniseries research harness

karpathy:
**Work in progress, changing rapidly. Not ready for review; the code on this branch moves around daily.**

## Motivation

nanochat is repositioning from "train your own ChatGPT clone for $100" to **the reference pretraining harness for small LLMs**: the simplest solid baseline that a researcher with a compute allocation forks. Small, boring dependencies, obsessively verified, a reference people fork rather than a framework people configure. Code is cheap now (anyone can ask an LLM for a feature), so the repo only has to contain the hard essence, and it should fit comfortably in one context window.

**The product is the miniseries curve.** `--depth` is the single dial; sweeping it traces a ladder of compute-optimal models, and `curve.log` is the deliverable: quality, speed and cost of every model in one file, comparable across experiments by compute multiplier. Chat is the payoff at the end of the pipeline, not a second research surface. Speedrunning one model is a different thing from sweeping a ladder, so the GPT-2 leaderboard is retired.

## What the branch does, at a high level

- An **experiment** is `(name, git commit, depth ladder, dataset)`. `run.sh <name>` runs one end to end; every stage is idempotent and a running ladder can be steered by touching files. There is no config file: the code is the config.
- **The logs are the source of truth.** Stages print `<tag> key=value` records to stdout; `harness/experiment.py` is the one experiment CLI (`init`, `curve`, `compare`).
- **All data work happens once, up front**: raw shards, tokenizer and packed token shards for pretraining; the SFT conversations packed into the same format.
- **A thin SFT** and a **hard prune**: RL, tool use, the sandbox, the generative evals, fp16, the report generator and the leaderboard are gone. `scripts/` is the grid `{base,chat}_{prepare,train,eval,inference}`.
- **A layout a reader can hold**: `nanochat/` is what a model is and how it trains and never imports the rest; `harness/` is the experiment around it; `evals/` measures.

`experiment_refactor.md` in the repo root has the motivation and the ship list.

## Breaking changes

Old checkpoints do not load, the on-disk layout changed (a runtime notice explains the one `mv` for an existing dataset download), scripts were deleted. `master` will be tagged before this merges so the old "$100 chatbot" stays pinnable.

Per the repo AI policy (disclosure): this refactor is being implemented in a pair-programming session with Claude; every change is verified as it lands.

https://claude.ai/code/session_016EftvGTc742B3JW3mr3SEA


svlandeg (commented):
Just a quick drive-by comment: I think it'd be nice to be able to share the tokenizer more easily across experiments. Currently whenever I start a new run, I'm just copying the data & tokenizer before starting the actual training, so it's all replicated X times on disk. Somehow collecting tokenizer+ID mappings on file and being able to refer to such a tokenizer by ID from within an experiment would be nice.

Also, could the log record the verbatim script calls, like the exact arguments given to `base_train` etc? When testing runs for PRs with new settings, these aren't always present in the logs and it would be nice to see the exact arguments that were given, for reproducibility's sake.

karpathy:
First full-ladder overnight run on the new harness: `DEPTHS="12 16 20 24" bash runs/run.sh run1`. Started 03:47, finished 10:00 (~6h15m wall on 8xH100, ~4.5h of it base training). All 16 stage cells (4 depths × 4 stages) completed and aggregated into `curve.log`:

| | d12 | d16 | d20 | d24 |
|---|---|---|---|---|
| params | 286M | 537M | 897M | 1.38B |
| tokens trained | 1.3B | 2.8B | 5.2B | 8.8B |
| pretraining FLOPs | 1.0e18 | 4.5e18 | 1.5e19 | 4.2e19 |
| base val_bpb | 0.845 | 0.781 | 0.737 | 0.703 |
| base CORE | 0.158 | 0.200 | 0.246 | 0.283 |
| chat ChatCORE | 0.083 | 0.151 | 0.198 | 0.255 |
| inference tok/s (bs=1) | 166 | 126 | 103 | 87 |

Health notes:

- Monotone in depth across every metric (val_bpb, CORE, ChatCORE, and each individual task), throughput degrades smoothly with model size.
- d12 reproduced yesterday's shakedown run bit-identically (val_bpb 0.845361 to 6 decimals) — the pipeline is deterministic across runs.
- d20 CORE = 0.246, right at the usual ~0.25 "GPT-2 grade" line, i.e. no silent training regression from the refactor.
- Provenance machinery worked: the experiment's `meta.json` recorded the dirty tree and the harness snapshotted the local edit (the DEPTHS change) into `code_diff.patch`.
- The "did it finish, is it healthy" check is now just `grep -q "^summary "` per stage log plus a glance at `curve.log` — which was the point of the log grammar.

karpathy:
@svlandeg good suggestions, thank you!

**Verbatim invocations**: agree, done in 82cdadb. Every stage script now prints two records at startup:

```
argv script=scripts.base_train args='--depth=12 --fp8'
config depth=12 fp8=True device_batch_size=-1 ...
```

- `argv` = copy-paste reproduction of the run
- `config` = fully resolved args *including defaults* (which argv doesn't show, and which drift as the code evolves — exactly the PR-testing case you mention)

**Tokenizer sharing**:

- the data half of this pain should already be fixed in this PR: datasets now live once in a shared store (`~/.cache/nanochat/datasets/<name>/`) and experiments reference them by name — no more copying
- the tokenizer is still per-experiment, but it's small (~few MB) and ~1 min to train, and since training is deterministic, two experiments on the same dataset get identical tokenizers anyway — so the per-experiment copy doubles as cheap provenance
- escape hatch that works today: `ln -s ../other_exp/tokenizer $NANOCHAT_BASE_DIR/experiments/new_exp/tokenizer` — run.sh sees the tokenizer.pkl and skips training
- a proper named-tokenizer store (mirroring the datasets pattern, e.g. `NANOCHAT_TOKENIZER=...`) is on the open questions list in experiment_refactor.md; leaning to keep it out of this already-large PR and add it if the symlink proves too clunky in practice

karpathy:
Two updates since the run above (26d912b, fa7f2e4):

- `base_train` now reports total pretraining compute in its `summary` record as `eflops` (1 EFLOP = 1e18 FLOPs), and it's a column in the curve table — compute is the natural x-axis of the cost-perf curve, and previously it only appeared as prose.
- The default depth ladder is now `12 14 16 20 24 28`, chosen to span exactly two orders of magnitude of pretraining compute, 1e18 → 1e20 FLOPs, denser at the cheap end:

| depth | dim | scaling params | tokens (ratio 12) | total FLOPs | rung ratio |
|---|---|---|---|---|---|
| 12 | 768 | 110M | 1.32B | 1.00e18 | — |
| 14 | 896 | 164M | 1.97B | 2.22e18 | 2.2× |
| 16 | 1024 | 235M | 2.82B | 4.47e18 | 2.0× |
| 20 | 1280 | 435M | 5.22B | 1.51e19 | 3.4× |
| 24 | 1536 | 730M | 8.76B | 4.18e19 | 2.8× |
| 28 | 1792 | 1.14B | 13.7B | 1.01e20 | 2.4× |

Six rungs fit a scaling trend better than four, the round endpoints make the x-axis self-explanatory, and the bottom three rungs (`12 14 16`, ~8 EFLOPs, under an hour end to end) form a self-contained mini-ladder for quick "does my idea scale" checks before spending on the big rungs. The full ladder is ~13h on 8xH100; drop the 28 (`DEPTHS="12 14 16 20 24"`) for an overnight-sized (~6h) run.

nanochat/checkpoint_manager.py
```diff
@@ -19,24 +19,16 @@ def log0(message):
...
-    if "x0_lambdas" not in model_data:
-        model_data["x0_lambdas"] = torch.zeros(n_layer)
-        log0(f"Patching missing x0_lambdas in model data to 0.0")
+def get_checkpoint_dir(model_tag, source):
+    """
+    Checkpoints belong to the experiment: experiments/<name>/<model_tag>/<source>/
+    e.g. experiments/my_exp/d12/base/. The model_tag defaults to d<depth> in the
+    training scripts but can be overridden (the --model-tag escape hatch).
+    """
+    assert source in ["base", "mid", "sft", "rl"], f"Invalid source: {source}"
```
svlandeg, line 28: `mid` was nuked, right?
```suggestion
    assert source in ["base", "sft", "rl"], f"Invalid source: {source}"
```

experiment_refactor.md
```diff
@@ -0,0 +1,286 @@
+# The Experiment Refactor
```
svlandeg, line 1: Maybe let's create a separate folder to keep these kind of logs in? (something like `/design/` ?)

nanochat/dataset.py
```diff
@@ -1,62 +1,99 @@
...
 
 # -----------------------------------------------------------------------------
-# The specifics of the current pretraining dataset
+# The canonical dataset: hosted, downloaded on demand
 
-# The URL on the internet where the data is hosted and downloaded from on demand
+CANONICAL_DATASET = "climbmix"
 BASE_URL = "https://huggingface.co/datasets/karpathy/climbmix-400b-shuffle/resolve/main"
-MAX_SHARD = 6542 # the last datashard is shard_06542.parquet
+MAX_SHARD = 6542 # the last datashard is shard_06542.parquet, and it is the val split
```
svlandeg, line 40: This `MAX_SHARD` is still hardcoded to be relevant only for the climbmix dataset?

nanochat/experiment.py
```diff
@@ -0,0 +1,162 @@
...
+        "name": name,
+        "created": datetime.now().isoformat(timespec="seconds"),
+        "git_commit": git("rev-parse", "HEAD"),
+        "git_branch": git("rev-parse", "--abbrev-ref", "HEAD"),
+        "git_dirty": bool(dirty_diff),
+        "dataset": get_dataset_name(),
+    }
+    with open(meta_path, "w") as f:
+        json.dump(meta, f, indent=2)
+    # a dirty tree would make the commit hash a lie, so record the diff for reproducibility
```
svlandeg, line 63: This is really nice 🌟 

I was testing this though and found a lot of noise recorded here from the `uv.lock`, which is updated by  the `uv sync` statement in `run.sh`. It then records noise such as:

```
-    { name = "nvidia-cublas-cu12", marker = "(sys_platform == 'linux' and extra != 'extra-8-nanochat-cpu') or (extra == 'extra-8-nanochat-cpu' and extra == 'extra-8-nanochat-gpu')" },
-    { name = "nvidia-cusparse-cu12", marker = "(sys_platform == 'linux' and extra != 'extra-8-nanochat-cpu') or (extra == 'extra-8-nanochat-cpu' and extra == 'extra-8-nanochat-gpu')" },
-    { name = "nvidia-nvjitlink-cu12", marker = "(sys_platform == 'linux' and extra != 'extra-8-nanochat-cpu') or (extra == 'extra-8-nanochat-cpu' and extra == 'extra-8-nanochat-gpu')" },
+    { name = "nvidia-cublas-cu12", marker = "(sys_platform == 'linux' and extra == 'extra-8-nanochat-gpu') or (extra == 'extra-8-nanochat-cpu' and extra == 'extra-8-nanochat-gpu')" },
+    { name = "nvidia-cusparse-cu12", marker = "(sys_platform == 'linux' and extra == 'extra-8-nanochat-gpu') or (extra == 'extra-8-nanochat-cpu' and extra == 'extra-8-nanochat-gpu')" },
+    { name = "nvidia-nvjitlink-cu12", marker = "(sys_platform == 'linux' and extra == 'extra-8-nanochat-gpu') or (extra == 'extra-8-nanochat-cpu' and extra == 'extra-8-nanochat-gpu')" },
```
It really obfuscates other code changes in the `code_diff.patch` file. In this particular case, I think it just happens because there's a slightly different version of `uv` running on my ubuntu machine.

I wonder what's best here. We could run with `uv sync --frozen` instead. If you want to change a dep, that should happen separatly in `pyproject.toml` and maybe with a deliberate `uv sync`, rather than it being a side product of an experiment run?

nanochat/experiment.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def read_base_summary(experiment_dir, model_tag):
+    """The final summary of one model's pretraining, or None if incomplete."""
+    log_path = os.path.join(experiment_dir, model_tag, "base_train.log")
+    summary = read_stage_summary(log_path)
+    if summary is None or "val_bpb" not in summary:
+        return None
+    if "eflops" not in summary:
+        # legacy fallback: logs written before the eflops summary field (Jul 2026)
```
svlandeg: Is this legacy fallback really needed? Will this new code ever run on old log files? 🤔

nanochat/logfmt.py
```diff
@@ -0,0 +1,94 @@
...
+def _parse_value(value_str):
+    """Infer the type of a value: int, then float, else string."""
+    try:
+        return int(value_str)
+    except ValueError:
+        pass
+    try:
+        return float(value_str)
+    except ValueError:
+        pass
```
svlandeg, line 58: Slightly less verbose:

```suggestion
    for cast in (int, float):
        try:
            return cast(value_str)
        except ValueError:
            pass
```

svlandeg (commented):
A few more quick comments 👇

nanochat/dataset.py
svlandeg: I guess it doesn't really matter as this is only used in the context of downloading Climbmix as default. I think to clarify that, I would personally move the definition of this var to within the `"__main__"` part.

svlandeg:
I wanted to check the migration message with the new PR on an old home folder and it works fine. Just some spurious thoughts here:
- there's other dirs that could/should be removed together with the checkpoints: `report`, `base_eval`. Not sure if this should be mentioned or not.
- maybe mention that `identity_conversations.jsonl` & `words_alpha.txt` aren't supported anymore either and can be deleted too (right?)
- `NUM_SHARDS` is now 1000 in `run.sh`. The previous default in `speedrun.sh` was 170, so this will surprise users that after the data migration, the script will still go ahead and download 830 parquet files.

tonydzi:
The choice to make completion a **record in the log** rather than a process exit status is the part of this design I'd defend hardest, and it's worth saying why out loud: an exit code is overloaded. In Python, exit 1 means both "ran fine and reported something" and "died on an uncaught exception," so any harness that reads meaning into it will eventually paint a crash as a normal outcome. `grep -q "^summary "` can't make that mistake. Same for `config` printing fully-resolved defaults — the drift @svlandeg raised is real and invisible in `argv` alone.

Two gaps we hit running the same shape unattended (different domain — long-running agent/ops pipelines, not training — so read these as failure shapes, not as claims about your code, which I haven't traced).

**1. The marker proves a line was printed, not that the artifact exists.**

`summary` is stdout; the artifact is a file. If a stage dies between the two, re-entry skips the cell and the next stage reads something absent or truncated. We got bitten by the compound version of this today: a scheduled job died on a `TypeError` before writing its status file, its sibling caught the exception, printed "couldn't write status," and continued — and a freshness check then read a file of *that same name*, written by a different producer, as proof of life. Every individual log line was true. The false conclusion was assembled by the reader.

The fix is one field inside the grammar you already have: let `summary` name what it produced.

```
summary ... out=d12/base/model.pt bytes=<n> sha256=<digest>
```

…and have `run.sh`'s skip check stat that path, not just match the line. Completion stops being an assertion and starts carrying its own evidence. The lesson we took: check the **writer** of an artifact, not only that an artifact with the right name is there.

**2. Idempotent re-entry silently mixes code versions.**

`meta.json` records commit + dirty diff for the *experiment*; skipping is decided per stage cell. Re-enter a half-finished ladder after editing the tree and the completed cells were built by different code — then `curve.log` has one `model` row per model with nothing recording which code each row came from. A month later that table reads as one experiment.

This matters most for the use you designed the bottom rungs for: `12 14 16` as the "does my idea scale" check is *by construction* edit-then-rerun. Since `argv`/`config` already exist, `code=<commit>[+<diff-sha>]` on `summary` makes each row self-describing and lets `run.sh` decide — re-run or warn — instead of the human remembering what they changed at 3am.

Both are small and additive; happy to put up a PR for either if it fits, or leave it as a note on the design doc's open questions if you'd rather keep this branch narrow.

*Disclosure, matching the convention in the PR body: written by an AI agent (Claude) contributing as Palo Alto AI Research Lab. The incident above is ours and happened today; I've kept out any number I haven't measured.*

PR #803, opened 2026-07-10
Fix SFT packing NaNs on short contexts

armoredmeatball:
## Summary

- make SFT best-fit packing truncate the shortest oversized conversation when an empty row cannot otherwise make progress
- skip packed batches that have no supervised target tokens after the causal shift
- add focused regression coverage and disable ChatCORE evaluation in the CPU reference run

## Why

With short sequence lengths, every buffered conversation can be too large for an empty row. The previous packer left those conversations in the buffer and emitted fully padded rows. Since all targets were ignored, mean cross-entropy evaluated as `0 / 0` and produced NaN.

The fallback now removes and truncates one oversized conversation only when starting an empty row, preserving normal best-fit behavior while guaranteeing progress. An additional guard prevents any all-masked batch from reaching the loss.

## Impact

SFT runs with short contexts no longer stall on oversized conversations or train on NaN-producing batches. Conversations that fit continue to be packed without truncation.

## Validation

- `uv run --frozen --extra cpu --group dev python -m pytest -q tests/test_sft_packing.py` (4 passed)
- `uv run --frozen --extra cpu python -m py_compile nanochat/dataloader.py scripts/chat_sft.py`
- `bash -n runs/runcpu.sh`
- `git diff --check upstream/master...HEAD`

### DIFF
--- a/nanochat/dataloader.py
+++ b/nanochat/dataloader.py
@@ -22,6 +22,37 @@ import pyarrow.parquet as pq
 from nanochat.common import get_dist_info
 from nanochat.dataset import list_parquet_files
 
+
+def pop_best_fit_conversation(conversation_buffer, max_length, truncate_if_needed=False):
+    """Pop the largest conversation that fits, optionally truncating as a fallback."""
+    assert max_length > 0
+    if not conversation_buffer:
+        raise ValueError("Cannot select a conversation from an empty buffer")
+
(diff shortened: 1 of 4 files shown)


armoredmeatball:
Closing because this PR was opened against the upstream repository by mistake; the branch will be reviewed in the fork instead.

PR #826, opened 2026-08-09
Fix SDPA fallback: sliding window doesn't reduce memory (fixes #825)

### LINKED ISSUE #825: SDPA fallback: sliding window doesn't reduce memory (full mask built regardless of window size)
In `_sdpa_attention` (nanochat/flash_attention.py), the sliding-window
branch (lines 99-110) builds a full `(Tq, Tk)` boolean mask before
applying the window restriction:

    row_idx = (Tk - Tq) + torch.arange(Tq, device=device).unsqueeze(1)
    col_idx = torch.arange(Tk, device=device).unsqueeze(0)
    mask = col_idx <= row_idx

    if window >= 0 and window < Tk:
        mask = mask & ((row_idx - col_idx) <= window)

    return F.scaled_dot_product_attention(q, k, v, attn_mask=mask, enable_gqa=enable_gqa)

The mask tensor is always shape `(Tq, Tk)`, regardless of `window`. The
window only narrows which entries inside that full-size mask are `True`
after it's built. Passing an arbitrary `attn_mask=` tensor also forces
SDPA onto its unfused math backend, which materializes the full
`Tq x Tk` score matrix before masking.

I measured this directly on a Tesla T4 (no FA2/FA3 available, so this
is the actual path used). At T=2048, batch=1, 6 query heads:

    window=16    -> peak memory ~296 MB
    window=64    -> peak memory ~301 MB
    window=256   -> peak memory ~300 MB
    window=1024  -> peak memory ~300 MB
    unlimited    -> peak memory ~305 MB

Memory is effectively flat regardless of window size - a window=16
layer costs about the same as unlimited attention on this fallback path.

I tried an alternative that slices k/v per query block before calling
SDPA, so SDPA never sees more than the window actually needs:

    for start in range(0, Tq, block_size):
        end = min(start + block_size, Tq)
        q_block = q[:, :, start:end, :]
        k_start = max(0, (offset + start) - window)
        k_end = offset + end
        k_block = k[:, :, k_start:k_end, :]
        v_block = v[:, :, k_start:k_end, :]
        # build a small mask sized to the block, not (Tq, Tk)
        ...
        out_block = F.scaled_dot_product_attention(q_block, k_block, v_block, attn_mask=mask, enable_gqa=enable_gqa)
        outputs.append(out_block)
    return torch.cat(outputs, dim=2)

Same T=2048/batch=1 shape, output verified identical to the original
(max abs diff 0.0 across several window sizes):

    window=16    -> ~41 MB   (296 -> 41)
    window=64    -> ~43 MB
    window=256   -> ~47 MB
    window=1024  -> ~64 MB

Memory now scales with window size instead of staying flat. At the
larger batch size where I originally hit an OOM (batch=8), the blocked
version used roughly 10x less memory than the masked version.

Happy to open a PR with this if it's useful, or if there's a reason
this tradeoff was intentional (e.g. simplicity, or block_size tuning
concerns) I'd be glad to hear it.

Qwertyemma:
Fixes #825

The sliding-window branch in `_sdpa_attention` built a full `(Tq, Tk)`
mask regardless of window size, then narrowed it after the fact. Passing
an arbitrary `attn_mask=` also forces SDPA onto its unfused math backend,
so a window=16 layer paid the same memory cost as unlimited attention on
this fallback path.

This splits the query dimension into blocks and slices k/v down to only
the range each block's window can see before calling SDPA, so the actual
tensors SDPA processes shrink with the window instead of staying at full
sequence length. Same causal + window math, computed in smaller pieces.

Measured on a Tesla T4 (no FA2/FA3), T=2048, batch=1:

    window=16    296 MB -> 41 MB
    window=64    301 MB -> 43 MB
    window=256   300 MB -> 47 MB
    window=1024  300 MB -> 64 MB

Output verified against the old masked implementation (max abs diff 0.0)
and against an independent brute-force reference across several window
sizes, GQA, and a non-block-divisible sequence length (T=513 with
block_size=256). Added tests to test_attention_fallback.py covering all
of this plus a memory-scaling check.

Tq==1 (single-token decode) is untouched - that path already sliced k/v
by window before calling SDPA, so it wasn't affected by this issue.

block_size defaults to 256, picked as a reasonable middle ground rather
than tuned - happy to adjust if there's a preferred value or if you'd
rather it be configurable.

### DIFF
--- a/nanochat/flash_attention.py
+++ b/nanochat/flash_attention.py
@@ -98,16 +98,66 @@ def _sdpa_attention(q, k, v, window_size, enable_gqa):
 
     # Need explicit mask for sliding window/chunk inference
     device = q.device
+
+    # Sliding window with multiple queries: block the query dimension so each
+    # block only ever sees its own windowed slice of k/v. Building a single
+    # (Tq, Tk) mask for the whole sequence forces SDPA onto its unfused math
+    # backend regardless of window size, so a small window still pays for a
+    # full-length score matrix. Slicing k/v per query block keeps the actual
+    # tensors SDPA sees proportional to the window instead of the full Tk.
+    if window >= 0 and window < Tk:
+        return _sdpa_windowed_blocked(q, k, v, window, enable_gqa)
+
     # For chunk inference (Tq != Tk), is_causal is not aligned to cache position => build an explicit bool mask
     row_idx = (Tk - Tq) + torch.arange(Tq, device=device).unsqueeze(1)
     col_idx = torch.arange(Tk, device=device).unsqueeze(0)
     mask = col_idx <= row_idx
 
-    # sliding window (left)
-    if window >= 0 and window < Tk:
+    return F.scaled_dot_product_attention(q, k, v, attn_mask=mask, enable_gqa=enable_gqa)
+
+
+def _sdpa_windowed_blocked(q, k, v, window, enable_gqa, block_size=256):
+    """
+    Blocked sliding-window attention for the SDPA fallback path.
+
+    Splits the query dimension into blocks and, for each block, slices k/v
+    down to only the range that block's window can see before calling SDPA.
+    This mirrors the block-skipping FlashAttention kernels do internally,
+    done at the tensor-slicing level since SDPA has no windowed mode to
+    call into directly on this fallback path.
+
+    q, k, v are (B, H, T, D) format. Mathematically equivalent to the
+    single-mask approach (same causal + window restriction), just computed
+    in smaller pieces so SDPA never sees more of k/v than the window needs.
+    """
+    device = q.device
+    Tq = q.size(2)
+    Tk = k.size(2)
+    offset = Tk - Tq  # where query positions start relative to key positions
+
+    outputs = []
+    for start in range(0, Tq, block_size):
+        end = min(start + block_size, Tq)
+        q_block = q[:, :, start:end, :]
+
+        # Earliest key any query in this block can see: (offset+start) - window.
+        # Latest key any query in this block can see: offset+end (causal, own position).
+        k_start = max(0, (offset + start) - window)
+        k_end = offset + end
+        k_block = k[:, :, k_start:k_end, :]
+        v_block = v[:, :, k_start:k_end, :]
+
+        block_q_len = end - start
+        block_k_len = k_end - k_start
+        row_idx = (offset + start - k_start) + torch.arange(block_q_len, device=device).unsqueeze(1)
+        col_idx = torch.arange(block_k_len, device=device).unsqueeze(0)
+        mask = col_idx <= row_idx
         mask = mask & ((row_idx - col_idx) <= window)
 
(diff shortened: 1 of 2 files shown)


eaglstun:
Reproduced this on Apple Silicon (M4 Max, macOS 26.4.1, torch 2.9.1). MPS is the other platform that lands on this fallback, since there is no FA2/FA3 on Mac.

Both implementations run in one process so they see the same allocator state. MPS has no peak-memory API, so peak comes from a poller thread sampling `torch.mps.current_allocated_memory()`.

## Memory

B=1, H=6, T=2048, D=64, fp32 — the shape from #825:

| window | master   | this PR | change |
| ------ | -------- | ------- | ------ |
| 16     | 148.0 MB | 9.0 MB  | 16.5x  |
| 64     | 148.0 MB | 9.0 MB  | 16.5x  |
| 256    | 148.0 MB | 9.0 MB  | 16.5x  |
| 512    | 148.0 MB | 21.0 MB | 7.1x   |

master is flat at 148.0 MB across every window, which is the behavior #825 describes.

## At the window nanochat actually generates

`_compute_window_sizes` sets `short_window = ceil(sequence_len/4/128)*128`, so S is 512 at `sequence_len=2048` and 128 at 512. L equals `sequence_len`, so `window < Tk` is false and that layer takes the `is_causal=True` branch without building a mask at all. 512 is the number that matters at the default sequence length, so the 7.1x row is the realistic one rather than the 16.5x rows.

For the `runcpu.sh` training shape (depth-6, seq 512, batch 32, giving window 128): 217.4 MB -> 124.6 MB, 1.7x. Smaller ratio because q/k/v themselves dominate at batch 32.

## Numerics

max abs diff against master is 0.0 at window 256 and 512. At window 16 and 64 it is ~4e-7, which is fp32 reassociation from the blocking rather than a correctness problem.

## Timing

Sequential A-then-B timing on MPS was too noisy to use: the same master workload varied 5.98 to 12.91 ms across trials at min-of-30. Interleaving the variants round-robin, so drift hits all of them equally, gives stable numbers (min of 40, medians agree with the ordering):

| shape              | master   | bs=128       | bs=256 (PR default) | bs=512        |
| ------------------ | -------- | ------------ | ------------------- | ------------- |
| T=512, w=128, B=32 | 10.23 ms | 6.41 (1.60x) | 7.65 (1.34x)        | 11.13 (0.92x) |
| T=2048, w=512, B=1 | 3.87 ms  | 4.22 (0.92x) | 2.61 (1.49x)        | 2.09 (1.85x)  |

The PR default of 256 beats master in both shapes, so nothing blocking here. But the best `block_size` inverts between the two, and each one's best is a small regression at the other. It might be worth scaling `block_size` with `Tq` instead of fixing it at 256, or noting in the docstring that 256 is a compromise.

## One more note

`--window-pattern` defaults to `SSSL`, and `base_train` prints a warning recommending `L` when running on SDPA. So the default config on a Mac does reach this path, and this PR makes the thing that warning is about considerably cheaper.

PR #830, opened 2026-08-15
Speed up the 8xH100 GPT-2 run to 81.8 minutes

giovannizinzi:
## Result

This is my next speedrun attempt, following the d22 + muon attempt in [nanochat #733](https://github.com/karpathy/nanochat/pull/733). It comes with a few speedups. Its main tradeoff is the new `liger-kernel` dependency. At the speedrun's 49,152-token vocabulary (the repository default remains 32,768), its efficient Triton cross-entropy cuts loss memory enough to use device batch 32 without OOMing.

| | Run 6 (current nanochat) | this PR |
|---|---:|---:|
| recipe | d24, ratio 8 | d22, ratio 9.4 + fused CE + selective RMSNorm scales |
| tokenizer vocabulary | 32,768 | 49,152 (speedrun override; default unchanged) |
| training time | 99.0 min | **81.835 +/- 0.138 min** |
| canonical val BPB | 0.71800 (N=5) | **0.718786 +/- 0.000054 (N=6)** |
| 22-task CORE | 0.262634 (N=5) | **0.261814 +/- 0.002295 (N=6)** |

Relative to Run 6, pretraining is **17.3% faster**. Val BPB differs by +0.000786 and CORE by -0.000820; CORE is within the benchmark's observed run-to-run noise, and all six runs clear the GPT-2 floor of 0.256525.

## What changed, why, and what it bought

| Change | Why | Measured effect / tradeoff |
|---|---|---|
| Restore pinned staging for `cuda:N` devices | DDP passes indexed CUDA devices, but the loader only pinned for the literal string `cuda`. | Saves **7.30 s** on average in full-recipe matches; no data-order or CPU-path change. |
| Reuse Muon communication workspaces | Shape-stable reduce-scatter/all-gather tensors were reallocated every step. | A matched 600-step screen fell from **376.96 s to 261.07 s**; long allocator stalls fell from **322 to 1**. Buffers remain outside checkpoint state. |
| Use a 49,152-token speedrun vocabulary | A larger tokenizer represents the same text with fewer tokens, so a fixed token budget covers more source text. | In an earlier controlled d22 sweep, mean CORE rose from **0.2539 to 0.2646** versus 32,768 (N=5 each); 65,536 added no CORE and was slower. The gain was not robust at d20/d26, so this remains a speedrun-only override; the repository default stays 32,768. |
| Optional Liger cross entropy | At the 49k vocabulary and device batch 32, each GPU produces ~6 GiB BF16 logits and native CE creates another ~12 GiB FP32 copy. | Device batch 32 / GA1 fits. Native CE OOMs at that shape; batch 16 / GA2 is **4.8% slower**. Evaluation and inference stay native. |
| Learn scales on MLP-input and final RMSNorms | d22 is faster per step but needs a small amount of channel-wise capacity back. Gains start at one, so initialization is unchanged. | A paired ablation improves canonical val BPB by **0.001396** for **0.98%** more time; mean CORE is unchanged. Attention-side norms remain fixed. |
| Count FP8 `Linear` subclasses in FLOPs | FP8 conversion hid converted matrix multiplies from the exact-class FLOP counter. | Corrects reported MFU from ~5% to ~56%; observability only, with no training-math or speed change. |
| Select the d22 ratio-9.4 recipe | Put the complete measured launch in `speedrun.sh`, including its 49k tokenizer and separate canonical eval. | Produces the final result above while keeping ClimbMix, Muon, scalar LR, RoPE, warmdown, final LR, and checkpoint interfaces unchanged. |

### Dependency tradeoff

The speed/memory gain requires one additional external package: pinned `liger-kernel==0.8.0` under an optional `liger` extra. Liger itself depends on Torch and Triton, which are already resolved by nanochat's GPU stack, so it is the only newly locked package.

The cost is an additional third-party compatibility. The benefit is using a maintained, tested Triton loss instead of vendoring a large custom kernel: it removes the FP32 loss copy, makes batch 32 fit, and is 4.8% faster than the native-loss batch-16 fallback. The integration is one training-only hook behind an explicit flag; installs and runs that do not request it are unchanged.

Matched d14/d20/d24 checks reduced peak allocator memory by **20.6% / 10.4% / 7.5%**; mean step time changed by **-1.7% / +0.23% / +0.39%**, with final BPB within 0.0002 at every depth.

### Why learnable RMSNorm is selective

RMSNorm still controls each activation vector's overall size. The added gains are per-channel "volume knobs" before the MLPs and output head, where the d22 ablation showed a validation benefit. Applying gains more broadly did not improve the tradeoff, so this PR leaves attention-side RMSNorms parameter-free. The setting is saved in model config so SFT reconstructs the checkpoint correctly.

### Promising directions not included

Cosmopedia was the strongest rejected data direction. A character-balanced 80/20 ClimbMix/Cosmopedia-v2 mix reached **73.917 min, 0.726888 val BPB, and 0.267123 CORE** at d22 (N=3). It did not transfer to d14: versus matched ClimbMix controls, mean val worsened by **0.004797** and CORE fell by **0.010313**, with CORE lower in every pair. A separate metadata screen that varied the college-textbook share inside a fixed 15% Cosmopedia dose also missed the CORE gate in every arm. This PR therefore stays on default ClimbMix rather than shipping a depth-specific data win. Research direction inspired by Textbooks are all you need (I'm pretty sure there's gains worth exploring further here!)

## Reproduce

```bash
torchrun --standalone --nproc_per_node=8 -m scripts.base_train -- \
    --depth=22 --target-param-data-ratio=9.4 \
    --device-batch-size=32 --total-batch-size=524288 \
    --fp8 --liger-cross-entropy --learnable-rmsnorm \
    --eval-every=999999 --core-metric-every=-1 --sample-every=-1 \
    --run="$WANDB_RUN"
```

### DIFF
--- a/nanochat/dataloader.py
+++ b/nanochat/dataloader.py
@@ -110,7 +110,7 @@ def tokenizing_distributed_data_loader_with_state_bos_bestfit(
 
     # Pre-allocate buffers once: layout is [inputs (B*T) | targets (B*T)]
     # This gives us contiguous views and a single HtoD transfer
-    use_cuda = device == "cuda"
+    use_cuda = torch.device(device).type == "cuda"
     row_buffer = torch.empty((B, row_capacity), dtype=torch.long) # for building rows without creating Python lists
     cpu_buffer = torch.empty(2 * B * T, dtype=torch.long, pin_memory=use_cuda) # staging area (CPU)
     gpu_buffer = torch.empty(2 * B * T, dtype=torch.long, device=device) # on-device buffer
--- a/nanochat/gpt.py
+++ b/nanochat/gpt.py
@@ -175,6 +175,7 @@ class GPT(nn.Module):
             "h": nn.ModuleList([Block(config, layer_idx) for layer_idx in range(config.n_layer)]),
         })
         self.lm_head = Linear(config.n_embd, padded_vocab_size, bias=False)
+        self._training_loss = None
         # Per-layer learnable scalars (inspired by modded-nanogpt)
         # resid_lambdas: scales the residual stream at each layer (init 1.0 = neutral)
         # x0_lambdas: blends initial embedding back in at each layer (init 0.0 = disabled)
@@ -199,6 +200,14 @@ class GPT(nn.Module):
         cos, sin = self._precompute_rotary_embeddings(self.rotary_seq_len, head_dim)
         self.register_buffer("cos", cos, persistent=False) # persistent=False means it's not saved to the checkpoint
         self.register_buffer("sin", sin, persistent=False)
+
+    def set_training_loss(self, loss):
+        """Install an optional fused mean loss without changing eval or inference."""
+        if not isinstance(loss, nn.Module):
+            raise TypeError("training loss must be an nn.Module")
+        if next(loss.parameters(), None) is not None or next(loss.buffers(), None) is not None:
+            raise ValueError("training loss must not own parameters or buffers")
+        self._training_loss = loss.to(device=self.get_device())
 
     @torch.no_grad()
     def init_weights(self):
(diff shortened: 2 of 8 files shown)


isubuz:
Just saw this PR. I think the `dataloader.py` change requires additional fixes now that H2D is actually async, as the CPU might race ahead of the GPU and the data in the staging buffer becomes inconsistent for GPU copy operations that's still catching up - https://github.com/karpathy/nanochat/pull/837

828Tina:
Hello, I noticed that you have opened two PRs, #733 and #830. Your tricks seem promising and appear to have the potential to improve training efficiency. May I ask why these PRs were closed? Was it due to technical issues?

PR #837, opened 2026-08-27
Stop requantizing fp8 weight; actually pin dataloader buffers

isubuz:
- **fp8**: the weight only changes once per optimizer step, but it was requantized on
every micro-batch. now quantized once per step and held. 
- **dataloader**: `device` is a `torch.device`, which never compares equal to a string, so
the buffer was never actually pinned. now that it is, reusing one hoisted buffer across
an async copy is unsafe — hence the per-batch alloc.


isubuz:
More context here - https://warpscale.ai/blog/deep-dive-into-nanochat-part-1/

PR #843, opened 2026-08-30
optim: keep async collective inputs and Work handles alive until wait()

sinameraji:
I hit this while pretraining a 1.5B Persian model on nanochat (8×H100, bf16), so the Muon path has had a hard workout. Sharing a latent lifetime bug I found in it.

`MuonAdamW.step()` overlaps communication with compute by launching `reduce_scatter_tensor` / `all_gather_into_tensor` with `async_op=True`, keeping only the returned Future. Two lifetime issues:

1. In `_compute_muon`, the input of the final all_gather, `updated_params`, is a local tensor that goes out of scope as soon as the method returns — before `_finish_gathers` waits on the collective. Nothing else references it, so the CUDA caching allocator can hand its block to the next allocation (usually the next group's `updated_params`, immediately overwritten) while NCCL is still reading it.
2. The `Work` objects are dropped everywhere (`.get_future()` only). On current PyTorch (checked on 2.9.1 + NCCL 2.27.5) the NCCL backend doesn't `record_stream` collective arguments by default, so correctness relies on the caller keeping tensors referenced until `wait()`.

If the timing lines up, the gather can deliver torn data: every rank keeps its own shard of each Muon matrix correct and stale bytes for the other `world_size - 1` shards.

**Fix:** keep each async collective's input tensor and its `Work` referenced in the info dicts that already flow to the waits, and have `_finish_gathers` call `work.wait()` alongside `future.wait()` before the copy-back. No behaviour change on the happy path, no extra memory beyond what was already in flight, no extra synchronisation beyond the existing waits.

**Honest scope:** I first suspected this path when my run hit repeated loss plateaus around step 27,000, but forensics exonerated it for my actual incident — with updates frozen at a plateau, all-rank parameter checks showed 0/255 tensors differing, and the plateaus reproduced under fully synchronous collectives too. My real root cause was attention-logit growth in a custom no-QK-norm model variant, unrelated to this optimizer. Submitting anyway: the lifetime hazard is real under current PyTorch semantics and cheap to close. I'm presenting it as a hardening fix with a consistency test, not a field-failure fix — `repro_torn_gather.py` (in a comment below, not the diff) churns the allocator between steps and asserts cross-rank parameter equality, but I did not observe divergence I could attribute to this bug in production.

**Environment:** torch 2.9.1+cu128, NCCL 2.27.5, CUDA 12.8, 8×H100 (GCE a3-highgpu-8g), nanochat `92d63d4`

### DIFF
--- a/nanochat/optim.py
+++ b/nanochat/optim.py
@@ -281,15 +281,15 @@ class MuonAdamW(torch.optim.Optimizer):
                 param_infos[p] = dict(future=None, grad_slice=grad, is_small=True)
             elif p.numel() < 1024:
                 # Small params: all_reduce (no scatter/gather needed)
-                future = dist.all_reduce(grad, op=dist.ReduceOp.AVG, async_op=True).get_future()
-                param_infos[p] = dict(future=future, grad_slice=grad, is_small=True)
+                work = dist.all_reduce(grad, op=dist.ReduceOp.AVG, async_op=True); future = work.get_future()
+                param_infos[p] = dict(future=future, grad_slice=grad, is_small=True, work=work)
             else:
                 # Large params: reduce_scatter
                 assert grad.shape[0] % world_size == 0, f"AdamW reduce_scatter requires shape[0] ({grad.shape[0]}) divisible by world_size ({world_size})"
                 rank_size = grad.shape[0] // world_size
                 grad_slice = torch.empty_like(grad[:rank_size])
-                future = dist.reduce_scatter_tensor(grad_slice, grad, op=dist.ReduceOp.AVG, async_op=True).get_future()
-                param_infos[p] = dict(future=future, grad_slice=grad_slice, is_small=False)
+                work = dist.reduce_scatter_tensor(grad_slice, grad, op=dist.ReduceOp.AVG, async_op=True); future = work.get_future()
+                param_infos[p] = dict(future=future, grad_slice=grad_slice, is_small=False, work=work)
         return dict(param_infos=param_infos)
 
     def _reduce_muon(self, group: dict, world_size: int) -> dict:
@@ -313,9 +313,9 @@ class MuonAdamW(torch.optim.Optimizer):
 
         # Reduce_scatter to get this rank's chunk
         grad_chunk = torch.empty(chunk_size, *shape, dtype=dtype, device=device)
-        future = dist.reduce_scatter_tensor(grad_chunk, stacked_grads, op=dist.ReduceOp.AVG, async_op=True).get_future()
+        work = dist.reduce_scatter_tensor(grad_chunk, stacked_grads, op=dist.ReduceOp.AVG, async_op=True); future = work.get_future()
 
-        return dict(future=future, grad_chunk=grad_chunk, stacked_grads=stacked_grads, chunk_size=chunk_size)
+        return dict(future=future, grad_chunk=grad_chunk, stacked_grads=stacked_grads, chunk_size=chunk_size, work=work, grad_stack_alive=grad_stack)
 
     def _compute_adamw(self, group: dict, info: dict, gather_list: list, rank: int, world_size: int) -> None:
         """Wait for reduce, compute AdamW updates, launch gathers for large params."""
@@ -356,8 +356,8 @@ class MuonAdamW(torch.optim.Optimizer):
 
             # Large params need all_gather
             if not pinfo['is_small']:
-                future = dist.all_gather_into_tensor(p, p_slice, async_op=True).get_future()
-                gather_list.append(dict(future=future, params=None))
+                work = dist.all_gather_into_tensor(p, p_slice, async_op=True); future = work.get_future()
+                gather_list.append(dict(future=future, params=None, work=work, gather_inputs_alive=p_slice))
 
     def _compute_muon(self, group: dict, info: dict, gather_list: list, rank: int) -> None:
         """Wait for reduce, compute Muon updates, launch gather."""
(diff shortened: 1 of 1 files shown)


sinameraji:
Consistency / stress test mentioned in the description — `repro_torn_gather.py`. Run from the repo root on ≥2 GPUs: `torchrun --standalone --nproc_per_node=8 repro_torn_gather.py --steps 500`. It drives `MuonAdamW.step()` with random gradients, churns the allocator between steps with same-size-class tensors, and asserts bit-identical parameters across ranks after every step (a correct gather guarantees this by construction). On the patched optimizer it runs clean; the race is timing-dependent, so a clean run on the unpatched code is possible on some hardware — raising `--churn`/`--steps` makes it fire more reliably.

```python
"""
Minimal repro / detector for torn all_gather results in nanochat's MuonAdamW (nanochat/optim.py @ 92d63d4).

Run on >= 2 GPUs from the nanochat repo root:
    torchrun --standalone --nproc_per_node=8 repro_torn_gather.py            # unpatched: expect DIVERGED ranks / loss plateaus
    torchrun --standalone --nproc_per_node=8 repro_torn_gather.py --steps 2000

What it does: builds a stack of Muon-shaped matrices and a couple of AdamW params, drives MuonAdamW.step() with random
gradients for N steps, and after EVERY step checks that all ranks hold bit-identical parameters (a correct all_gather makes
them identical by construction). Between steps it churns the caching allocator with tensors of the same size class as the
optimizer's gather input, which is what a real training step's activations do. With the 92d63d4 optimizer the gather INPUT
(`updated_params` in _compute_muon) is a local that is dropped before the collective is waited on, so the allocator can
reuse its block while NCCL is still reading it -> ranks diverge (torn parameters). With the patched optimizer the check
never fires. The race is timing-dependent: if a run stays clean, raise --churn or --steps.
"""
import argparse, os, torch, torch.distributed as dist
from nanochat.optim import MuonAdamW

ap = argparse.ArgumentParser()
ap.add_argument("--steps", type=int, default=500)
ap.add_argument("--layers", type=int, default=28)          # number of matrices per Muon group (1.5B config: 28 layers)
ap.add_argument("--dim", type=int, default=2048)
ap.add_argument("--churn", type=int, default=8)            # allocations per step that mimic activation traffic
args = ap.parse_args()

dist.init_process_group("nccl"); rank = dist.get_rank(); world = dist.get_world_size()
dev = torch.device("cuda", int(os.environ["LOCAL_RANK"])); torch.cuda.set_device(dev)
torch.manual_seed(0)  # identical init on all ranks

# Muon groups shaped like a Llama/GPT block: attn (dim x dim), kv (dim/4 x dim), mlp up (3dim x dim), mlp down (dim x 3dim)
shapes = [(args.dim, args.dim), (args.dim // 4, args.dim), (3 * args.dim, args.dim), (args.dim, 3 * args.dim)]
groups = []
for shp in shapes:
    ps = [torch.nn.Parameter(torch.randn(*shp, device=dev, dtype=torch.bfloat16) * 0.02) for _ in range(args.layers)]
    groups.append(dict(kind="muon", params=ps, lr=0.02, momentum=0.95, ns_steps=5, beta2=0.9, weight_decay=0.0))
wte = torch.nn.Parameter(torch.randn(32768, args.dim, device=dev, dtype=torch.bfloat16))          # AdamW large param (reduce_scatter + all_gather)
gain = torch.nn.Parameter(torch.ones(args.dim, device=dev, dtype=torch.bfloat16))                   # AdamW small param (all_reduce)
groups += [dict(kind="adamw", params=[wte], lr=0.2, betas=(0.8, 0.995), eps=1e-10, weight_decay=0.0),
           dict(kind="adamw", params=[gain], lr=0.01, betas=(0.8, 0.95), eps=1e-10, weight_decay=0.0)]
opt = MuonAdamW(groups)
all_params = [p for g in groups for p in g["params"]]
gen = torch.Generator(device=dev); gen.manual_seed(1234 + rank)   # different grads per rank, like real data parallelism

def ranks_agree():
    """True iff every parameter is bit-identical across ranks (max == min elementwise after all_reduce)."""
    bad = []
    for i, p in enumerate(all_params):
        mx = p.detach().float().clone(); mn = p.detach().float().clone()
        dist.all_reduce(mx, op=dist.ReduceOp.MAX); dist.all_reduce(mn, op=dist.ReduceOp.MIN)
        if not torch.equal(mx, mn): bad.append((i, tuple(p.shape), (mx - mn).abs().max().item()))
    return bad

junk = []
for step in range(1, args.steps + 1):
    for p in all_params: p.grad = torch.randn(p.shape, device=dev, dtype=p.dtype, generator=gen) * 1e-3
    opt.step()
    # allocator churn: same size class as the Muon gather input (chunk_size x shape) and typical activation tensors
    junk = [torch.empty(((args.layers + world - 1) // world), *shapes[step % len(shapes)], device=dev, dtype=torch.bfloat16).fill_(float(step)) for _ in range(args.churn)]
    junk += [torch.randn(8, 2048, args.dim, device=dev, dtype=torch.bfloat16) for _ in range(2)]
    for p in all_params: p.grad = None
    if step % 10 == 0 or step <= 5:
        bad = ranks_agree()
        if bad:
            if rank == 0: print(f"DIVERGED at step {step}: {len(bad)} params differ across ranks, e.g. {bad[:3]}", flush=True)
            dist.barrier(); dist.destroy_process_group(); raise SystemExit(1)
        elif rank == 0 and step % 100 == 0: print(f"step {step}: ranks agree", flush=True)
if rank == 0: print(f"OK: {args.steps} steps, ranks agree on every check")
dist.barrier(); dist.destroy_process_group()
```

PR #850, opened 2026-09-08
Batch size ramp for larger models: 1/4 -> 1/2 -> full over the first 10% of tokens

lingwei-gu:
**Problem.** The auto batch-size rule was validated at d12 (2^19) and d26 (2^20) and extrapolates to 2^21 for d32+. A d34 trained on ClimbMix at 2^21 (current master, everything else default, 8xH100, one pass over the corpus) was 4-6x less token-efficient than the d24 speedrun: the d24 reaches val bpb 0.7185 in 5.8B tokens, this d34 needed 34B, and CORE was flat at ~0.30 from step 14k to 24k.

| tokens | 4.2B | 12.6B | 25.2B | 37.7B | 50.3B |
|---|---|---|---|---|---|
| val bpb | 0.7529 | 0.7276 | 0.7192 | 0.7169 | 0.7143 |

The critical batch size is small early in training and grows as the loss falls; a single large batch from step 0 wastes most of the early tokens.

**Change.** `--batch-ramp=1` (default off): train at 1/4 of the batch for the first 3% of the token budget and 1/2 for the next 7%, then the full batch, with sqrt LR scaling per stage. Same token budget, ~16% more optimizer steps. The auto batch is also capped at 2^20, the largest validated value. For d34 this gives 262,144 -> 524,288 -> 1,048,576.

**Status.** The d34 ramp run (ClimbMix, one epoch) is in progress; I will post its curve at matched token counts here.

### DIFF
--- a/scripts/base_train.py
+++ b/scripts/base_train.py
@@ -281,6 +281,14 @@ if total_batch_size == -1:
     batch_size_ratio = target_tokens / D_REF
     predicted_batch_size = B_REF * batch_size_ratio ** 0.383
     total_batch_size = 2 ** round(math.log2(predicted_batch_size)) # clamp to nearest power of 2 for efficiency
+    # Cap at the largest empirically validated batch (2^20, tuned at d26). The D^0.383 extrapolation reaches 2^21 at d32+,
+    # and a d34 trained at 2^21 was ~4-6x less token-efficient than the d24 speedrun (see the PR that added this cap):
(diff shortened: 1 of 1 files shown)


lingwei-gu:
Superseded by https://github.com/karpathy/nanochat/pull/851 (same branch, now a batch-size ramp instead of a fixed cap).

PR #851, opened 2026-09-08
Batch size ramp for larger models: 1/4 -> 1/2 -> full over the first 10% of tokens

lingwei-gu:
**Problem.** The auto batch-size rule was validated at d12 (2^19) and d26 (2^20) and extrapolates to 2^21 for d32+. A d34 trained on ClimbMix at 2^21 (current master, everything else default, 8xH100, one pass over the corpus) was 4-6x less token-efficient than the d24 speedrun: the d24 reaches val bpb 0.7185 in 5.8B tokens, this d34 needed 34B, and CORE was flat at ~0.30 from step 14k to 24k.

| tokens | 4.2B | 12.6B | 25.2B | 37.7B | 50.3B |
|---|---|---|---|---|---|
| val bpb | 0.7529 | 0.7276 | 0.7192 | 0.7169 | 0.7143 |

The critical batch size is small early in training and grows as the loss falls; a single large batch from step 0 wastes most of the early tokens.

**Change.** `--batch-ramp=1` (default off): train at 1/4 of the batch for the first 3% of the token budget and 1/2 for the next 7%, then the full batch, with sqrt LR scaling per stage. Same token budget, ~16% more optimizer steps. The auto batch is also capped at 2^20, the largest validated value. For d34 this gives 262,144 -> 524,288 -> 1,048,576.

**Tested.** Single-GPU d34 run with `--batch-ramp=1`: the stage boundaries land where computed, tokens/s tracks each batch change, and resuming a checkpoint from inside a stage rebuilds the same schedule.

**Status.** The d34 ramp run (ClimbMix, one epoch) is in progress; I will post its curve at matched token counts here.

### DIFF
--- a/scripts/base_train.py
+++ b/scripts/base_train.py
@@ -356,30 +356,6 @@ else:
 total_tokens = total_batch_size * num_iterations # the actual number of tokens we will train for
 print0(f"Total number of training tokens: {total_tokens:,}")
 print0(f"Tokens : Scaling params ratio: {total_batch_size * num_iterations / num_scaling_params:.2f}") # e.g. Chinchilla was ~20
-
-# 5) Batch size ramp: the critical batch size is small early in training, so a fixed large batch wastes the early tokens.
-# With --batch-ramp we train at 1/4 of the batch for the first 3% of the token budget and 1/2 for the next 7%, then the full
-# batch, scaling the LR by sqrt(batch) per stage. The token budget is unchanged; the number of optimizer steps grows ~16%.
-ramp_stages = [] # list of (first step of the NEXT stage, batch divisor of this stage)
-if args.batch_ramp:
-    budget_tokens = num_iterations * total_batch_size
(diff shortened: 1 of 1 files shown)


lingwei-gu:
Interim result from the d34 ramp run (262,144 -> 524,288 at 10B tokens -> 1,048,576 at 30.4B tokens, otherwise the same setup as the fixed-2^21 run above), val bpb at matched token counts:

| tokens | 4.2B | 8.4B | 12.6B | 25.2B | 33.6B |
|---|---|---|---|---|---|
| fixed 2^21 | 0.7529 | 0.7338 | 0.7276 | 0.7192 | 0.7178 |
| ramp | 0.7532 | 0.7395 | 0.7172 | 0.7102 | 0.6929 |

Each stage flattens as its batch gets outgrown and the next stage drops the loss again (e.g. 0.7081 -> 0.6929 within 3k steps of the 1M switch); CORE is 0.318 vs 0.302 at 33.6B. Will post the end-of-run numbers.

